PR #1, opened 2016-10-26
Basic CSS reorg in components

furilo:


furilo:
@ferblape - Probably it makes sense to get rid of the assets/stylesheets/gobierto_budgets and leave all CSS in assets/stylesheets/

PR #2, opened 2016-11-02
Update Rails configuration files to match 5.0 syntax

danguita:
This PR applies most of the diff generated by running the `rails app:update` task, keeping the previous settings whenever possible.


ferblape:
Looking good 👍

PR #11, opened 2016-11-04
Sandbox admin WIP

furilo:
This contains a partial reorg of CSS files and some admin screens, in case you want to take a look. 

Pending: 
- JS files reorg
- More admin screens


danguita:
@furilo Thanks, this helps a lot 👍

furilo:
I've added basic pages for the Consultations module: 

- http://madrid.gobierto.dev/sandbox/admin_consultations
- http://madrid.gobierto.dev/sandbox/admin_consultations_consultation - in this screen, if you click on the second tab, then on the first budget line, the modal shows up. 

gobierto_consultations

I've also added a new library to handle the date selection; but it's only working fine on the first page load, not if you have browsed through any page (even though its inside the $(document).on('turbo links:load') - same result if you put it in the rebindAll function.

Issue #10, opened 2016-11-04
Admin namespace preparations


danguita:
Here's a brief checklist:

- [x] `Admin` model
- [x] `admin` namespace
- [x] Authentication and authorization
- [x] Starting `admin` layout

danguita:
Updating this one to not make reference to any Gobierto modules integration for now, so these processes can be fine-grained.

PR #13, opened 2016-11-04
Admin namespace starting point

danguita:
This PR implements #10.

### What does this PR do?

It provides a starting point to work on the Admin namespace, including a basic authentication system and some test guidelines.

You should be asked for login credentials when trying to reach the admin namespace at http://gobierto.dev/admin/.


ferblape:
👍

PR #23, opened 2016-11-08
Update the database seeding process to make use of fixtures

danguita:
### What does this PR do?

It updates the `db:seeds` task to perform a fixture load operation. This way we can take advantage of the current fixtures to properly seed new databases without having to maintain them twice.

Run the `rake db:seeds` task and check you have proper `Site` and `Admin` records.


ferblape:
👍

Issue #17, opened 2016-11-07
Admin namespace: Sites management


danguita:
### Concept

TBD

### Check-list

- [x] Site management: Create, Update, switch statuses.
- [x] Draft status concept.
- [x] Module assignments.
- [x] Extra fields management.

danguita:
First iteration was done, will create new Issues for handling any other changes.

PR #25, opened 2016-11-09
Admin's Site management refactor

danguita:
This PR implements #17.

### What does this PR do?

It completes a new iteration on the Admin's Site management section by setting some more foundations, allowing module management and improving the abstraction layer between the Site form and the actual Repository.

It also implements the behavior behind Admin's authorization levels by allowing `manager` labeled one to access every Site available out there. On the other hand, `regular` Admins are only allowed to manage their own Sites.

Check out the Sites management section in Admin's namespace at http://gobierto.dev/admin/sites/, and perform any kind of operation to check how data is transformed and stored.


danguita:
A new iteration is about to come. The build is bright green so I'm moving this forward 🚀

PR #30, opened 2016-11-09
Quick and dirty strategy for hiding draft Sites

danguita:
This PR implements #17.

### What does this PR do?

It implements a quick and dirty strategy for hiding draft Sites, just as a first approach.

Check that none of the draft Sites can't be accessed through the default namespace.


danguita:
Nothing more to say 😛

PR #33, opened 2016-11-10
Refactor SiteForm object implementation for convenience

danguita:
This PR implements #17.

### What does this PR do?

- Namespace `SiteForm` with `Admin` module.
- Let's pretend any `SiteForm` represents an actual Rails model by delegating `to_model` method.
- Submit SiteForm parameters under `site` key.

Check that any Site management actions still works as expected.


danguita:
Those are just a few improvements, not a big deal.

Issue #29, opened 2016-11-09
Track activity log


ferblape:
Every action that updates the database (creates, updates or destroys a record) must be tracked in the system.

These actions need to reflect:

- the action itself. I.e. `site.created` or `user.updated`
- the author of the action
- the subject of the action
- the receiver of the action, if applies
- the IP from which the author executed the action
- the timestamp

There are three kinds of actions:

1 - actions performed by regular users (citiziens) in the public or private site. I.e.: follow a budget line, vote for a comment, or update her personal data
2 - actions performed by an admin in a site, such as create a new participatory process
3 - actions performed by an admin not linked to a site, such as create a new admin, enable a site module, or disable a user account

The activity log will be displayed in different parts of the admin UI, not totally defined yet. For the moment:

- when visiting a user profile or admin profile, all the activity authored by the user will be displayed
- pending to define: 
  - where to view generic activity (3)

ferblape:
Take a look @danguita. Tomorrow I'll start a branch and we can discuss about it on Friday morning

furilo:
We have to start thinking how we'll handle actions like votes which must be
tracked but not related to any user

ferblape:
I'm going to close this issue and create more specific ones.

PR #35, opened 2016-11-10
Track activity log

ferblape:
This PR implements #29

### What does this PR do?

This PR introduces a basic model Activity, taking into consideration the requirements described in #29.

It also creates a Pub/Sub architecture based in ActiveSupport::Notifications to decouple Activity tracking from the base code. It can be seen in the `Admin::SitesController` which I have used to validate this implementation.

For the moment we are not rendering any activity, just storing it in the db. 

Go to the sites UI and create, update and delete a site. Each of the three actions should produce an Activity if the site is valid and the action could finish successfully (i.e. if you update a site with invalid attributes, no action will be recorded).

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161111060751) do
+ActiveRecord::Schema.define(version: 20161111111116) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -32,6 +32,15 @@ ActiveRecord::Schema.define(version: 20161111060751) do
     t.index ["site_id"], name: "index_activities_on_site_id", using: :btree
     t.index ["subject_id", "subject_type"], name: "index_activities_on_subject_id_and_subject_type", using: :btree
     t.index ["subject_type", "subject_id"], name: "index_activities_on_subject_type_and_subject_id", using: :btree
+  end
+
+  create_table "admin_permissions", force: :cascade do |t|
+    t.integer "admin_id"
+    t.string  "namespace",     default: "", null: false
+    t.string  "resource_name", default: "", null: false
+    t.string  "action_name",   default: "", null: false
+    t.index ["admin_id", "namespace", "resource_name", "action_name"], name: "index_admin_permissions_on_admin_id_and_fields", using: :btree
+    t.index ["admin_id"], name: "index_admin_permissions_on_admin_id", using: :btree
   end
 
   create_table "admin_sites", force: :cascade do |t|


ferblape:
@danguita only if you have time, take a look to discuss it tomorrow in the call 👼

danguita:
@ferblape 👀 !

danguita:
@ferblape The approach looks great to me 👍  Let me just leave you some comments so that we can discuss them tomorrow.

app/pub_sub/subscribers/site_activity.rb
```diff
@@ -0,0 +1,33 @@
+module Subscribers
+  class SiteActivity < ::Subscribers::Base
+    def site_created(event)
+      return if event.payload[:subject].nil? || event.payload[:subject].new_record?
```
danguita: Regarding these checks before actually tracking anything, I'm wondering if we should just trust every message at this point so we're not adding any extra logic here. I mean to broadcast events only when things have been done instead of using callbacks that are executed regardless the result. How about placing the activity tracking methods (`track_<action>_activity`) inside every controller action, when the record is saved, deleted and so on?

Since we'll be adding some more complexity to these actions, I'd also consider abstracting these group or operations (i.e. get parameters + save the record + trigger callbacks) into service objects. Let's catch-up tomorrow if it's OK with you.

app/pub_sub/subscribers/base.rb
```diff
@@ -0,0 +1,25 @@
+module Subscribers
+  class Base
+    class_attribute :subscriptions_enabled
```
danguita: What is this attribute used for?

app/pub_sub/publisher.rb
```diff
@@ -0,0 +1,36 @@
...
+      event_name = [pub_sub_namespace, event_name].compact.join('.')
+      if block_given?
+        ActiveSupport::Notifications.instrument(event_name, payload) do
+          yield
+        end
+      else
+        Rails.logger.info [event_name, self.pub_sub_namespace].join(' - ')
+        ActiveSupport::Notifications.instrument(event_name, payload)
+      end
+      return true
```
danguita: `return` statement not needed here. Not big deal at all 😛

app/pub_sub/subscribers/site_activity.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def site_deleted(event)
+      return if !event.payload[:subject].destroyed?
+
+      create_activity_from_event(event)
+    end
+
+    private
+
+    def create_activity_from_event(event)
```
danguita: I'd say storing record changes would be great for providing context. How about keeping them in raw format, whenever available?

app/pub_sub/subscribers/site_activity.rb
ferblape: I totally agree, I wasn´t sure of this approach, but once implemented it doesn´t makes sense to populate the system with messages that need to be ignored!

app/pub_sub/subscribers/site_activity.rb
ferblape: Yes, that´s something I suspect we´lll be adding in the closer future, once we start to render the activity and see how it looks. For the moment I was just skipping this until we face the wall.

app/pub_sub/subscribers/base.rb
ferblape: My fault, the idea is to disable subscriptors during the tests, unless we want to test them explicitly. 

This is probably premature optimization, but I remember how slow was Movida test suite because of the Activity tracking.

danguita: Got it, makes sense. I'd leave it as is in that case.

app/pub_sub/subscribers/site_activity.rb
danguita: Sure, that seems sensible. I think we can manage to make simpler subscribers (almost logic-less) by triggering publishers after certain events. That may add some complexity to the action's workflow but I'd say it's totally worth it.

danguita:
Thanks for your feedback! LGTM 👍

ferblape:
@danguita should I wait for your Operation service class PR?

danguita:
@ferblape I'd take care later in a new PR to not mess with the activity tracking stuff. Let's just merge these changes if you agree.

Issue #16, opened 2016-11-07
Admin namespace: Admins management


danguita:
### Concept

TBD

### Check-list

- [x] Admin creation.
  - [x] Manual.
  - [x] Invitation.
- [x] Protected Admin concept (*Godmode*).
- [x] Status management.
- [x] Site assignments.
- [x] Module assignments.
- [x] Other permissions.

danguita:
All checked!

PR #37, opened 2016-11-10
Admin’s admin management starting point

danguita:
This PR implements #16.

### What does this PR do?

It implements the Admins management starting point, taking care of Site assignment and Site Module authorization as part of the permissions back-end.

Check out the Admin management section under admin namespace at http://gobierto.dev/admin/admins.


danguita:
Moving this one forward.

PR #40, opened 2016-11-12
Refactor SiteConfiguration class to handle properties

danguita:
### What does this PR do?

It slightly updates the `SiteConfiguration` class to take control on passed parameters in terms of data structure. It also introduces a simple API to access these properties in a more conventional way. 

It affects almost every part in the application, so let's just check everything regarding site modules and demo mode is working as expected.


ferblape:
Looks fantastic!

PR #41, opened 2016-11-14
Admin's Site username and password management when in draft mode

danguita:
This PR implements #17.

### What does this PR do?

It takes advantage of the `password_protection_username` and `password_protection_password` Site configuration options to store and check site credentials when in draft mode. This implementation is based on the `SiteForm` object so these parameters are being properly mapped into the actual configuration hash.

It also integrates the "password protected" widget provided in sandbox files.

- Go to the Sites management section in Admin's namespace at http://gobierto.dev/admin/sites/ and pick one Site to edit.
- Check that a record can't be saved without setting username and password while in draft mode.
- Regarding the UI, check that the username and password fields are shown only when the Draft visibility level radio button is checked.

### DIFF
--- a/test/controllers/admin/sites_controller_test.rb
+++ b/test/controllers/admin/sites_controller_test.rb
@@ -25,7 +25,7 @@ class Admin::SitesControllerTest < ActionController::TestCase
   end
 
   def test_create_site_broadcasts_event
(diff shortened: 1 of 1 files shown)


ferblape:
LGTM, although we need to fix conflicts. I can take care.

danguita:
@ferblape Oh, let me know if you need some help!

PR #44, opened 2016-11-14
Implement Admin active scope

danguita:
This PR implements #16.

### What does this PR do?

Implement Admin active scope taking the `authorization_level` attribute into account.

Just check that any Admin with the *disabled* `authorization_level` can't log in to the system.


danguita:
I'm merging this since it is not a big deal.

PR #45, opened 2016-11-14
Add activities to admin UI

ferblape:
This PR implements part of #29

### What does this PR do?

This PR adds a couple of lists for activities in the UI

1 - a list just for admins with activities non related to any site specifically

2 - a list of the activities performed by an admin

First, update a site, or create a new site. After that:

1 - Go to /admin/activities and see if there are activities

2 - Go to your admin user page and see the activities


danguita:
LGTM 👍

PR #50, opened 2016-11-15
New types of site update activities

ferblape:
This PR implements #29

### What does this PR do?

This PR splits site updated activities to highlight when a site visibility has changed or the modules of a site has been enabled or disabled.

Just update the site visibilty or the modules of a site, and you should see the new types of activities.


ferblape:
Merging, because it's not a big deal.

Issue #31, opened 2016-11-10
User (Citizen) session handling


danguita:
### Concept

TBD

### Check-list

- [x] Registration
- [x] Confirmation
- [x] Password recovery
- [x] Authentication and authorization
- [x] Communications
  - [x] Confirmation email
  - [x] Password reset email
  - [x] Welcome email

danguita:
All checked!

PR #52, opened 2016-11-16
User data model and starting behavior

danguita:
This PR implements #31.

### What does this PR do?

This PR introduces the `User` model and its starting behavior. It also refactors some parts of the system that can be shared between resources that handle any kind of user session.

This cannot be tested since it just prepares the application for coming features.


ferblape:
Looking good!

PR #61, opened 2016-11-18
Limit user management to the current site scope

danguita:
Connects to #57.

### What does this PR do?

This PR limits the user management feature through the Admin UI by filtering the users whose source site does not match the current site session.

Go to the user management section within the Admin namespace (http://gobierto.dev/admin/users), and check how the user list changes as you switch the current site session.


ferblape:
👍

PR #62, opened 2016-11-18
Auto set first available Site when no other is selected

danguita:
Connects to #60.

### What does this PR do?

The PR's title is self explanatory, there is not much else to say :)

Just check that the first available Site (alphabetically sorted) is selected after getting into the Admin UI for the first time.


ferblape:
As simple as that 👍

Issue #59, opened 2016-11-18
Protect User sign in and sign up actions against already signed in users


danguita:
- [x] Redirect them to `after_sign_in_path`.
- [x] Show them the "You are already signed in" message.

/cc @ferblape

ferblape:
👍

PR #64, opened 2016-11-18
Require no authentication in determined User session related actions

danguita:
Connects to #59.

### What does this PR do?

As the title says, we're requiring no authentication on every User and Admin session related actions. Those are:

#### In Admin namespace

- Sign in
- Confirmation
- Invitation acceptance
- Password change request
- Password change

#### In User namespace

- Registration
- Sign in
- Confirmation
- Invitation acceptance
- Password change request
- Password change

It also increases the test coverage and normalizes implementations across namespaces.

Just check every single action from the list above halts the workflow by showing the "You are already signed in" message.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,6 +10,7 @@ Rails.application.routes.draw do
   # Admin module
   namespace :admin do
     get '/' => 'welcome#index', as: :root
+    get '/login' => 'sessions#new'
 
     resource :sessions, only: [:new, :create, :destroy]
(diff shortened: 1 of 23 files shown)


config/routes.rb
```diff
@@ -10,6 +10,7 @@
   # Admin module
   namespace :admin do
     get '/' => 'welcome#index', as: :root
+    get '/login' => 'sessions#new'
```
ferblape: 👀 not a big deal, but you are duplicating here sessions#new

PR #66, opened 2016-11-18
Limit Site management actions to Admins with manager authorization level

danguita:
Connects to #63.

### What does this PR do?

This PR implements both `Admin::LayoutPolicy` and `Admin::SitePolicy` objects to provide an access control mechanism to UI pieces on the admin layout and actions respectively.

In terms of behavior, only Admins with manager authorization level are authorized to perform any kind of operations against Sites. That includes showing the "Manage sites" link option in the layout.

Just check the "Manage sites" link option is never shown to regular Admins, and every Site management related URL redirects to the request referrer rendering the following message: "You are not authorized to perform this action.".

Here are some of those now protected action URLs:

- http://gobierto.dev/admin/sites
- http://gobierto.dev/admin/sites/new
- http://gobierto.dev/admin/sites/1/edit

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -10,5 +10,7 @@ class Admin::BaseController < ApplicationController
   helper_method :managed_sites
   helper_method :can_manage_sites?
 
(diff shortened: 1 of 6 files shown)


ferblape:
Works like a charm

Issue #56, opened 2016-11-17
User verification process against census data


danguita:
Depends on https://github.com/PopulateTools/gobierto-dev/issues/55.

Check-list:

- [x] The user will explicitly request a verification process.
- [x] The verification process may require filling in some more fields within the User context.
- [x] The verification process will produce a verification status for the given User and data.
- [x] Any verification request will be stored, so that it can be regenerated whenever needed.

PR #72, opened 2016-11-22
User verification process against Census data

danguita:
Connects to #56.

### What does this PR do?

Continuing with the changes introduced in #69, this PR implements a verification process against that Census data. It can be requested by any signed in User, and it may require filling up new fields to complete the verification process. The current approach adds a generic user verification mechanism that could have multiple implementations, starting from the verification against Census data which is already working here.

In this case we're asking for the document number and date of birth fields to match the Census data through the recently created `CensusRepository`. Digging a little bit deeper on the implementation, the `User::Verification::CensusVerification` model inherits from `User::Verification` and specializes it by adding some Census-specific behavior such as handling the required fields, that are serialized for ease and portability.

The rest of logic is handled by the corresponding form object (`User::CensusVerificationForm`), as usual.

The verification process can be requested by any unverified Users, so just go to http://madrid.gobierto.dev/user/verifications/new (also reachable from the session navbar, on top of page when signed in, just to keep it in mind) and submit the form to start the process. If everything goes well, you should be redirected to the verification status page at http://madrid.gobierto.dev/user/verifications. Here you can see how your current verification request is going (if it couldn't be calculated instantly).

Please note this is a very early implementation of the UI, just as a starting point for any coming features.

### Notes

I'm not implementing this point here to have better context:

> - Any verification status will be regenerated each time the census data source changes (i.e. reload the census CSV file).

Will take care in a new issue.

### DIFF
--- a/docs/user-namespace.md
+++ b/docs/user-namespace.md
@@ -8,11 +8,11 @@ URL: http://madrid.gobierto.dev/user/login
 
 Available users:
 
-| Email               | Password | Notes                                                                        |
-| ---                 | ---      | ---                                                                          |
-| dennis@gobierto.dev | gobierto | regular user, madrid.gobierto.dev as source site                             |
-| reed@gobierto.dev   | gobierto | regular user, unconfirmed, madrid.gobierto.dev as source site                |
-| susan@gobierto.dev  | gobierto | regular user, santander.gobierto.dev as source site, verified against census |
+| Email               | Password | Notes                                                                                     |
+| ---                 | ---      | ---                                                                                       |
+| dennis@gobierto.dev | gobierto | regular user, madrid.gobierto.dev as source site, not verified but with verification item |
+| reed@gobierto.dev   | gobierto | regular user, unconfirmed, madrid.gobierto.dev as source site                             |
+| susan@gobierto.dev  | gobierto | regular user, santander.gobierto.dev as source site, unsuccesfuly verified against census |
 
 Available sites:
 


danguita:
I've just refactored a little this code by doing the following:

- Moved all verification check behavior from the Form object to the corresponding `User::Verification` Model itself.

A couple of UI improvements:

- Show a detailed verification summary view to every signed in user instead of just the last/current one. This way he/she can notice any validation status changes due to data source changes, such as the Census repository.
- Ask for birth date in select format. That UI felt incomplete to me.

I do not plan to make any further changes here so feel free to check it! 🚀

docs/user-namespace.md
```diff
@@ -8,11 +8,11 @@ URL: http://madrid.gobierto.dev/user/login
...
 Available users:
 
-| Email               | Password | Notes                                                         |
-| ---                 | ---      | ---                                                           |
-| dennis@gobierto.dev | gobierto | regular user, madrid.gobierto.dev as source site              |
-| reed@gobierto.dev   | gobierto | regular user, unconfirmed, madrid.gobierto.dev as source site |
-| susan@gobierto.dev  | gobierto | regular user, santander.gobierto.dev as source site           |
+| Email               | Password | Notes                                                                        |
+| ---                 | ---      | ---                                                                          |
+| dennis@gobierto.dev | gobierto | regular user, madrid.gobierto.dev as source site                             |
```
ferblape: Just a stupid thing:

- dennis row should mention he's verified against census
- susan row should mention she's not verified

I can take care of the change.

ferblape:
The rest of the code looks fantastic, very well organized and modularized.

docs/user-namespace.md
danguita: Good catch!

Dennis isn't actually verified against census at User level (the `User#census_verified` attribute) although there would be a census item that will verify him on next tries if that were seeded as well. Anyways I totally agree on being explicit about those seeds so let's mark Dennis and Reed as not verified against census 👍

Issue #73, opened 2016-11-22
Recalculate User verification status after the Census data source changes


danguita:
This comes from https://github.com/PopulateTools/gobierto-dev/issues/56.

PR #77, opened 2016-11-23
Recalculate User verification statuses in background

danguita:
Connects to #73.

### What does this PR do?

This PR implements the last step on the Census data importing process. If everything goes well – any old records have been replaced by new ones– the process triggers a User verification statuses recalculation that, due to its nature, can be executed in background so the main thread remains available for coming requests.

It also configures the background processing queue adapters for each environment. We'll be using the `async` (default) adapter in `development` and `sidekiq` in production. To do so, it is installing and setting the Sidekiq gem, and adding it to the development environment dependencies through Docker Compose. After running `$ docker-compose up` the updated environment should just be up and running.

The best way to try it out is to load some Census data that will be matched by a Verification request from the User namespace. After that, do a new load of data without the previously matching item and check out your latest User verification is not verified anymore.

### DIFF
--- a/app/forms/admin/census_import_form.rb
+++ b/app/forms/admin/census_import_form.rb
@@ -20,7 +20,11 @@ class Admin::CensusImportForm
       complete_import
     end
 
-    census_import.reload.completed?
+    if (result = census_import.reload.completed?)
+      recalculate_user_verifications
+    end
+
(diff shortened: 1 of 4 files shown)


ferblape:
LGTM 👍

Issue #76, opened 2016-11-23
Change User's source site once verified against census data


danguita:
This comes from #56.

PR #78, opened 2016-11-23
Set UserVerification’s site as User’s source site when it succeeds

danguita:
Connects to #76.

### What does this PR do?

This PR implements the User's Source Site switch when a Verification process against Census data succeeds.

Digging a little deeper, it also stabilizes the test fixtures by making then more consistent and increases test coverage.

This can't be tested in the UI actually because the Source Site attribute is just for internal purposes. Anyways, the main thing here is that the User's source site should be updated in database when verified.


ferblape:
👍

Issue #38, opened 2016-11-11
Autocomplete site locations


ferblape:
Site entity association will be available, for the moment, just for municipalities.

To get the list of municipalities we are going to use The Big Indexer municipalities collection.

The endpoint to obtain the suggestions is to a given query is `https://tbi.populate.tools/gobierto/collections/c-municipios-espanya/suggestions`.

It requires a parameter `query` and an authorization header:

- `query`: contains the query argument to the suggestions.
- `authorization: Bearer <TOKEN_BODY>`

(I'll provide the token body in private). Maybe it'll make sense to set that token as an installation setting, because I suspect we are going to use The Big Indexer in some places in the future.

In javascript you can do this:

```
var settings = {
  "crossDomain": true,
  "url": "https://tbi.populate.tools/gobierto/collections/c-municipios-espanya/suggestions",
  "method": "GET",
  "dataType": "json",
  "data": { "query": "foo"},
  "headers": {
    "authorization": "Bearer <TOKEN_BODY>"
  }
}

$.ajax(settings).done(function(response) {
  console.log(response);
});
```

The output of the endpoint is an object with a key `suggestions` and all the suggestions, composed of a value, and a raw data object:

```json
{
"suggestions": [
  {
    "value": "Abarán",
    "data": {
      "id": "30002",
      "name": "Abarán",
      "slug": "abaran",
      "municipality_id": 30002,
      "province_id": 30,
      "autonomous_region_id": 14
    }
  },
  {
    "value": "Abárzuza/Abartzuza",
    "data": {
      "id": "31002",
      "name": "Abárzuza/Abartzuza",
      "slug": "abarzuza-abartzuza",
      "municipality_id": 31002,
      "province_id": 31,
      "autonomous_region_id": 15
    }
  },
  {
    "value": "Abarca de Campos",
    "data": {
      "id": "34001",
      "name": "Abarca de Campos",
      "slug": "abarca-de-campos",
      "municipality_id": 34001,
      "province_id": 34,
      "autonomous_region_id": 7
    }
  }
]}
```

ferblape:
Issue updated @danguita

danguita:
Looking good, let's go for it 👍

PR #80, opened 2016-11-24
Site location autocomplete control against municipalities endpoint

danguita:
Connects to #38.

### What does this PR do?

First of all, it applies a complete cleanup on the Admin namespace layout by creating specific manifest files that don't load any (hopefully) unused vendors or components. Right now there's some duplication in user (default) and Admin namespaces so it's quite difficult to detect what is actually being used in each one. I'd say it's a good starting point to keep things clean from now on.

Secondly, it implements a Site location selection user control based on jquery-ui's autocomplete component. It performs a request to an external endpoint that provides a structured source of municipalities to choose from. As part of the user control implementation, the form now sets the `Site#municipality_id` attribute to keep an internal reference to the selected municipality regardless the location field content.

The Site location autocomplete user control should be available in both new and edit actions, so just check that the location field is being autocompleted in http://gobierto.dev/admin/sites/new and any of the `http://gobierto.dev/admin/sites/<site_id>/edit` URLs.

Regarding the Admin cleanup, it'd be great to do a quick tour on every Admin section to be sure everything is still working. Integration tests are passing but the coverage isn't still specially high BTW.
### Final notes

How about generating a new auth token for the `http://gobierto.dev` Origin so that we can be consistent in our development environments? /cc @ferblape

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,5 +10,5 @@ PRESET_ADMIN_NAME="Gobierto Admin"
 PRESET_ADMIN_PASSWORD=gobierto
 EMAIL_DEFAULT_FROM="Gobierto <admin@gobierto.dev>"
 EMAIL_DEFAULT_REPLY_TO="Gobierto <admin@gobierto.dev>"
-TBI_API_TOKEN=Nu40O9bteWecZz-fpBYEhA
-MUNICIPALITIES_SUGGESTIONS_ENDPOINT=https://tbi.populate.tools/gobierto/collections/c-municipios-espanya/suggestions
+TBI_API_TOKEN=<API TOKEN>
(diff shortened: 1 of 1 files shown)


ferblape:
Agreed, we need a token for the development environment. I'll take care.

danguita:
I've just synced this branch with `master` and recovered some missing assets in the Admin namespace. Sorry for the noise, I wanted to make sure everything works as expected.

ferblape:
Merging this one. I have removed the credentials from the example env file, I'll give you new ones.

danguita:
@ferblape 👌 Go for it! Thanks for taking care.

Issue #82, opened 2016-11-27
Implement Admin namespace as a built-in Gobierto module


danguita:
### Motivation

After having prepared the codebase to accept new modules, it feels like some Admin-related resources are not properly namespaced or could be out of context, so I'd say it's perfect time to refactor the current implementation and make it behave like a built-in module for consistency.

### Check-list

- [x] Do the complete refactor and make tests pass 😛

PR #83, opened 2016-11-27
Refactor the entire Admin namespace as the GobiertoAdmin module

danguita:
Connects to #82.

### What does this PR do?

As described on the corresponding issue, it adds a `gobierto-admin` built-in module in which we're wrapping any Admin related resource, for ease and portability.

This is kind of huge so I'm relying it on automated tests, which are passing now. Anyways, let's tray to do a quick tour through the UI. I'll take care of sharing those results here.


danguita:
I've just added a couple of small improvements. No big deal, but good to have.

danguita:
Merging since nothing controversial here and we want to have these changes available for any coming modules.

PR #86, opened 2016-11-29
Gobierto Budget Consultations module (first iteration)

danguita:
Connects to #79.

### What does this PR do?

This PR introduces the new `gobierto-budget-consultations` module and implements a first iteration in both User (`gobierto-core`) and Admin (`gobierto-admin`) namespaces, mainly focused in back-end and architecture.

#### Gobierto-admin

Having http://gobierto.dev/admin/budgets/consultations/ as the main entry point to this feature, an Admin user can do the following actions:

- [x] If `budget-consultations` module is not enabled, it isn't listed and cannot be accessed in any way.
- [x] Manage Consultations
  - [x] List, Create and Edit.
- [x] Manage every Consultation's Items
  - [x] List (sorting is not yet implemented), Create and Edit.
  - [x] Allow selecting a Budget lines and customizing data such as title and amount.
- [x] Access Consultation responses (that's a pretty basic listing for now).

#### Gobierto-core

Having http://madrid.gobierto.dev/presupuestos/consultas/ as the main entry point to this feature, an User can do the following actions:

- [x] If not signed in, ask for Log in.
- [x] If not verified against Census data (level 2), ask for Verification.
- [x] List Active and Past consultations.
- [x] If Consultation is not active, don't allow participating.
- [x] Show an estimated time of completion.
- [x] Participate in active Consultations
  - [x] List every Consultation's Budget line and allow selecting an option (please note that the current UI is a pretty basic implementation).
  - [x] Allow submitting a participation.
  - [x] Allow going back to the Consultation participation view from Summary view.
  - [x] Allow confirming Consultation results from summary.
  - [x] Access the Consultation done view (final notes, sharing and so on).

PR #87, opened 2016-11-29
Responsive adaptation for budget consultations

furilo:
Connects to #86 

### What does this PR do?

It provides a responsive version of the budget consultations screens. Mostly is CSS, but there is some extra HTML tags and classes here and there.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -23,9 +23,12 @@
 //= require vis_lineas_tabla
 //= require vis_evo_line
 //= require execution
+
 //= require module-admin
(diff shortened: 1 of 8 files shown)


danguita:
👀

danguita:
It seems we have mixed use of tabs and spaces for indentation. It is not a big deal since sandbox files are not strictly part of the codebase, but let's try to normalize this because stylesheets can be integrated just as they are here. How about using spaces everywhere?

Let's move this forward anyways. We can take care of the style normalisation in a new Issue. LGTM 👍

PR #88, opened 2016-11-29
Catalan locale

ferblape:
### What does this PR do?

This PR introduces Catalan locale into Gobierto budgets, which is the section we need in Catalan for a small project.

It also introduces an important change: budget categories are read now from Populate Data index, because we have versions for the different locales.

It can'tt be tested yet, as we don't have a way to set a locale for a Site, or a way of selecting the locale in the UI.

### DIFF
--- a/lib/tasks/gobierto_budgets/fixtures.rake
+++ b/lib/tasks/gobierto_budgets/fixtures.rake
@@ -47,6 +47,7 @@ namespace :gobierto_budgets do
 
       budgets_for_place = BUDGETS_INDEXES.map do |index|
         categories_fixtures do |category|
+          category.merge!('kind' => category['kind'] == 'income' ? 'I' : 'G')
           {
             index: {
               _index: index,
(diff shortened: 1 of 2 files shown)


lib/tasks/gobierto_budgets/fixtures.rake
```diff
@@ -114,6 +114,8 @@ namespace :gobierto_budgets do
     def import_categories
       puts "== Importing categories =="
       categories = categories_fixtures do |category|
+        category.merge!('kind' => category['kind'] == 'I' ? 'income' : 'expense')
```
danguita: Regarding this same change, I'd mention that a complete `script/setup` is required in order to keep things working. Not big deal though.

danguita:
Can't say no to the caganer! 😄

config/environments/development.rb
```diff
@@ -46,7 +46,7 @@
   config.assets.quiet = true
 
   # Raises error for missing translations
-  # config.action_view.raise_on_missing_translations = true
+  config.action_view.raise_on_missing_translations = true
```
ferblape: 👍

lib/tasks/gobierto_budgets/fixtures.rake
ferblape: Makes sense, I have moved the change to the categories.yml, but we still need the mapping, because whe use categories.yml to create the budget lines.

danguita: Nice, this way it looks more consistent to me.

Issue #91, opened 2016-12-04
Retrieve budget lines from TBI when managing Budget Consultations


danguita:
TBD

PR #98, opened 2016-12-05
Populate Data API wrapper and integration

danguita:
Connects to #91.

### What does this PR do?

This PR introduces the `BudgetLineCollectionBuilder` service object, which is responsible of retrieving and building Budget line records for the passed `Site`, through the Populate Data (a.k.a. The Big Indexer) service. To do so, we're also introducing the `PopulateData::Gobierto` API wrapper, which is just an abstraction to request and use resources with ease.

Finally, it replaces some `TODO`s around there regarding this matter. Before today we were just fetching hardcoded data while working in this feature.

Want to mention a detail in the implementation that will probably be updated later today: Regarding the Populate Data wrapper classes, the `PopulateData::Gobierto::Entity` one is now taking the `location_name` param into account and that was not the expected behavior. We plan to just drop it after doing some maintenance on the ES indices.

The `BudgetLineCollectionBuilder` service is now taking part of the Consultation item management actions, so you can check that the select user control in properly populated in both create and update actions:

- `http://gobierto.dev/admin/budgets/consultations/<consultation_id>/items/new` 
- `http://gobierto.dev/admin/budgets/consultations/<consultation_id>/items/<consultation_item_id>/edit`

Note that the select input has a custom option formatting, as follows:

screen shot 2016-12-05 at 5 26 36 pm
 
And most of the form fields are auto-filled when one of the budget lines is selected.

### DIFF
--- a/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb
+++ b/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb
@@ -44,8 +44,7 @@ module GobiertoAdmin
 
       def entities
         @entities ||= PopulateData::Gobierto::Entity.new(
-          municipality_id: @municipality_id,
-          location_name: @location_name
+          municipality_id: @municipality_id
         ).fetch
       end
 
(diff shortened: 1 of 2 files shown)


ferblape:
I have just updated the entity filters to use only the municipality_id. When we add request filtering in Populate data we'll be able to fetch the entities faster.

danguita:
@ferblape Thanks! Let me push a couple of small adjustments to keep consistency across the API wrapper classes.

danguita:
@ferblape Here you go: https://github.com/PopulateTools/gobierto-dev/pull/104. Not a big deal at all, I'm just about to merge it.

Issue #74, opened 2016-11-22
Budget categories localization


ferblape:
Budget categories need to be localized. In order to achieve that we need to follow the follwing steps:

1 - use a collection from the big indexer, instead of a local index

2 - parameterize the collection depending on the site locale (each locale has a collection)

PR #105, opened 2016-12-06
Budget description localization

ferblape:
Connects to #74

### What does this PR do?

This PR introduces the concept of site locale, as a configuration parameter. This is the minimum we need for the moment (no UI yet).

It also sets a controller filter to change the locale based on the site configuration.

And, of course, parameterizes budget categories description based on the locale, to show the correct description for the current locale.

You should set a locale in the site configuration:

```
s = Site.first
s.configuration.locale = :ca`
s.save!
```

And then go to the public URL of the site, and you should see the budgets and budget descriptions in catalan.


danguita:
👍

Issue #93, opened 2016-12-04
Improve Consultation response UI


danguita:
Make it look and behave like the prototype.

PR #111, opened 2016-12-07
Implement Consultation response UI

danguita:
Connects to #93.

### What does this PR do?

This PR implements the Consultation response UI, which consists on a client-side component that is managing the Consultation response state and some styles and markup that were intentionally left aside on the first iteration.

Well, now the Consultation response UI should look and behave like the prototype. That means showing a Consultation item (Budget line) at a time, allowing forward navigation until the Consultation gets completed, and allowing to switch between Context and Choice tabs.

Also, it should keep the current Budget state according to the User choices per Budget line. So, the "Con tus decisiones" budget amount should be updated. In the special case that the User budget exceeds the original one, a deficit warning will be shown.

screen shot 2016-12-07 at 11 15 15 am

screen shot 2016-12-07 at 11 15 07 am

Finally, right after having submitted the Consultation response form, the User should have been redirected to the summary view, showing the same Budget amount and keeping the deficit warning it if exceeds the original. This time the User's budget is calculated in the back-end, but it's just applying the same rules so the values must not differ.

### Extra 🎱 

Dealing with the Sandbox files is a bit messy due to the lack of synchronization between base templates such as layouts. As part of this PR we're updating the `SandboxController` to make it compatible to the actual layouts, which are now synced and properly used.

### DIFF
--- a/app/views/gobierto_budget_consultations/consultations/consultation_confirmations/show.html.erb
+++ b/app/views/gobierto_budget_consultations/consultations/consultation_confirmations/show.html.erb
@@ -27,7 +27,9 @@
 
           <p>
             También puedes compartir esta URL con quien prefieras:<br/>
-            <%= link_to budget_consultation_participation_url(@consultation_response.sharing_token) %>
+            <%= link_to(
+              budget_consultation_participation_url(@consultation_response.sharing_token),
+              budget_consultation_participation_url(@consultation_response.sharing_token)) %>
           </p>
         </div>
 


ferblape:
Hey @danguita the PR looks great, but I found a small bug in the confirmation screen, because in the link to share, with the anonymous token, the `href` attribute is the current URL, not the anonymous link.

screen shot 2016-12-07 at 13 17 43

Do you want to fix it in this PR or should I open an issue?

danguita:
@ferblape Good catch! Let me fix it right here.

danguita:
@ferblape Ready to check.

ferblape:
🚀

Issue #112, opened 2016-12-07
Get rid off translated routes


ferblape:
We have decided to keep routes static, without localization.

PR #113, opened 2016-12-07
Remove translated routes

ferblape:
Connects to #112

### What does this PR do?

This PR removes translated routes support. In the public resources, we have picked Spanish as main language.

If you visit budgets visualizations or budgets consultations you'll see the routes main segments in Spanish.

Routes paths are defined in Spanish for the time being


danguita:
Post-merge 👍

Issue #92, opened 2016-12-04
Improve Consultation items management UI


danguita:
Make it look and behave like the prototype.

Check-list:

- [x] Sortable Consultation items list
- [x] Manage Consultation items in modals
- [x] Autocomplete Budget lines

danguita:
Finally moving this one to Acceptance.

danguita:
Yay!

PR #116, opened 2016-12-07
Implement Sortable list behavior in Consultation items management UI

danguita:
Connects to #92.

### What does this PR do?

This PR comes as part of the batch of UI enhancements that need to be performed regarding the Consultation items management screen.

Basically, it implements the Sortable list behavior using native HTML5 drag & drop API through the [html5sortable](https://github.com/voidberg/html5sortable) library.

To do so, it also implements a back-end component to abstract a sorting operation via a `ActionController::Parameters` struct.

Both the UI and the back-end components are ready to be just attached to any other entities across the system so this Sortable behavior should be easily repeatable.

Regarding the Consultation items (Budget lines) management screen, just check that items can be sorted via drag & drop and the new positions get persisted in database: `http://gobierto.dev/admin/budgets/consultations/<consultation_id>/items`.

screen shot 2016-12-07 at 5 04 54 pm

In addition to that, let's check that the Budget lines are shown in the right order when Responding to Consultation items right at `http://madrid.gobierto.dev/presupuestos/consultas/<consultation_id>/participar`.

### DIFF
--- a/app/assets/javascripts/admin/app/controllers/consultation_items_controller.js
+++ b/app/assets/javascripts/admin/app/controllers/consultation_items_controller.js
@@ -1,5 +1,13 @@
-this.App.ConsultationItemsController = (function() {
+this.GobiertoAdmin.ConsultationItemsController = (function() {
   function ConsultationItemsController() {}
+
+  ConsultationItemsController.prototype.edit = function() {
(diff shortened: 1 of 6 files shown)


ferblape:
👍

PR #134, opened 2016-12-09
Consultations feature feedback preparations

danguita:
### What does this PR do?

This PR acts as the starting point for the second iteration by doing some preparations on it. It sets up a more standard routing strategy and increases test coverage.


danguita:
Merging this one since it is just preparing the second iteration.

Issue #117, opened 2016-12-08
Skip consultations home if there´s only one active consultation


ferblape:
If there's only one **active** consultation, skip the consultations home page and redirect to it.

PR #135, opened 2016-12-09
Implement single active Consultation behavior

danguita:
Connects to #117.

### What does this PR do?

This PR implements a special behavior in the Consultations index when there's only one active Consultation. In that case we're skipping the index view and moving forward to that Consultation's view.

Check that, when only one active Consultation is present, requesting http://madrid.gobierto.dev/presupuestos/consultas/ redirects to `http://madrid.gobierto.dev/presupuestos/consultas/<consultation_id>`.


ferblape:
👍 works as a charm

Issue #120, opened 2016-12-08
Redirect after a consultation is saved


ferblape:
After a consultation is saved, we should keep the admin in the edit action instead of redirecting her to the consultations list.

In general, we should apply this pattern to all edit actions in the Gobierto admin, I'll create a separated issue to review this.

PR #139, opened 2016-12-12
Update Consultation management workflow

danguita:
Connects to #120.

### What does this PR do?

This PR slightly updates the Consultation management workflow by redirecting back to the edit action when performing an update request.

Digging a bit deeper, it also prepares the testing environment for Rails 5.0+ Controller testing, which is now based in `ActionDispatch::IntegrationTest`. To do so, we're adding a bunch of helper methods to let managing both Admin and User sessions. They are just available by inheriting your controller tests from the `GobiertoControllerTest` class.

This could be tested by submitting a Consultation update and checking that the request gets redirected to the same Consultation edit URL showing the proper flash message.

i.e.: `http://gobierto.dev/admin/budgets/consultations/<consultation_id>/edit`.


ferblape:
👍 works as expected

Issue #140, opened 2016-12-12
Exception using BudgeLineCollectionBuilder class


ferblape:
This is an exception happenning in staging [see Rollbar item](https://rollbar.com/fernando/gobierto/items/255/)

This is the traceback:

```
NameError: uninitialized constant #<Class:0x007f7dc48df920>::PopulateData
  File "/var/www/gobierto_staging/releases/20161209171523/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb", line 45, in entities
  File "/var/www/gobierto_staging/releases/20161209171523/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb", line 11, in block in call
  File "/var/www/gobierto_staging/releases/20161209171523/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb", line 10, in tap
  File "/var/www/gobierto_staging/releases/20161209171523/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb", line 10, in call
  File "/var/www/gobierto_staging/releases/20161209171523/app/controllers/gobierto_admin/gobierto_budget_consultations/consultations/consultation_items_controller.rb", line 77, in get_budget_lines
  File "/var/www/gobierto_staging/releases/20161209171523/app/controllers/gobierto_admin/gobierto_budget_consultations/consultations/consultation_items_controller.rb", line 20, in new
  File "/var/www/gobierto_staging/shared/bundle/ruby/2.3.0/gems/actionpack-5.0.0.1/lib/action_controller/metal/basic_implicit_render.rb", line 4, in send_action
  File "/var/www/gobierto_staging/shared/bundle/ruby/2.3.0/gems/actionpack-5.0.0.1/lib/abstract_controller/base.rb", line 188, in process_action
  File "/var/www/gobierto_staging/shared/bundle/ruby/2.3.0/gems/actionpack-5.0.0.1/lib/action_controller/metal/rendering.rb", line 30, in process_action
  File "/var/www/gobierto_staging/shared/bundle/ruby/2.3.0/gems/actionpack-5.0.0.1/lib/abstract_controller/callbacks.rb", line 20, in block in process_action
  File "/var/www/gobierto_staging/shared/bundle/ruby/2.3.0/gems/activesupport-5.0.0.1/lib/active_support/callbacks.rb", line 126, in call
[...]

PR #141, opened 2016-12-12
Make `PopulateData` lib usage explicit to deal with autoloading issues

danguita:
Connects to #140.

### What does this PR do?

This is is likely a class autoloading problem in staging/production. Even when everything under `/app` is autoloaded in development and eager loaded in staging/production, I'd keep this library out of the app context so we're making the `PopulateData::Gobierto` class usage explicit. It is now needed to `require "populate_data"` when making use of that library.

Ensure that the `PopulateData::Gobierto` is properly loaded from `GobiertoAdmin::GobiertoBudgetConsultations::BudgetLineCollectionBuilder` in an eager loaded environment config such as `staging` or `production`.

It is only used in the Consultation Item management action so let's check this form still works:

`http://gobierto.dev/admin/budgets/consultations/<consultation_id>/items/<consultation_item_id>/edit`.

### DIFF
--- a/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb
+++ b/app/services/gobierto_admin/gobierto_budget_consultations/budget_line_collection_builder.rb
@@ -1,3 +1,5 @@
+require "populate_data"
+
 module GobiertoAdmin
   module GobiertoBudgetConsultations
     class BudgetLineCollectionBuilder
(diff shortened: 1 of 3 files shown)


ferblape:
Let´s try it in staging. Thank you!

danguita:
Thanks man! 🙌

ferblape:
It works!

Issue #119, opened 2016-12-08
Budget line amounts


ferblape:
In the consultation item page, when a budget line is selected we should simplify the amounts displayed:

- [x] add the unit
- [x] round the thousands (i.e: 124.000€ instead of 123.567€)
- [x] avoid milions (i.e. 123M€; 1,24M€; 123.000€)

danguita:
This one depends on https://github.com/PopulateTools/gobierto-dev/pull/130. Let's keep in back-log for now.

danguita:
Nevermind, we can move it forward.

PR #143, opened 2016-12-12
Consultation's budget amounts formatting

danguita:
Connects to #119.

### What does this PR do?

This PR normalizes the budget amounts format across different sections on the UI, taking care of both server and client sides.

In the server side we're using a helper method to wrap the number formatting, while in client side we're propagating the same format options through a numeric format library which is included as a `GobiertoBudgetConsultations` specific vendor.

Let's check that any budget amount in the Consultation response process match the following assertions:

```ruby
assert_equal "124000", budget_amount_to_human(123567)
assert_equal "123000", budget_amount_to_human(123000)
assert_equal "123 M", budget_amount_to_human(123000000)
assert_equal "1.24 M", budget_amount_to_human(1240000)
```

screen shot 2016-12-12 at 12 38 18 pm

### Final notes

There would be slight differences between the server and client-side number formatting, although I would not care about them since there's an enhancement that will drop any repeated logic at https://github.com/PopulateTools/gobierto-dev/issues/133.


ferblape:
👍  Works OK
👍 I agree with your thoughts, no need to worry about that now

Merging!

danguita:
Cool, thanks!

PR #144, opened 2016-12-13
Cover empty parameters edge case in Confirmable behavior

danguita:
Unplanned.

### What does this PR do?

This PR is enhancing the User Confirmable behavior by covering an edge case in which the User can pass a empty `confirmation_token` parameter.

Due to this module nature, it can be applied to both `User` and `GobiertoAdmin::Admin` resources.

Just check that a Confirmation URL with empty or missing `confirmation_token` parameter doesn't match any Users:

- http://gobierto.dev/admin/admin/confirmations?confirmation_token=
- http://gobierto.dev/admin/admin/confirmations


ferblape:
👍

PR #147, opened 2016-12-13
Cover empty parameters edge case in Recoverable behavior

danguita:
Unplanned.

### What does this PR do?

Similar to https://github.com/PopulateTools/gobierto-dev/pull/144, this PR is enhancing the User Recoverable behavior by covering an edge case in which the User can pass a empty `reset_password_token` parameter.

Due to this module nature, it can be applied to both `User` and `GobiertoAdmin::Admin` resources.

It is also refactoring the current fixtures to be more specific about that behavior.

Just check that a Password reset URL with empty or missing `reset_password_token` parameter doesn't match any Users:

- http://gobierto.dev/admin/admin/passwords/edit?reset_password_token=
- http://gobierto.dev/admin/admin/passwords/edit

### DIFF
--- a/app/controllers/gobierto_admin/admin/passwords_controller.rb
+++ b/app/controllers/gobierto_admin/admin/passwords_controller.rb
@@ -22,7 +22,7 @@ module GobiertoAdmin
     end
 
     def edit
-      admin = Admin.find_by(reset_password_token: params[:reset_password_token])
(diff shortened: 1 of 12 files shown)


ferblape:
👍

danguita:
Thanks! 🙌

Issue #99, opened 2016-12-06
Localize Admin UI


danguita:
Since the latest features are pretty much stabilized, cover any missing translations within the `GobiertoAdmin` module's UI.

Also, consider enabling the `config.action_view.raise_on_missing_translations` setting to ensure everything is localized.

ferblape:
Have you ever used https://github.com/glebm/i18n-tasks?

danguita:
Yep, I was thinking of it when I wrote these lines 😛  I meant to enable the `raise_on_missing_translations` setting right after everything gets covered to early detect any new missing translations at run time, though using this tool as part of our CI process could be a good idea as well. Thanks for pointing it out!

PR #152, opened 2016-12-13
Complete Admin UI localization

danguita:
Connects to #99.

### What does this PR do?

This PR runs through the entire Admin localizing every single action, adding missing translations, cleaning up and refactoring small pieces of code when needed.

What is missing?

- Locale selector at Admin level.
- Emails, because we don't have definitive content for them.
- Catalan translation.
- Integration of a static code analysis tool to detect missing translations. I'd do that after getting this one merged.

Let's please talk about that.

Just get into the Admin namespace and check that every single section is translated into Spanish (which is the default locale for now), even when everything's also available in English as the master language for development:

http://gobierto.dev/admin/

### Final notes

That's definitely a big one, thanks for checking 🙏


ferblape:
Checking this one

ferblape:
This looks OK to me. I´ll take care of creating issues for the missing things.

I can't talk now about the missing keys, I'll ping you!

danguita:
Thanks for taking care! No rush, this is not blocking us 👍

Issue #149, opened 2016-12-13
Update Magnific Popup library to its latest version


danguita:
It seems this library is still having some incompatibilities with Safari 8 and 9 versions. Let's upgrade it to [1.1.0](https://github.com/dimsemenov/Magnific-Popup/releases/tag/1.1.0) to investigate them from a clean state.

Just for the record, the error is:

```
Syntax error: - Invalid character '\u0215'
```

@furilo Thanks for reporting!

ferblape:
Picking this one.

PR #158, opened 2016-12-14
Update Magnific Popup to v1.1.0

ferblape:
Connects to #149

### What does this PR do?

This PR updates the library to version 1.1.0

You should check magnific popup works now, for example, by creating or editing a budget line consultation item.


danguita:
Checking this one.

danguita:
It works just fine in Safari `10.0.2 (12602.3.12.0.1)` and Chrome `55.0.2883.95 (64-bit)`. I'd move this one forward to `master` in order to let @furilo check it with his reported version of Safari.

screen shot 2016-12-14 at 16 19 44

So, LGTM 👍

PR #157, opened 2016-12-14
Implement site links as a HTML snippet

ferblape:
Connects to #115 

### What does this PR do?

This PR changes how the links of a site are implemented: instead of an array of links we are implementing a snippet of HTML code.

### How should this be manually tested?

1 - Go to the Admin interface and update the attribute 

2 - Check in the public site that the links are visible

PR #163, opened 2016-12-15
Update Site fixtures to match the current configuration schema

danguita:
Unplanned. Relates to #157.

### What does this PR do?

It just updates the Sites fixtures to make use of the `links_markup` attribute instead of the `links` array. This is good to have since they are used to populate both test and development databases as part of the seeding process.

Execute the seeding process to ensure the `@site.configuration.links_markup` struct has some content.


ferblape:
Good catch!

Issue #126, opened 2016-12-08
Integrate sandbox sign in


ferblape:
Integrate sandbox sign in / sign up screens.

danguita:
Just to keep in mind, it involves both front-end and behavioral changes.

PR #164, opened 2016-12-15
Integrate Confirmation request and Password reset actions into the new session workflow

danguita:
Comes up from #126.

### What does this PR do?

First of all, this was not specified in the prototype but, after checking the entire workflow I missed some actions that would need to be implemented in order to keep a good experience.

This one implements the Confirmation email request as part of the new session workflow. This means that any User should be able to request a new confirmation email in the new session screen, as follows:

screen shot 2016-12-15 at 07 32 11

Clicking in "¿Instrucciones de confirmación?" ⬇️ 

screen shot 2016-12-15 at 07 32 17

Also, we're integrating the User password change behavior into the new front-end, and any alternative URLs have been dropped from the system actions:

screen shot 2016-12-15 at 07 59 45

In addition, any new screens are being localised into the three languages available now, so once again I'd need some help from my beloved Catalan translator @ferblape 🙏 

Just get into the login screen at http://madrid.gobierto.dev/user/login, and check that:

- [x] The Confirmation request form can be reached from the new session block.
- [x] The Confirmation request works as expected, allowing the User to start a new confirmation process through an informative email.

In addition, let's check that the Password reset feature still works as expected and the Password edit form has the right styling:

- [x]  `http://madrid.gobierto.dev/user/passwords/edit?reset_password_token=<reset_password_token>` looks good.

### Help wanted ✋ 

@furilo Does that work for you?


ferblape:
LGTM!

danguita:
Thanks for those translations! 🙌  Waiting for the build to pass.

Issue #124, opened 2016-12-08
User after login redirect


ferblape:
After a user logs in, redirect her to the previous page.

PR #165, opened 2016-12-15
Keep context after User sign in

danguita:
Connects to #124.

### What does this PR do?

This one updates the `after_sign_in_path` behavior at User (`GobiertoCore`) level to take a referrer URL into account. To do so, the `HTTP_REFERER` is passed through the sign in form to keep a context to which we can redirect on success.

Just check that the context is kept after a sign in action. In example, a sign in request from http://madrid.gobierto.dev/presupuestos/ejecucion/2016 should redirect back to the same URL.


ferblape:
Works ok 💪 !

Issue #166, opened 2016-12-15
User Subscriptions feature catch-up


danguita:
- [x] As a User, I want to Subscribe to a topic in the boundaries of Gobierto's data model.
  *e.g. Changes in Consultations*.
- [x] As a User, I want Subscribe to changes in a single record.
  *e.g. Changes in Consultation named "Consulta sobre presupuestos del Ayuntamiento de Madrid"*.
- [x] As a User, I want to see all my active Subscriptions.
- [x] As a User, I want to manage my active Subscriptions, so that I can remove any of them.

PR #171, opened 2016-12-19
User Subscriptions feature starting point

danguita:
Connects to #166.

### What does this PR do?

This PR introduces a new feature which allows any signed in User to subscribe to determined system resources in both generic and specific contexts. To do that, we're adding a new model called `User::Subscription` to the system and a simple interface to access it.

It also allows any not signed in User to subscribe to those exposed resources by starting a new User registration process with the pre-created `User::Subscription` item.

#### Extra 🎱 

- We are preparing the User namespace here by trying to keep the default layout generic and extracting a new one for User namespace only.
- There has been some progress in the User namespace localisation process, and any new implementations are being localised as well.
- We are taking care of the "every-interaction-has-a-form-object" approach by covering any missing scenarios around there, and making those form objects contain most of the model logic instead of the `ActiveRecord` models itself.

As an integration sample, the `GobiertoBudgetConsultations::Consultation` model has been set up as a  the only "subscribable" resource for now, so these are the available user interactions:

**A. When signed in as a User**

The subscription box should be shown as follows:

Generic context URL: http://madrid.gobierto.dev/consultas_presupuestos/

screen shot 2016-12-19 at 07 26 01

Specific context URL: `http://madrid.gobierto.dev/consultas_presupuestos/<consultation_id>`

screen shot 2016-12-19 at 09 02 30

**B. When not signed in**

The subscription box should be shown as follows:

Generic context URL: http://madrid.gobierto.dev/consultas_presupuestos/

screen shot 2016-12-19 at 07 25 56

Specific context URL: `http://madrid.gobierto.dev/consultas_presupuestos/<consultation_id>`

screen shot 2016-12-19 at 09 02 37

#### ✏️ Known improvements

Finally, I'm keeping some known improvements here for future iterations:

- Detect when a signed in user has already subscribed to the current resource, and show an "Unsubscribe from this" button instead.
- Perform any subscribe/unsubscribe actions asynchronously to avoid the annoying redirections.


ferblape:
Looks very promising! Merging after the catalan magic.

danguita:
Yo, thanks for that! 👏

danguita:
The build has just passed in `master` branch. I'm rebuilding this one's for mental sanity.

Issue #170, opened 2016-12-16
Create a S3 bucket to store uploaded assets


danguita:
This comes from https://github.com/PopulateTools/gobierto-dev/pull/169.

Since we're now uploading assets to a S3 bucket and serving them directly from there, it'd be needed to create a new one with a `s3:GetObject` policy and provision the corresponding credentials at app instance level.

Here is a bucket policy example:

```
{
  "Version":"2012-10-17",
  "Statement":[
    {
      "Sid":"AddPerm",
      "Effect":"Allow",
      "Principal": "*",
      "Action":["s3:GetObject"],
      "Resource":["arn:aws:s3:::gobierto-dev/*"]
    }
  ]
}
```

Once the bucket is available, these are the ENV vars that we would need to provision:

```
AWS_ACCESS_KEY_ID=
AWS_SECRET_ACCESS_KEY=
AWS_REGION=eu-west-1
S3_BUCKET_NAME=gobierto-dev
```

danguita:
Just to keep in mind, let's revert [this commit](https://github.com/PopulateTools/gobierto-dev/pull/169/commits/371580f10447621ff92bcc18665a64d5dd12830a) if everything goes well after applying the bucket policy.

ferblape:
Hey @danguita, when I create the policy I get a validation error like this: `This policy contains the following error: Has prohibited field Principal For more information about the IAM policy grammar, see AWS IAM Policies.`

Does it rings a bell to you? Because I can't find any information in Google nor in the documentation.

danguita:
Nope, that just works well for me 😕

Are you adding anything else, such as IP conditions?

Check out the sample policy from which that one comes in AWS docs: https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-2.

This is nothing new, but here you go a screenshot from my current config:

![screen shot 2016-12-19 at 7 39 16 pm](https://cloud.githubusercontent.com/assets/126392/21324682/1a6095d2-c623-11e6-8a3b-720e7de2c348.jpg)

Hope it helps!

PR #172, opened 2016-12-20
Remove explicit ACL setting

ferblape:
Connects to #170

### What does this PR do?

This PR removes the ACL setting commit from the other day. We have also created the S3 buckets for staging and production.


ferblape:
Merging!

danguita:
Post 👍!

Issue #167, opened 2016-12-15
User Notifications feature catch-up


danguita:
- [x] As a User, I want to configure the periodicity for receiving Notification related communications.
- [x] As a User, I want to receive Notification related communications in my email address, taking the configured periodicity into account.
- [x] As a User, I want to be able to disable all email communications.
- [x] As a User, I want to see all my Notifications in a single page, as an Activity feed.

danguita:
Extracted some user stories from this Issue to https://github.com/PopulateTools/gobierto-dev/issues/189.

danguita:
All done, closing this one!

PR #173, opened 2016-12-20
GobiertoCommon::Trackable extension and User Notifications feature starting point

danguita:
Connects to #167.

### What does this PR do?

This PR implements an extension library that allows any of the `Form` objects in the system to track changes on its managed resources. Those changes also produce events to be consumed through our pub/sub implementation.

To cover a complete first iteration, the `GobiertoCommon::Trackable` extension is being integrated into the `GobiertoAdmin::GobiertoBudgetConsultations::ConsultationForm` class so that we're producing events that are consumed by a custom Subscriber.

It also adds the `User::Notification` implementation which is the first resource populated through events within the `trackable` namespace. The logic for building those `User::Notification` records based on User Subscriptions is abstracted in the `User::NotificationBuilder` service class.

Some other details on the implementation can be found in the doc that comes with this PR: https://github.com/PopulateTools/gobierto-dev/blob/fb34ccd159c470238d3bc82fe0e0f2601f963a50/docs/trackable-extension.md

Since the `User::Notification` implementation is already there, a starting User notifications view can be reached at http://madrid.gobierto.dev/user/notifications. It is being populated through changes in the `GobiertoBudgetConsultations::Consultation` model which the user is subscribed to (see https://github.com/PopulateTools/gobierto-dev/pull/171).

#### ✏️ Known improvements

- The `User::Notification` population process should be done asynchronously, in a background job.
- The User's activity/alerts view should be formatted properly.
- It would be great to have some end-to-end tests regarding changes in Trackable Form objects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ ___this is a development version, not ready for production___
 - [Setting up the development environment](docs/development-environment.md)
 - [Accessing the Admin namespace](docs/admin-namespace.md)
 - [Accessing the User namespace](docs/user-namespace.md)
+- [Integrating the Trackable extension](docs/trackable-extension.md)
 
 ## License
 
--- a/app/controllers/user/notifications_controller.rb
+++ b/app/controllers/user/notifications_controller.rb
@@ -1,0 +1,16 @@
+class User::NotificationsController < User::BaseController
+  before_action :authenticate_user!
+
+  def index
+    @user_notifications = find_user_notifications.sorted
+  end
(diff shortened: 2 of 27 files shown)


ferblape:
It's a very good starting point. In the next steps we´ll also adjust the kind of notifications per resource.

danguita:
Yep, the idea was to develop the tools here and set them up later 👌 Thanks mate.

PR #179, opened 2016-12-20
User Notification mailers starting point

danguita:
Connects to #167.

### What does this PR do?

This one implements the first approach for delivering User Notification-related emails for each notification type.

We're adding mailer templates per event name (e.g. `created`, `title_changed`, `visibility_level_changed`...) so that any of them can be completely customised. Also, we can customise any email subject through the localisation files.

That's slightly out of context, but it also adds some enhancements in the `User::NotificationBuilder` service class, and wraps any `User::Notification` database accesses in a well-known `User::NotificationForm` class.

In addition, any new messages are being localised into the three languages available now, so once again I'd need some help from my beloved Catalan translator @ferblape 🙏 

At this time we can check that changing any watched attribute in a Consultation, such as the title, an event is produced and immediately consumed by the subscriber, creating a new `User::Notification` record and delivering the corresponding email. That email should have a consistent subject and body, that are just ready to be localised when we get the definitive texts.

Please note that the notification digest mailer is implemented too, but it isn't reached actually for now.

Don't forget about the User alerts panel, which is still there showing these notification records in the same format: http://madrid.gobierto.dev/user/notifications.

### DIFF
--- a/app/forms/user/notification_form.rb
+++ b/app/forms/user/notification_form.rb
@@ -28,7 +28,7 @@ class User::NotificationForm
   end
 
   def run_callbacks
-    true
+    deliver_notification_email
   end
 
   def create_notification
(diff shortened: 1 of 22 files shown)


ferblape:
🚀 looking good!

PR #181, opened 2016-12-21
User Notification Digest Agent and common interface

danguita:
Connects to #167.

### What does this PR do?

Taking the User preferences from https://github.com/PopulateTools/gobierto-dev/pull/180 into account, this PR provides a common interface to reach the `User::NotificationDigestAgent` service, which is in charge of collecting all User Notifications ready to be sent in a determined period, and deliver them in form of a digest email.

The interface to access this service from outside the application is a Rake interface, and it is defined as follows:

```shell
rake user:notification_digest_agent:daily           # Delivers daily notifi...
rake user:notification_digest_agent:hourly          # Delivers hourly notif...
rake user:notification_digest_agent:weekly          # Delivers weekly notif...
```

So, it would be needed to schedule all these Rake tasks as cron tasks to be executed in the same frequency to keep working the User Notification delivery mechanism.

It also updates the current implementation to make use of the User notification frequency preferences in combination to the new User Notification sent flag.

After seeding the database, just try the `hourly` Rake task and wait for a digest email to come:

```shell
$ bundle exec rake user:notification_digest_agent:hourly
```

Run it twice to check that the email is not being delivered this time.

### DIFF
--- a/app/forms/user/notification_form.rb
+++ b/app/forms/user/notification_form.rb
@@ -14,25 +14,29 @@ class User::NotificationForm
   def save
     return unless valid?
 
-    run_callbacks if create_notification
+    run_callbacks if save_user_notification
   end
-
-  def record
-    user_notification
-  end
-
-  private
 
   def user_notification
     @user_notification ||= User::Notification.new
   end
+  alias record user_notification
 
-  def run_callbacks
-    deliver_notification_email
+  private
+
+  def user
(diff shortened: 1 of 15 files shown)


ferblape:
I like to use cron frequency to control the frequency of the delivery.

danguita:
Nice, thanks man 👌

Issue #108, opened 2016-12-06
Extract User-specific logic from ApplicationController


danguita:
As seen today in the Staging environment, some User namespace-specific logic is defined at `ApplicationController` level so any other controllers are inheriting them.

In example, the `authenticate_user_in_site` callback should not be triggered in other namespaces but the User one.

Let's just write it down for a possible refactor.

ferblape:
Picking this one, which is getting a bit annoying in staging.

PR #182, opened 2016-12-21
Prevent user authentication in gobierto admin module + a bit of cleanup

ferblape:
Connects to #108

### What does this PR do?

This PR fixes a problem in gobierto admin module when the current site has password protection enabled. To keep the simplicity and because most of the modules will be public, the preferred solution for the time being is to skip the filter in those modules that don't require http web authentication.

1. Open a private browser session
2. Go to gobierto.dev/admin and login
3. Protect with password all the sites
4. Logout
5. Open a private browser session
6. Go to gobierto.dev/admin and login
7. You shouldn't be asked for a http password

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
   end
 
   def current_site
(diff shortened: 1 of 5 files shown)


danguita:
LGTM. That refactor is very welcomed 👍

PR #186, opened 2016-12-21
ActiveNotifier service and common interface

danguita:
Connects to #167.

### What does this PR do?

This one implements a mechanism to seek for events that should be actively notified to the system, and wraps it within the `GobiertoCommon::ActiveNotifier` service. Those are events that can't be driven by the application workflow itself, so they need to be triggered explicitly.

Here we're integrating some `GobiertoBudgetConsultations::Consultation` events to continue with our sample case. These are: `opened`, `closed` and `about_to_close`. Messages and mailer templates have been added as well.

The interface to access this service from outside the application is a Rake interface too, and it is defined as follows:

```shell
rake common:active_notifier:daily
```

Due to the nature of the currently watched events, only a daily performed task is needed. It is expected to define interfaces in other kind of frequencies or contexts though.

Just run the only rake task here to see how events are being published and consumed, so the corresponding `User::Notification` entries should be created in this case:

```shell
$ rake common:active_notifier:daily
```

### DIFF
--- a/app/models/gobierto_budget_consultations/consultation.rb
+++ b/app/models/gobierto_budget_consultations/consultation.rb
@@ -15,11 +15,15 @@ module GobiertoBudgetConsultations
 
     enum visibility_level: { draft: 0, active: 1 }
 
-    scope :sorted,    -> { order(created_at: :desc) }
-    scope :active,    -> { select(&:open?) }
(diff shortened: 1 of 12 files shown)


ferblape:
👍 lgtm!

PR #187, opened 2016-12-22
User Notifications seen control

danguita:
Connects to #167.

### What does this PR do?

This one adds a new status flag at `User::Notification` level to control which of them have been already seen by the User so that they aren't shown in coming requests.

To do so, we're implementing a simple API to deal with this state and a callback that is triggered just after retrieving the unseen Notifications list in the User UI.

In addition, we're refactoring a little other flags' implementation to be consistent in our code. Test coverage has been increased as well.

#### Reviewer notes

A new localisation string has been added in all three available languages, but it would be great if we could confirm that the Catalan translation is correct.

After reseeding the database, just get into http://madrid.gobierto.dev/user/notifications to see a couple new notifications there, and refresh the page to check they are replaced by a proper message.

### DIFF
--- a/config/locales/user/views/notifications/ca.yml
+++ b/config/locales/user/views/notifications/ca.yml
@@ -4,6 +4,7 @@ ca:
     notifications:
       index:
         title: Activitat recent
+        no_new_notifications: No hi ha notificacions noves
         headers:
           created: Creat
           subject: Asumpte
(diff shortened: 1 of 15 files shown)


config/locales/user/views/notifications/ca.yml
```diff
@@ -4,6 +4,7 @@ ca:
     notifications:
       index:
         title: Activitat recent
+        no_new_notifications: No hi ha notificacions noves
```
ferblape, line 7: 👍

ferblape:
Works like a charm!

PR #196, opened 2016-12-23
Javascript reorganization

ferblape:
### What does this PR do?

This PR reorganizes the javascript according to the dicussions we had yesterday with @furilo and @danguita in a solution of compromise, which is summarized in the following points:

- we are going to separate specific javascript in a separated manifest for each module, including the admin
- we are going to try to have an unique vendor, for the frontend modules and the admin module
- the layouts will load the common manifest and the specialized manifest
- the gobierto budgets javascript was a mess
- if we have common behaviour, such as modals or tabs navigation, we'll put it in the main manifesto. This manifesto will be clean of specific behaviours.

Good parts:

- we were going to share most of the vendor files between modules and admin, so let's just create a common manifesto
- we'll have a separate file per module, which is a separate request, but I prefer that in favour of code organization

So, what this PR does is:

- updates all the layouts to load the custom manifest and the general manifest
- removes all the separated vendors except the gobierto-budgets-consultations, because it requires a bit of coding
- extracts gobierto budgets javascript into its own module

I'm not going to merge it yet, I'll wait probably until Monday.


danguita:
Nice job, thanks for taking care! I agree this approach is quite a bit simpler overall. I would just keep an eye on the common manifest file, in case it keeps growing or holding very different kind of vendors 👌 

Regarding the `gobierto-budget-consultations` refactoring, that carousel plugin (`slick.js`) is not being used actually so we can get rid of it in case we don't want it in the current implementation. I can handle that anyways.

Having said that, let me have a look at that and come back to you! 🎄

app/views/gobierto_budgets/layouts/application.html.erb
```diff
@@ -1,4 +1,8 @@
 <% content_for :javascript_include do %>
+  <%= javascript_include_tag 'gobierto_budgets/application', 'data-turbolinks-track' => true %>
+<% end %>
+
+<% content_for :javascript_include do %>
```
danguita: Maybe we should keep only one of the `:javascript_include` yield blocks here.

app/assets/javascripts/application.js
```diff
@@ -74,67 +57,25 @@ $(document).on('turbolinks:load', function() {
...
-    groupBy: 'category'
-  };
 
-  $autocomplete.autocomplete($.extend({}, AUTOCOMPLETE_DEFAULTS, searchOptions));
+  $('.close_modal').click(function(e) {
+    $.magnificPopup.close();
+  });
 
+  // TODO: move to gobierto budget consultations
   $('.carousel').slick({
```
danguita, line 79: As said on the comment, this one is not being used actually. What is your plan here?

danguita (approved):

app/views/gobierto_budgets/layouts/application.html.erb
ferblape: 👍

app/assets/javascripts/application.js
ferblape: Yeah, it's not implemented yet, but it's used in the sandbox template. So I'll wait until we finish the budget consultations module to remove it or move it to another part.

danguita: Alright, that makes sense 👍

PR #207, opened 2016-12-27
Validate record creation through Admin Invitations UI

danguita:
Connects to #202.

### What does this PR do?

This PR fixes a bug when requesting an Admin invitation with an invalid email address. That is when the email address is already present in database due to the uniqueness email constraint.

Also, we're improving the Admin invitations UI by displaying which of the email addresses couldn't be delivered for some reason.

Get into the Admin invitations UI at http://gobierto.dev/admin/admin/invitations, and mix up valid and invalid email addresses in the form input. It is expected to work in a non transactional way, so any invitations with valid email addresses should be delivered, and process should return successfully by printing any not delivered email addresses in a flash message:

screen shot 2016-12-27 at 6 45 38 pm

### DIFF
--- a/config/locales/gobierto_admin/controllers/admin_invitations/ca.yml
+++ b/config/locales/gobierto_admin/controllers/admin_invitations/ca.yml
@@ -5,4 +5,5 @@ ca:
       invitations:
         create:
           success: Les invitacions s'han enviat correctament
+          success_with_errors: "Algunes invitacions no han pogut ser enviades: %{email_addresses}."
           error: Hi hagué un problema al enviar les invitacions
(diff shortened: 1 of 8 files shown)


config/locales/gobierto_admin/controllers/admin_invitations/ca.yml
```diff
@@ -5,4 +5,5 @@ ca:
       invitations:
         create:
           success: Les invitacions s'han enviat correctament
+          success_with_errors: "Algunes invitacions no han pogut ser enviades: %{email_addresses}."
```
ferblape, line 8: 👏

ferblape:
👍 thanks!

PR #194, opened 2016-12-23
Gobierto People module (first iteration)

danguita:
Connects to #190.

### Review notes

I'm not deploying this code to Staging since we're taking advantage of the PostgreSQL JSONB data type to store and query unstructured data, so we would need to upgrade our PostgreSQL instances to 9.4+ to release these changes.

Please note that we were using PostgreSQL 9.6.1 in development env (the latest release available at that time) until we decide the version to go in production. Let's take it into account to keep exactly the same version in every environment. i.e. production, staging, development and test.

---

### What does this PR do?

This PR introduces the new `gobierto-people` module and implements a first iteration in the Admin (`gobierto-admin`) namespace, mainly focused in back-end and architecture. It also implements a new component called `DynamicContent`, that The DynamicContent which allows managing schema free content blocks via a collection of helpers that significantly ease the integration process.

This component is key and will be extensively used in this new module, for now. Many other details on the implementation can be found in the doc that comes with this PR: https://github.com/PopulateTools/gobierto/blob/3556c523e1e65d50f5fd9e15762d1ed5a262afcb/docs/dynamic-content-component.md

Any new strings are being localised into the three languages available now. I tried to do my best with the Catalan translations, but once again I'd need some help from my beloved @ferblape. Most of the strings are already localised, but there're some in which I had doubts 😛 

---

### How should this be manually tested?

First, run migrations and re-seed the database to make sure the new `GobiertoPeople` module is reachable via the `madrid.gobierto.dev` Site in the left sidebar.

Having http://gobierto.dev/admin/people/people as the main entry point for this feature, let's check that:

- [x] A Person can be created.
- [x] A Person can be updated.
- [x] New (dynamic content) fields can be added in this context.
- [x] A new Content block can be created.
- [x] A Content block can be updated, by changing its title and any of its fields' properties.
- [x] New fields can be added or deleted in a Content block context.
- [x] A Content block can be deleted as well.
- [x] Everything is localised.
- [x] There's a good test coverage for every new part!

---

#### ✏️ Missing stuff

- Integrate the latest front-end from https://github.com/PopulateTools/gobierto/pull/188.
- Extend the `GobiertoPeople::Person` resource by adding new fields present in the latest specification.
- Limit the Content block management options to privileged Admin users (Manager).
- Continue developing the rest of resources at `Person` level such as Events, Statements and Posts.

Issue #190, opened 2016-12-22
People module bootstrap


danguita:
- [x] Set up the new Gobierto module
- [x] Dynamic content block feature design and implementation
- [x] Feature implementation
  - [x] People (w/ Dynamic content extension)
  - [x] Events
  - [x] Statements (w/ Dynamic content extension)
  - [x] Posts

danguita:
I'm moving this big guy forward to Acceptance to hopefully close a first iteration on the People module 🎉

danguita:
I'll be happy to close this one after having https://github.com/PopulateTools/gobierto/pull/224 (https://github.com/PopulateTools/gobierto/issues/219) merged, to not leave anything behind.

danguita:
Forgot to close these 🙃

PR #211, opened 2016-12-30
Gobierto People: Person management front-end

danguita:
Connects to #190.

### Review notes

These changes depend on https://github.com/PopulateTools/gobierto/pull/194. Let's try to get it merged first to keep a significantly better context here.

### What does this PR do?

This PR takes the back-end introduced in #194 and integrates most of the front-end stuff from sandbox files in it. This is including markup, styles and interactions.

First, the Content Block Record UI has been completely redefined according to the spec. All operations such as Add new record, Edit record, Delete record or Cancel changes while editing are now being handled client-side. All those interactions are being tested properly since the most relevant integration test cases have been updated to execute JS and to cover more features. It also reimplements the Content Block management UI itself (Add or delete Block, manage columns in Block...). It now exposes new styles and normalised interactions across sections.

Secondly, It extends the Person form by adding a couple new attributes that were required in the latest spec:

- `avatar_url`, and file upload management.
- `bio_url`, and file upload management.

Also, it adds a policy class to determine which actions concerning Content Blocks are allowed per Admin user.

Last but not least, we have integrated the [Basecamp's Trix](https://github.com/basecamp/trix) WYSIWYG editor into the app, which includes:

- A `FormBuilder` helper that can be used to abstract the markup behind the rich text editor user control. It allows specifying a custom path for handling file attachments:

```ruby
f.rich_text_area :wadus, attachment_path: admin_wadus_file_attachments_path
```

- A custom implementation for handling file uploads through the user control. We're storing these files as attachments in our main asset store (S3 bucket) for now.

Having http://gobierto.dev/admin/people/people as the main entry point for this feature, let's check that:

- [x] The Person management form looks and behaves like the spec/sandbox.
- [x] An Avatar image can be uploaded at Person level, and it's properly shown below the input field when available.
- [x] A generic attachment (PDF, although we're supporting any type of file) can be uploaded at Person level, and its URL to download is properly shows below the input field when available.

screen shot 2017-01-02 at 4 38 58 pm

Regarding Content Block management, here's the check-list:

- [x] A new Record can be added to any of the Content Blocks.
- [x] A Record from any Content Block can be Edited.
- [x] Changes are discarded when hitting the "Cancel" link, in any case.
- [x] When editing Content Blocks, columns can be added or deleted in all available languages (ES, CA at the time being).
- [x] Regular Admins aren't allowed to Create, Edit or Delete Content Blocks.
- [x] Managing Admins (Manager Admin and God Admin) are allowed to do whatever they want to Content Blocks.

These may look like:

screen shot 2017-01-02 at 4 47 52 pm

screen shot 2017-01-02 at 5 11 42 pm

Finally, regarding the WYSIWYG user control:

- [x] Person's bio can be edited in rich text format.
- [x] Any kind of images can be dropped or pasted there and they are automatically uploaded to our main store.

screen shot 2017-01-03 at 8 00 53 am


danguita:
@ferblape Hey, I've just updated this PR's description and check-list so that we can push it forward. I'm removing the **wip** label in the meantime 💪

ferblape:
While testing the PR this exception happened: https://rollbar.com/fernando/gobierto/items/316/

It's fixed in https://github.com/PopulateTools/gobierto/pull/211/commits/3b9f3e015daa8951bb672d8eb08f101583ec2180

danguita:
Good catch, thanks for that 👍

ferblape (approved):
Thank you, everything works as expected.

danguita:
Great! Thanks again :raised_hands:

PR #216, opened 2017-01-04
Indicators Sandbox (WIP)

ferblape:
Connects to #65 

### What does this PR do?

This PR implements the sandbox of the indicators module.


martgnz:
I would say this is OK for merging as a v1.

### Works
- General card implementation with percentage changes, easy to customize
- Three main graphics. One static and the two others with nice voronoi-powered tooltips.

### To be done
- Implement fixed header on scroll (with subsections ?)
- Add different card types
  - Sparkline background card
- Improve the existing graphics
  - Age distribution: add population projections? add comparison between province, state and genders.
  - Unemployment: comparison with province, probably all the cities (w/ light lines). Comparison with state.
  - Rent: button to switch between gross and net rent

### Needs review
- Check if the indicators are correct (percentages, currency where is needed).
- Data review: there is a lot of data transformation going on in the background for some graphics. We should check if our math and code is correct.

cc @ferblape @furilo.

ferblape:
Closed in favour of #248

ferblape:
Let's keep the branch for the moment, just in case the squash of the 65 commits is not correct.

PR #221, opened 2017-01-09
Gobierto People: Person Statements management back-end and UI

danguita:
Connects to #190.

### Review notes

These changes depend on #215.

### What does this PR do?

This PR extends the `GobiertoPeople::Person` data model by adding the Person Statement resource, which groups structured data through the Dynamic content blocks feature.

### How should this be manually tested?

Having `http://gobierto.dev/admin/people/people/<person_id>/statements` (e.g. http://gobierto.dev/admin/people/people/933465536/statements) as the main entry point for this feature, let's check that, after running migrations and re-seeding the database:

- [x] All Person's Statements are listed in the index view.
- [x] A Statement can be Added or Edited in this Person's context.
- [x] The Statement publishing date (`publishes_on` field) is being handled by a date picker control.
- [x] Every Statement has a couple of default Content Blocks that will be customised later per installation, and they allow to do everything regarding Content Block and Record management. See https://github.com/PopulateTools/gobierto/pull/194 and https://github.com/PopulateTools/gobierto/pull/211.

This is how the UI looks like:

screen shot 2017-01-09 at 11 21 55 am

screen shot 2017-01-09 at 11 22 03 am

PR #222, opened 2017-01-09
Gobierto People: Person Blog management back-end and UI, and final touches

danguita:
Connects to #190.

### Review notes

These changes depend on #221.

### What does this PR do?

This PR extends the `GobiertoPeople::Person` data model by adding the Person Post resource, which allows managing Post entries in a Person context. Every Post can have a rich text formatted body field, taking advantage of the `rich_text_area` `FormBuilder` helper (see https://github.com/PopulateTools/gobierto/pull/211). It can be categorised through a list of tags or keywords as well, and they are being stored within the `PersonPost` record. Also, it is providing a simple API to query `PersonPost`s by one or multiple tags.

Finally, this one wraps up some details on the Person resource integration by making People index dynamic, and setting up the deletion strategy.

Having `http://gobierto.dev/admin/people/people/<person_id>/blog` (e.g. http://gobierto.dev/admin/people/people/933465536/blog) as the main entry point for this feature, let's check that, after running migrations and re-seeding the database:

- [x] All Person's Posts are listed in the index view.
- [x] A Post can be Added or Edited in this Person's context.
- [x] The Post body field can be managed in rich text format.
- [x] It is also allowed to upload files through the body field.
- [x] The tags field can be managed as a keyword list, separated by comma.

Regarding the final touches on the People view (http://gobierto.dev/admin/people/people):

- [x] All Events, Statements and Posts counters are being calculated dynamically.

---

This is how the UI looks like:

screen shot 2017-01-09 at 2 43 26 pm

---

screen shot 2017-01-09 at 2 44 09 pm


app/views/gobierto_admin/gobierto_people/people/person_posts/index.html.erb
```diff
@@ -39,7 +39,9 @@
         <%= l(person_post.created_at, format: :short) %>
       </td>
       <td>
-        <%= person_post.tags.to_sentence %>
+        <% if person_post.tags.present? %>
+          <%= person_post.tags.to_sentence %>
```
danguita, line 43: Good catch! What about using `person_post.tags.try(:to_sentence)` instead?

app/controllers/gobierto_admin/gobierto_people/people_controller.rb
```diff
@@ -6,7 +6,7 @@ class PeopleController < BaseController
       before_action { module_enabled!(current_site, "GobiertoPeople") }
 
       def index
-        @people = current_site.people.sorted
+        @people = current_site.people.includes(:events, :statements, :posts).sorted
```
danguita: @ferblape I'm refactoring this thing a little bit because it may be a performance killer. Let me please merge this one once is approved.

app/views/gobierto_admin/gobierto_people/people/person_posts/index.html.erb
ferblape: I'm not a big fan of `#try` and it's quite possible we'll add a text such as "No tags" in a future iteration.

app/controllers/gobierto_admin/gobierto_people/people_controller.rb
ferblape: 👍

app/views/gobierto_admin/gobierto_people/people/person_posts/index.html.erb
danguita: Alright, that smells like a premature optimization but I totally get your point. Let's leave the block 👍

app/controllers/gobierto_admin/gobierto_people/people_controller.rb
danguita: Hey, I ended up configuring a counter cache strategy for those associations to avoid messing with kind of raw queries. It looks a bit premature to me but I feel much better now. What do you think? Feel free to punch the merge button if that looks good to you!

ferblape:
I think it's fine. Let's merge it.

danguita:
Thanks, mate 👍

Issue #219, opened 2017-01-09
Take Dynamic content field's type into account when rendering form fields in the UI


danguita:
### Motivation

Since we are already choosing and storing every Dynamic content field's type from one of "Text", "Date" or "Currency", it is needed to take it into account when asking users to fill those fields in the Admin UI.

### Implementation notes

- [x] When type is "Text", render a text input.
- [x] When type is "Date", render a read-only text input handled by a date picker control.
- [x] When type is "Currency", render a number input in local currency format.

PR #224, opened 2017-01-10
Support rendering based on type and behavior in Dynamic content record fields

danguita:
Connects to #219.
Relates to #190.

### What does this PR do?

It implements a custom Record field initialization mechanism based on each field's type and behavior data attributes:

- A field with `data-type=“date”` will invoke a date picker UI component, taking the current user’s locale into account.
- A field with `data-type=“currency”` will replace the current input type with “numeric” since it behaves almost the same in modern browsers.
- Also, we're refactoring the way we handle the "geocomplete" behavior, for consistency.

Getting into a section in which we have Currency and Date field formats, and check that:

- [x] A Date type field invokes the date picker UI control, but allows to enter any content. I would not make this input read only for the convenience of the user.
- [x] A Currency type field is shown as a number input, whatever it implies in different browsers.

Here is how it should look like:

screen shot 2017-01-10 at 9 31 14 am

screen shot 2017-01-10 at 9 31 36 am

### DIFF
--- a/app/assets/javascripts/admin/app/components/dynamic_content_component.js
+++ b/app/assets/javascripts/admin/app/components/dynamic_content_component.js
@@ -2,6 +2,7 @@ this.GobiertoAdmin.DynamicContentComponent = (function() {
   function DynamicContentComponent() {}
 
   DynamicContentComponent.prototype.handle = function(wrapper, namespace) {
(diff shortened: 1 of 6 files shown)


ferblape (approved):
All good here! 👍

Issue #226, opened 2017-01-11
Site municipality autocomplete is not being triggered to the proper URL


ferblape:
Right now, when creating or updating a site, the autocomplete tag for municipalities is not triggering the request to the url defined in `data-autocomplete-url` but to the current url.

See the log trace:

```
Started GET "/admin/sites/new?query=Ses" for 127.0.0.1 at 2017-01-11 15:53:40 +0100
Processing by GobiertoAdmin::SitesController#new as TEXT
  Parameters: {"query"=>"Ses"}
...
```

danguita:
Let me check it 👀

PR #228, opened 2017-01-11
Bind jquery-ui’s autocomplete explicitly to avoid collisions between plugins

danguita:
Connects to #226.

### What does this PR do?

This PR initializes the jQuery UI's `autocomplete` plugin explicitly to avoid collisions between plugins using the same prototype.

It turns out that, after the JS reorganization in https://github.com/PopulateTools/gobierto/pull/196, we are loading two different plugins (https://api.jqueryui.com/autocomplete/ and https://github.com/devbridge/jQuery-Autocomplete) which expose the `autocomplete()` prototype, so the last loaded one was overriding the first 🙃.

In our specific case, jQuery UI's `autocomplete` couldn't be reached thus it was not throwing any exceptions but obviously it was not working as expected.

We've also checked that both plugins are being used, one in the Admin namespace, and other by the `gobierto-budgets` module, so I would keep both of them for now taking into account that the load order is now relevant, and we should try to normalize both implementations to keep only one of them. That's a bit out of scope, but I would go for jQuery UI's one since it is being more actively maintained. If agreed, let's just create a new Issue to deal with that.

Also, we've found a side problem in the way because of the custom `xhr` content type response on certain Budget Content Item management actions. It has been refactored as well to always return well know HTML format.

Just get into the Site edit or Budget Consultation items edit views (both were using the jQuery UI autocomplete widget) and check that they now work as expected.

- [x] Sites management: `http://gobierto.dev/admin/sites/<site_id>/edit` (http://gobierto.dev/admin/sites/782762298/edit).
- [x] Budget Consultation Items management: `http://gobierto.dev/admin/budgets/consultations/<consultation_id>/items` (http://gobierto.dev/admin/budgets/consultations/321127170/items).


ferblape:
Thank you!

I agree with you about the versions, I'll create the issue.

PR #248, opened 2017-01-13
Gobierto Indicators module

ferblape:
This PR implements the sandbox of #65 

@martgnz says this is OK for merging as a v1.

### Works
- General card implementation with percentage changes, easy to customize
- Three main graphics. One static and the two others with nice voronoi-powered tooltips.

### To be done
- Implement fixed header on scroll (with subsections ?)
- Add different card types
  - Sparkline background card
- Improve the existing graphics
  - Age distribution: add population projections? add comparison between province, state and genders.
  - Unemployment: comparison with province, probably all the cities (w/ light lines). Comparison with state.
  - Rent: button to switch between gross and net rent

### Needs review
- Check if the indicators are correct (percentages, currency where is needed).
- Data review: there is a lot of data transformation going on in the background for some graphics. We should check if our math and code is correct.

cc @ferblape @furilo.


ferblape:
I have just released this branch to staging. We can see real indicators for Madrid in http://madrid.gobify.net/indicadores

Next steps: @furilo and @ferblape to review and create issues if necessary.

Issue #232, opened 2017-01-12
Person > Events: Add an "All" filter?


furilo:
Now when you are seeing a person events, it's not clear what you are seeing, because no option of the filter is show as selected. Should we add a first option "All". Also, add a column in the table with the status.

danguita:
Yeah, agree. Let me take care 👌

PR #252, opened 2017-01-18
Add “All events” filtering option to Person Events index view

danguita:
Connects to #232.

### What does this PR do?

As requested in #232, it adds a new filtering option to the Person Events index view to provide better feedback to the Users.

Taking a URL like http://gobierto.dev/admin/people/people/933465536/events, check that:

- [x] The new "All" filtering option is active by default.
- [x] The counters are correct (here we are counting both published and pending events).

This is how it looks like:

screen shot 2017-01-18 at 8 15 32 am


ferblape (approved):
Works as expected.

Issue #234, opened 2017-01-12
Simplify the label for buttons


furilo:
Right now we have: Crear post, Actualizar declaración, ... In some situations the column is narrow, so the text gets cut, or paddings are not respected: 

![captura de pantalla 2017-01-12 a las 16 21 41](https://cloud.githubusercontent.com/assets/24740/21895483/362718f4-d8e3-11e6-8cb5-544c98151831.png)

So I would just leave the verb: Crear, Guardar...

danguita:
No prob. Would you do so in every single case?

furilo:
In the back at least, yes.

danguita:
Alright, it makes perfect sense to me 👌

PR #253, opened 2017-01-18
Drop the model name from all localised form submit tags

danguita:
Connects to #234.

### What does this PR do?

Here we are changing the texts of all localised submit tags throughout the entire UI, affecting both Admin and User namespaces. We could have gone for a model-specific localisation process instead but, after checking some uses of the submit tag helper, it seemed that it can be applied at global level.

Integration tests are passing but let's check that there are no side effects in other namespaces.

- [x] Just check that every submit button within the Admin UI displays only the corresponding verb.

screen shot 2017-01-18 at 8 45 28 am

### DIFF
--- a/config/locales/defaults/ca.yml
+++ b/config/locales/defaults/ca.yml
@@ -135,9 +135,9 @@ ca:
     select:
       prompt: Si us plau tria
     submit:
-      create: Crear %{model}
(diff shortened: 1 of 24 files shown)


ferblape (approved):
Looks like everything works

Issue #236, opened 2017-01-12
Text for the published state


furilo:
![captura de pantalla 2017-01-12 a las 17 06 16](https://cloud.githubusercontent.com/assets/24740/21897287/7200f9f2-d8e9-11e6-96d1-ae78d7efbd69.png)

Right now we are using "Activa". Maybe it's best "Publicado/a"?

danguita:
Sure, it's just localisation stuff.

PR #254, opened 2017-01-18
Review models’ `visibility_level` attribute labels

danguita:
Connects to #236.

### What does this PR do?

As described in #236, this one just replaces all "Active" labels with "Published" for any `visibility_level` attributes around there.

- [ ] Every user control to select a Resource's visibility level, generally as radio buttons, displays "Published" instead of "Active" for the `active` visibility level option, this way:

screen shot 2017-01-18 at 9 29 41 am

screen shot 2017-01-18 at 9 29 50 am


ferblape (approved):
Easy one!

PR #255, opened 2017-01-18
Review file attachments feature in WYSIWYG editor

danguita:
Connects to #242.

### What does this PR do?

As requested in #242, in order to assist users on attaching files right in the WYSIWYG editor, we are displaying a hints block below every `rich_text_area` field. Also, we're implementing an editor action to allow attaching files right from the toolbar.

So far, we are using WYSIWYG editors in the following views:

- Person edit: http://gobierto.dev/admin/people/people/933465536/edit
- Person's Post edit: http://gobierto.dev/admin/people/people/933465536/blog/585972795/edit

So, taking those as entry points, let's check that:

- [x] Hints are displayed below the editor control.
- [x] A file can be attached and uploaded by using the toolbar button.
- [x] The drag and drop interface is still available.

This is how it looks like: 

screen shot 2017-01-18 at 11 05 49 am


ferblape (approved):
👍

Issue #241, opened 2017-01-12
Content blocks: format content according to its field type?


furilo:
![captura de pantalla 2017-01-12 a las 17 25 16](https://cloud.githubusercontent.com/assets/24740/21898129/196dc6c8-d8ec-11e6-8a8f-0a84d0b616c6.png)

If the field is Currency, let's format the output. 

Also, if you enter a formatted number (ie. 123.456,98) it looses the quantity. 

We can solve the previous problem if we implement some format-as-you-type system (ie. https://nosir.github.io/cleave.js/ - but let's think if we can solve it with existing included libraries to avoid including more)

ferblape:
I'm going to close this issue. At the end, storing the numbers with a format is not a big deal because the spreadsheets work fine with the decimals.

Issue #239, opened 2017-01-12
Content blocks: show the form again after entering a row


furilo:
When you are entering a row, it's likely that you are entering several. So maybe it makes sense that after adding a row, you are presented with the form opened again (and as in #237 with the focus on the first input)

Issue #237, opened 2017-01-12
Content blocks: focus on first input when you click on "Add new row"


furilo:

danguita:
No big deal 👌

PR #256, opened 2017-01-18
Dynamic content UI component review and refactor

danguita:
Connects to #237.
Connects to #239.
Connects to #241.

### What does this PR do?

This PR brings together several improvements regarding the Dynamic content UI component. We also take advantage of the review to make some adjustments in terms of performance and maintainability.

- [x] When adding a new row, set focus on the first input (#237).
- [x] Related to the point above, set focus on the first input when editing a row (extra 🎱)
- [x] Right after committing a record, display a new empty row (#239).
- [x] Currency fields are being formatted as you type via the Cleave.js library, and they have the right grouping and decimal marks for EUR currency (#241).

### DIFF
--- a/app/assets/javascripts/admin/app/components/dynamic_content_component.js
+++ b/app/assets/javascripts/admin/app/components/dynamic_content_component.js
@@ -142,10 +142,14 @@ this.GobiertoAdmin.DynamicContentComponent = (function() {
   }
 
   function _cleanupRecordField(selector) {
-    if (selector.attr("type") === "text") {
-      selector.val("");
-    } else if (selector.attr("type") === "select") {
+    if (selector.attr("type") === "hidden") {
+      return true;
+    }
+
+    if (selector.attr("type") === "select") {
       selector.find("option:selected").prop("selected", false);
+    } else {
+      selector.val("");
     }
   }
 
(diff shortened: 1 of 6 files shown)


ferblape (commented):
Instead of having static values for numbers format, I have changed this to read the settings from I18n.

ferblape:
Taking care of the tests...

danguita:
👌 Let me know if I can help with anything.

ferblape:
Cleave is not playing well with commas, so I have changed slightly the tests. 

Anyway, I'm not sure now if this is the best solution, and maybe it's my fault for changing my mind here, but I think that the number fields should be stored as strings but without any format (just the decimal string) and we should format the output dynamically.

This is because I'm thinking when we export the data as CSV and the numbers must look as numbers, not as strings with the thousands separator.

So I'll accept this PR and I'll keep #241 open.

ferblape:
I think Circle is not working with my push --force, because those tests are green in my laptop. I'm going to merge.

danguita:
Yeah, it's hard to say what would fit better here in the future. Anyways, since we are not storing numeric fields at the moment, it is not trivial to perform calculations. I was thinking of keeping a "front" field in which we can do any kind of cool formatting, and an well-typed one for internal use only. I agree on pushing this one forward and taking care later 👍

danguita:
Oh, I force pushes very often and CI builds are almost always triggered 😕 Let's see how it goes in `stable`.

danguita:
Forgot to thank you for checking!

ferblape:
Hmmm, I'll fix it, definitely is broken :D 

Thing is that I need to run `rake assets:precompile` to compile the assets for the tests each time.

danguita:
Uhm, not happening here... The test runner is dealing with all the stuff for me:

```
$ bundle exec rails test test/integration/gobierto_admin/gobierto_people/person_statement_create_test.rb
```

danguita:
Yay! 😜

Issue #240, opened 2017-01-12
Content blocks: warning if you have added content and click on "Add new block"


furilo:
I had added several rows for a block, and thought that it was being saved; then went to add a new block, and came back to find out that nothing was saved ;) 

So: show a warning it you are leaving the page with the content unsaved. Don't know if we should think about something general, applicable to any form/page. What do you think @ferblape ?

danguita:
As reported in https://github.com/PopulateTools/gobierto/issues/178, I was thinking of just going for a global solution at Admin namespace level. Anyways it's always good to be aware of very specific cases like these so that we can take special care of them. How does it sound to you?

furilo:
Ok great, let's go for a global solution. Please reorg issues (close/new/...) as it fits you best.

ferblape:
I was about to say the same as @danguita, so 👍

danguita:
Alright, thank you guys 👌

danguita:
Hey guys, I have ended up implementing that generic component so we can take advantage of it across the entire Admin area. I think that it would be useful in most cases so it is enabled everywhere by default. What do you think? Let's check it out at https://github.com/PopulateTools/gobierto/pull/258.

ferblape:
It's great to have a generic solution for everything, thank you! Closing in favour of #258

furilo:
+1 to enable it by default everywhere!

PR #258, opened 2017-01-19
Implement Dirty forms component with Turbolinks and custom event support

danguita:
Connects to #240.

### What does this PR do?

This PR implements a JS component which will be responsible for warning for unsaved changes when leaving a page. It is supporting Turbolinks and implements custom events that can be triggered by other UI components.

This behavior can be skipped by adding the `.skip-dirty-check` class at `form` tag level.

We would need to check that changes are detected and the alert dialogs are shown as expected:

- [x] Detects changes any kind of form input controls (`<input>`, `<select>`, `<textarea>`).
- [x] Detects text changes in a WYSIWYG control (Trix).
- [x] Detects file attachments in a WYSIWYG control (Trix).
- [x] Detects changes in a date picker control.
- [x] It can intercept a Turbolinks event, or leave it to the `beforeunload` window event.
- [x] The confirm/alert dialogs are displayed when forms are dirty 🙌 

Confirm dialog triggered while listening to Turbolinks events:

screen shot 2017-01-19 at 9 14 52 am

Native `beforeunload` event dialogs (leave and reload):

screen shot 2017-01-19 at 9 15 18 am

screen shot 2017-01-19 at 9 14 58 am

### DIFF
--- a/app/assets/javascripts/admin/app/components/dirty_forms_component.js
+++ b/app/assets/javascripts/admin/app/components/dirty_forms_component.js
@@ -1,0 +1,40 @@
+this.GobiertoAdmin.DirtyFormsComponent = (function() {
+  var isDirty;
+
+  function DirtyFormsComponent() {}
+
+  DirtyFormsComponent.prototype.handle = function(message) {
+    $(document).on("turbolinks:load", _handleDirtyFlag);
+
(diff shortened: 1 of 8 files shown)


ferblape (approved):
Works really well!

Issue #250, opened 2017-01-17
Admin: Can't select site


furilo:
In http://gobify.net/admin when you click on a site, you only get the page you are on reloaded

![captura de pantalla 2017-01-17 a las 15 36 15](https://cloud.githubusercontent.com/assets/24740/22024568/b23d1ae8-dcca-11e6-97e1-955147e7dc2b.png)

danguita:
Since we are using sessions to keep the site context, could it be related to the session store configuration for that domain? Are the rest of session-based resources, such as for Admins and Users sign in, working just fine?

This is not happening on my development env. Let me know if I can help further.

/cc @ferblape

furilo:
OK, seeing it with @ferblape I saw the site is being changed, but since you are left in the same page, it seems that nothing has changed. I mean: if you are editing one site's properties and you use the dropdown to select another site, you are left in the same original site you were in. So it seems nothing has changed. 

I would expect that you get redirected to the same page but for the site you have changed to. What do you think?

danguita:
Yeah, now I see. I guess you were on a `sites/edit` page (the one titled with "Personalizar site" in it) when switched Sites. As you mentioned, we are redirecting the user to the same URL she came from, so in this very specific case it is correct but not much user friendly. I would say covering this specific scenario won't hurt, how about letting me take care of it now?

danguita:
Hope you don't mind I picked up this one and took care in https://github.com/PopulateTools/gobierto/pull/259. I think it's a good improvement anyways, thanks for reporting 👍

PR #259, opened 2017-01-19
Implement a specific redirect behavior when switching Sites in Admin

danguita:
Connects to #250.

### What does this PR do?

This one implements a specific redirection behavior when switching Sites from a specific path.

This path is `edit_admin_site_path(<site>)` (e.g. http://gobierto.dev/admin/sites/1009469985/edit). In brief, when the request comes from `edit_admin_site_path`, we should be redirecting to the same path for the recently created Site session. Otherwise, we keep redirecting to `HTTP_REFERER` URL as we do already.

Taking the Site seeds available right now, let's check that:

- [x] Switching from "Santander" to "Madrid" when in an Santander's Site edit path (e.g. http://gobierto.dev/admin/sites/782762298/edit), the user gets redirected to Madrid's edit path (e.g. http://gobierto.dev/admin/sites/1009469985/edit) instead of to the actual referrer URL (e.g. http://gobierto.dev/admin/sites/782762298/edit).
- [x] Vice versa 😛

### DIFF
--- a/app/controllers/gobierto_admin/sites/sessions_controller.rb
+++ b/app/controllers/gobierto_admin/sites/sessions_controller.rb
@@ -3,7 +3,7 @@ module GobiertoAdmin
     def create
       if allowed_site?(params[:site_id])
         enter_site(params[:site_id])
-        redirect_to(request.referrer)
+        redirect_to(redirect_path)
       else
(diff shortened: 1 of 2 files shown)


ferblape (approved):
👍

Issue #213, opened 2017-01-02
Implement an alternative uploader for file attachments


danguita:
### Motivation

Dealing with file attachments may be a little annoying in `development` environment since every uploading process is actually reaching the AWS S3 bucket specified for uploads. It would be great to implement an alternative uploader to be used on certain environments and avoid hitting external services.

Please note that we are not hitting any of these external services in `test` environment.

### Implementation notes

- [x] Implement a local file uploader that writes the uploaded file to the local file system, à la Paperclip/Carrierwave.
- [x] Determine when to use one or another uploader, taking the current environment into account.

### Workflow notes

- Keeping it in back-log since it is not a priority for now.
- May depend on https://github.com/PopulateTools/gobierto/issues/212.

/cc @ferblape

ferblape:
👍

Issue #212, opened 2017-01-02
Normalize and sanitize attachment file names


danguita:
### Motivation

Right now we are setting file names manually every time an attachment file is uploaded through the `FileUploader` library. That may cause inconsistencies and both functional and maintainability issues, so let's avoid them by implementing a Service class that can act as interface to the actual file uploading library.

### Implementation notes

- [x] Refactor the current `FileUploader` lib by splitting it into a library to just deal with the file transport thing, and a Service class that may know a little about business logic.
- [x] Make the `FileUploader` service take kind of Site, Module, Resource and Attribute parameters to be able to build a normalized file name.
- [x] Also take care of sanitizing that file name.

/cc @ferblape

ferblape:
👍

PR #260, opened 2017-01-19
Implement a Service class to handle uploads across multiple adapters

danguita:
Connects to #212.
Connects to #213.

### What does this PR do?

In this PR we are refactoring the current `FileUploader` kinda service-library by splitting it into a simpler library and a service class. The new service class is called `FileUploadService` and it should be always implemented to deal with file uploads through attachment behaviors (#212).

We are taking advantage of this refactor to implement an alternative file upload adapter (#213) for ease and portability in development environment. In these cases, the adapter is automatically switched and all files are stored locally instead of hitting external services such as AWS S3.

- [x] The `FileUploader::S3` adapter is still working fine.
- [x] The `FileUploader` adapter is automatically switched to `Local` when in development environment.
- [x] By using the `FileUploader::Local` adapter, all files are stored locally under `/public/attachments`, and everything regarding the UI still works as before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ config/deploy/production.rb
 
 # Coverage reports
 /coverage
+
+# Uploaded files
+/public/attachments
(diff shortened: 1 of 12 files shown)


ferblape (approved):
🚀

Issue #266, opened 2017-01-20
Link to public resources


ferblape:
Follow up of #223

All Gobierto People resources when listed in the admin interface have a link to view the resource in the "public" part. Implement this link.

danguita:
I can handle this one 👌

PR #225, opened 2017-01-11
Integrate Gobierto People module in User namespace

danguita:
Connects to #223.

### Review notes

In this PR we're for now bootstrapping the Gobierto People module within User namespace. It is here just to open the discussion on how these new features should be implemented. Please, do not merge it for the time being.

### What does this PR do?

Here's what we have done so far:

#### Module's homepage

Entry point: http://madrid.gobierto.dev/cargos-y-agendas/

- People summary
  - People filtering (categories, parties and political groups) links.
  - View all link.
- Events summary
  - Past events link.
  - View all link.
- Posts summary
  - View all link.
- Other sections links:
  - "Salarios y retribuciones" links to Statements view.
  - "Obsequios y regalos" links to an environment-level defined URL (`$PEOPLE_GIFTS_SERVICE_URL`).
  - "Viajes y desplazamientos" links to an environment-level defined URL (`$PEOPLE_TRAVELS_SERVICE_URL`).

#### People section

Entry point: http://madrid.gobierto.dev/cargos-y-agendas/cargos

- People grid view.
- People filtering options:
  - By Government party: http://madrid.gobierto.dev/cargos-y-agendas/cargos (that's the default filter).
  - By Opposition party: http://madrid.gobierto.dev/cargos-y-agendas/cargos-en-oposicion.
  - By Executives category: http://madrid.gobierto.dev/cargos-y-agendas/cargos-directivos.
  - View all, reset filters: http://madrid.gobierto.dev/cargos-y-agendas/todos-los-cargos.
  - By political groups (those are managed in the Admin UI). e.g.: http://madrid.gobierto.dev/cargos-y-agendas/grupos-politicos/742272447/cargos.
- Subscription module.

#### Events section

Entry point: http://madrid.gobierto.dev/cargos-y-agendas/agendas/eventos

- Events list view.
- Events filtering options:
  - By Government party: http://madrid.gobierto.dev/cargos-y-agendas/agendas/eventos (that's the default filter).
  - By Opposition party: http://madrid.gobierto.dev/cargos-y-agendas/cargos-en-oposicion/agendas/eventos.
  - By Executives category: http://madrid.gobierto.dev/cargos-y-agendas/cargos-directivos/agendas/eventos.
  - View all, reset filters: http://madrid.gobierto.dev/cargos-y-agendas/todos-los-cargos/agendas/eventos.
[...]

PR #269, opened 2017-01-20
Add public references to every Person resource in the Admin UI

danguita:
Connects to #266.

### What does this PR do?

This PR replaces some placeholders we left intentionally in the Admin UI until having the [User's part of this Module](https://github.com/PopulateTools/gobierto/pull/225) merged. Those placeholders were links to refer the User namespace UI (public) from the Admin UI.

#### Extra 🎱 

Here we are also changing the way we install the recently added `rails/actionpack-action_caching` gem. As we can read in [GitHub help docs](https://help.github.com/articles/which-remote-url-should-i-use/#cloning-with-https-urls-recommended), it is recommended to clone repos with HTTPs URL. Also, there may be issues when installing gems within the Docker-based environment. So, since `bundler` can handle both formats, I would go for the most compatible one.

In every index page, we should be rendering a link to the specific resource in the User's namespace when it is publicly available. Otherwise, that link should look kind of disabled.

The pages in which we would need to check this behavior are:

- [x] People index: http://gobierto.dev/admin/people/people
- [x] Person Events index: http://gobierto.dev/admin/people/people/933465536/events
- [x] Person Statements index: http://gobierto.dev/admin/people/people/933465536/statements
- [x] Person Posts index: http://gobierto.dev/admin/people/people/933465536/posts

This is how those links should look like:

screen shot 2017-01-20 at 5 27 07 pm

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem "ruby_px"
 gem "responders"
 gem "dalli"
 gem "cookies_eu"
-gem "actionpack-action_caching", git: "git@github.com:rails/actionpack-action_caching.git", ref: "9044141824650138bf27741e8f0ed95ccd9ef26d"
+gem "actionpack-action_caching", git: "https://github.com/rails/actionpack-action_caching.git", ref: "9044141824650138bf27741e8f0ed95ccd9ef26d"
 
 # Frontend
 gem "sass-rails", "~> 5.0.0"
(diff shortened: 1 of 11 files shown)


Gemfile
```diff
@@ -11,7 +11,7 @@ gem "ruby_px"
 gem "responders"
 gem "dalli"
 gem "cookies_eu"
-gem "actionpack-action_caching", git: "git@github.com:rails/actionpack-action_caching.git", ref: "9044141824650138bf27741e8f0ed95ccd9ef26d"
+gem "actionpack-action_caching", git: "https://github.com/rails/actionpack-action_caching.git", ref: "9044141824650138bf27741e8f0ed95ccd9ef26d"
```
ferblape, line 14: 👍

ferblape (approved):
Thank you!

Issue #305, opened 2017-02-06
Update admin > password is required even when the password hasn't changed


ferblape:
If you try to update an admin information her password is required by the application.

PR #309, opened 2017-02-06
Don't require password confirmation if the password hasn't changed

ferblape:
Connects to #305

### What does this PR do?

This PR fixes admin update error that required to introduce a new password each time.

### DIFF
--- a/app/forms/gobierto_admin/admin_form.rb
+++ b/app/forms/gobierto_admin/admin_form.rb
@@ -2,8 +2,6 @@ module GobiertoAdmin
   class AdminForm
     include ActiveModel::Model
 
-    EMAIL_ADDRESS_REGEXP = /\A(.+)@(.+\..+)\z/
-
     attr_accessor(
       :id,
       :name,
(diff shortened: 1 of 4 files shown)


app/forms/gobierto_admin/admin_form.rb
```diff
@@ -23,7 +23,8 @@ class AdminForm
 
     validates :name, :email, presence: true
     validates :email, format: { with: Admin::EMAIL_ADDRESS_REGEXP }
-    validates :password, presence: true, confirmation: true
+    validates :password, presence: true, if: Proc.new { |f| f.admin.new_record? }
+    validates :password, confirmation: true, if: Proc.new { |f|  f.password != f.admin.password }
```
danguita: Hey, sorry for intruding here, but wanted to point out that the conditional password confirmation check may not be needed here. I mean that the default `confirmation` validation behavior is taking the field presence into account. Also, I'd say we can take advantage of the `new_record?` property which is already there to condition the password presence check.

Let me please add a couple of commits to discuss that further.

danguita: @ferblape because code is worth than thousand words, I've just pushed my thoughts here 😛 Would you say this is still readable and should work as expected?

ferblape:
Hey @danguita great to see you around! Yes, your approach is nicer and makes sense to use `new_record?` here. Thanks!

Issue #342, opened 2017-02-14
People/Person: send email through form


furilo:
- [x] Markup
- [x] Integration

* Captcha: I guess we should use a Captcha since anybody should be able to send an email. Google seems to be working in an invisible captcha: http://www.ghacks.net/2016/12/05/googles-invisible-captchas/ (but is not ready yet: https://www.google.com/recaptcha/intro/comingsoon/invisiblebeta.html). Also: https://github.com/markets/invisible_captcha
* Destination email: for a further iteration, we may offer a conf option so this emails are received in a different email than the user's one.

PR #346, opened 2017-02-14
342 email person

furilo:
Connects to #342 

### What does this PR do?

Provides the markup for the email form.


ferblape:
This is ready and can be tested in staging.

Issue #330, opened 2017-02-13
Share buttons integration + OG metatags


furilo:
- [x] Metatags: description
- [x] OG tags for TW + FB; title, img (default: logo), description
- [x] Share buttons: with the page title + url

PR #357, opened 2017-02-15
Share buttons and meta tags

ferblape:
Connects to #330 

### What does this PR do?

This PR adds the meta tags for OG and Twitter + adds a meta tag desription too.

Review all the pages headers, and test the share buttons.


ferblape:
@furilo I'm adding a description based on the title and the current page, but nothing very sophisiticated. In the case of a person, I'm adding the bio, but I'm not sure it's a good idea.

furilo:
Let's go with this for the moment.

PR #372, opened 2017-02-20
Initial CSS theme, and some other CSS adjustments

furilo:
Connects to #327 

### What does this PR do?

- Prints the Header and Footer HTML blocks on the Admin UI into the application layout. 
- Adds a basic initial theme (Dival)


ferblape (approved):
👍

PR #403, opened 2017-02-25
Development setup docs (without Docker)

amiedes:
Hi! I decided to give it a try to set up gobierto without using Docker, since the docker startup process for running the tests was taking almost 3 minutes for me.

### What does this PR do?

This PR adds the development setup instructions to run gobierto on localhost without Docker.

Not directly (read below)

### Comments

I managed to do it but there are a couple of things I had to change within the code **(and which are not present in this docs)** related to the configuration of Redis and ElasticSearch. I feel the reason I had to do this is that I was missing something about how environment variables work when combined with rails, so maybe you can give me some insight about this.

I did fill in the `REDIS_URL`and `ELASTICSEARCH_URL` values in the `.env` file, but I was getting weird errors related to wrong URIs.

1. For making Redis work I had to change the `redis_client.rb` from this:

```ruby
$redis = Redis.new({
  url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/0" },
  namespace: APP_CONFIG["site"]["name"]
})
```

To this:

```ruby
$redis = Redis.new({
  host: 'localhost',
  port: 6379,
  namespace: APP_CONFIG["site"]["name"]
})
```

2. I also had to change the `elastic_url` in `config/secrets.yml` from:

```yaml
elastic_url: <%= ENV.fetch("ELASTICSEARCH_URL") { "http://localhost:9200" } %>
```

to:

```yaml
elastic_url: http://localhost:9200
```


danguita:
First of all, thanks so much for sharing your feedback! 🙌

Secondly, hope you don't mind me sharing my thoughts here. I'm not against having multiple ways to build and set up the Gobierto's development environment, but I'd try to make the current options better instead of diversifying them at a first try. Of course, the native way (w/o Docker) is always an option, and it might not be necessary to write instructions down since all the configuration is based on environmental variables, so the `.env` file can be totally customised for your needs, and setting up the environment natively should be kinda trivial.

Regarding the Docker-based set up, which is working good for some of us, I'd extend the current documentation to get around I/O performance issues that would speed up **dramatically** the test build and the development process itself. I'd say that there a couple of quite straightforward tools to manage that:

- If you were using [Docker machine](https://docs.docker.com/machine/), [this script](https://github.com/adlogix/docker-machine-nfs) will mount shared folders via NFS. Please note it's just available at Homebrew's repo: `$ brew install docker-machine-nfs`.
- If you [went native](https://www.docker.com/products/docker), please check [these instructions](https://github.com/IFSight/d4m-nfs) to replace osxfs with NFS.

What do you think? Would you say we can have a try to improve the Docker-based environment, which IMO is the most portable option for now?

amiedes:
Thanks a lot for your opinion, @danguita !

I agree with you about improving the Docker-based environment.

For the moment I think I'll stick with the native setup while I catch up on how Docker works. I'll also try to improve the related documentation meanwhile.

Regarding the speeding tools your recommended, I was using both `docker-machine` and `docker`. Does this mean I was using native or not?

danguita:
Hey @amiedes, if your Docker host is managed, in example, by VirtualBox, then you are using Docker machine. Nevertheless, the `docker` client should be available at host level so it allows you to access the containers that are running on the VM. If the environment just works well for you, I'd install and run the `docker-machine-nfs` command to check if the I/O performance can be improved. Thanks!

amiedes:
I've been trying several things to solve this problem:

* Using the `docker-machine-nfs` command to use NFS for the Docker containers.
* Increase the number of memory and CPUs Docker can use.
* Increase the number of memory and CPUs the underlying MV can use, re-creating it with the `docker-machine create --driver virtualbox --virtualbox-cpu-count 4 --virtualbox-memory 5120 default`.

I've been measuring times for different combinations of these options but I haven't seen any major improvement yet, or discovered a clue about what's happening. The total time for running the tests with Docker it's still around 7 ~ 9 min. while on native it's around 2:30 😕

BTW, is this happening to anyone else or just me?

Also, I think we can close this PR and maybe move this discussion into an issue.

danguita:
@amiedes those steps look like a great starting point to address this issue, thanks for pointing them out so clearly.

Regarding the test build times, there is definitely a huge gap between configurations, so we could be missing something 😕. In my case, running the Dockerized environment in a Docker machine, a complete build is generally taking less than 150s:

```
$ docker-compose run test bundle exec rails test

...
Finished in 146.35399s
642 tests, 1725 assertions, 0 failures, 0 errors, 0 skips
```

The CI service [takes around 140s](https://circleci.com/gh/PopulateTools/gobierto/700), so these numbers seem to make sense.

In case it helps, I'm assigning all physical CPU cores and 25% of the total available memory to the VirtualBox machine:

```
$ docker-machine create \
  --driver virtualbox \
  --virtualbox-cpu-count 2 \
  --virtualbox-memory 4096 \
  default
```

Having said that, this huge build time difference could be telling us something. I would take a new look, perhaps recreating the machine to get numbers from a clean state. Sorry for being that guy, but I would exhaust all options before diversifying the environment 😛

furilo:
@blat maybe more info about the .env file would be useful for a non-docker setup?

amiedes:
Hello again!

I'd like to share a few observations I made so maybe your experience can help me throw up some light on this 😅.

* I've went through the whole docker setup several times and the `docker-machine-nfs` doesn't seem to make any kind of improvement. 😕
* I said the total time for running the tests in Docker was around 7 ~ 9 min. Well, to be more precise I think that what is exactly happening is the following:
  * The first time the tests are run, after a fresh vm is set up, they take around 4 min.
  * But on following 'runs', they take each time more time to complete. Increasing more or less at the following ratio: 4' - 6' - 8' - 10' etc.
* Sometimes, the first time the tests are run after a fresh vm is set up, I'll get several errors of the following kind:

  captura de pantalla 2017-02-28 a las 20 28 49

  This is weird to me. I would expect duplicate key errors to appear on 'test runs' after the first one, so if the first one left some records undeleted, keys would collide. But what it's happening it's exactly the opposite. 

* I found the following [post](http://bradgessler.com/articles/docker-bundler/), which explains that one of the ways of speeding up docker is keeping the needed gems in a `/bundle` shared volume, so they don't need to be installed again each time tests are run. Since I saw you're using a `bundle` container with a `bundle` volume I assume you're using this technique. In my case, when I run `docker-compose run web bundle install` I get the following output, which means the gems are always being re-installed again:

  captura de pantalla 2017-03-02 a las 0 23 32

  This makes me wonder if the shared `/bundle` volume is indeed working. 🤔

* After a few 'test runs' I *sometimes* get postgres deadlocks. ¿Is this usual? I had never seen it before.

  captura de pantalla 2017-02-28 a las 20 16 16

  I know '*sometimes*' is a pretty vague word but I haven't been able to narrow the problem to see when is this exactly happening.

* After a few 'test runs' I started getting also this kind of errors:

  captura de pantalla 2017-03-02 a las 0 02 00

I know I posted a lot of things but I'm feeling pretty lost right now 😵. Any help is really appreciated and welcomed!

danguita:
Hey @amiedes, thanks for the thorough report. I'm afraid I've been using the Dockerized environment from the very beginning but can't reproduce any of these problems.

Regarding the incremental build time, it does not make any sense to me since it is an isolated and idempotent operation, by definition. Also, every build is executed in an exclusive container that is only sharing vendors such as bundled gems through the `bundler` container and the database service itself, so there is no way of being affected by any other parts on the system. I also could not get any database deadlocks nor constraint violations 😕 

It seems that something related to the testing database cleanup process, or even concurrent executions of testing builds against the same database service could be happening. Have you set up any kind of continuous testing process? Let's ensure you only run a build at a time.

We could be missing something, and I know that's not a proper answer but I would suggest you to clean things up by recreating the Docker machine and trying things out from scratch. I'm sorry I cannot be of more help for now.

ferblape:
Thank you both for your comments. This is what I'd do:

- first, environment variables are quite a bit complex and un-documented so we have writen a small document: https://github.com/PopulateTools/gobierto/pull/480. Please, update your branch and maybe you want to link to this file in the instructions

- I like to have two separated environment files, because right now it seems that Docker is a requirement, and it shouldn't be. We need to make it as easier as possible for the user, even though we have a lot information duplicated in two documents

- any comment and improvement in the Docker setup would be great, of course, because I don't use it and it might get broken or outdated without noticing it

danguita:
@ferblape Hey, thanks for sorting things out 🙌 

I don't see the Docker setup as a requirement because all configuration files are compatible with any kind of setup (or they should be!), but I totally get your point since we do not provide any other installation instructions. As discussed in this thread, I'm not against having multiple options to build Gobierto, but I would try addressing any problems we could have to enhance the old docs and to reinforce users before moving forward to an alternative approach.

Having said that, since it is not yet clear what could be wrong with @amiedes' Docker setup, I would just move on this because it was a very sensible approach to me from the beginning 👍 

@amiedes Hope I did not bother you much with all the performance stuff 😛

PR #406, opened 2017-02-27
Add privacy page link to login / signup / subscription elements

ferblape:
Connects to #336

### What does this PR do?

This PR adds a helper which is rendered in the login, signup and subscription elements with a link to the privacy page if it's configured in the site.

It also adds a new control in the site form to select the privacy page.


ferblape:
@furilo please review if it's ok with this for the moment or should we add also the link to the Terms page and / or a checkbox.

Issue #435, opened 2017-03-02
Send consultation improvements


martgnz:
When the user fills all the cards we should paint a banner to encourage sending the form in the desktop version.

In mobile we can improve the current warning, possibly switching the background to a neutral color (now is green, which could make the user think their consultation is already sent).

PR #444, opened 2017-03-02
Send consultation improvements

martgnz:
Connects to #435.

### What does this PR do?
This PR adds a banner when the user fills out all the budget lines, encouraging their to send the consultation.

It also switches the color of the info popup to blue.


ferblape (approved):
👍

Issue #450, opened 2017-03-03
Consultation information screens


ferblape:
Right now there are two different screens to inform about the consultations:

- [x] merge both screens in a single one
- [x] use the consultation description as the content of the information page

PR #453, opened 2017-03-03
Remove intermediate screen and allow to configure the page of the summary

ferblape:
Connects to #450

### What does this PR do?

This PR updates consultation page to use the description as content, keeping the links to the next page.

Pending to think about: what to do with the images.


ferblape:
@furilo review if the feature is ok, and let me know if you want to solve the the images issue. I can upload them to an S3 storage.

furilo:
This can be merged. Image can be uploaded anywhere.

Issue #434, opened 2017-03-02
Show a banner in budget visualization if there is an active consultation


furilo:
- [x] Design 
- [x] Markup
- [x] Implement

furilo:
Once moved, we have to add in Sencelles Production the code in the theme_top_notice.html snippet.

PR #458, opened 2017-03-03
Move 

furilo:
Connects to #434

### What does this PR do?

Move the place in the html where site.configuration.head_markup is called, so we have flexibility to add things around the header.


ferblape (approved):
👍

Issue #461, opened 2017-03-03
Contrast ratio review


martgnz:
We should review the contrast ratio of every page to meet WCAG 2, while maintaining our branding:
- Text should be 4.5:1.
- Headlines, 3:1.

We can exclude disabled UI elements and logos.

## Homepage
![screen shot 2017-03-03 at 15 58 55-fullpage](https://cloud.githubusercontent.com/assets/1236790/23555741/5f2b6df4-002a-11e7-8cce-adc195b035ed.png)

## Consultations
![screen shot 2017-03-03 at 16 00 34-fullpage](https://cloud.githubusercontent.com/assets/1236790/23555808/9a80a0a4-002a-11e7-9bbb-e2c887f8672a.png)

## Cargos y agendas
![screen shot 2017-03-03 at 16 02 36-fullpage](https://cloud.githubusercontent.com/assets/1236790/23555891/df4c8a22-002a-11e7-81ad-dafb03ee24be.png)

PR #474, opened 2017-03-03
[WIP] Increase contrast ratio

martgnz:
A review of the contrast ratio in the Budget, Consultations and Indicators pages.
Everything is lighter but we don't pass WCAG 2 yet. Need to consider changing the main color.

Connects to #461.


ferblape:
Is this a WIP because not all the affected parts have been modified? Can we merge it?

martgnz:
It's WIP because we still need to review some colors; and hence it can affect the identity of Gobierto.

PR #475, opened 2017-03-03
Fix consultation markup

furilo:
Connects to none

### What does this PR do?

Prepares styles to enable markup in the budget consultation front page editor.


ferblape (approved):
👍

Issue #282, opened 2017-01-30
Download attachments on click


ferblape:
Now attachments are rendered inline, they should get downloaded when clicking.

PR #485, opened 2017-03-06
Download attachments

amaia:
Connects to #282

### What does this PR do?

Adds the original filename of the uploaded file to the file url (I've kept the uuid part of the filename to avoid name collisions).

Adds a `download`html5 attribute to links to attached files, so the browser downloads them instead of rendering them inline.

It also sets the `content_disposition` property to `attachment` for S3 uploads of files.

Check that the files are downloaded instead of rendered inline when clicking in the download links (in people's cv for example).

Check also that newly uploaded files are stored with the original filename after the uuid that there was before.

### DIFF
--- a/app/forms/gobierto_admin/file_attachment_form.rb
+++ b/app/forms/gobierto_admin/file_attachment_form.rb
@@ -21,7 +21,8 @@ module GobiertoAdmin
         site: site,
         collection: collection,
         attribute_name: :attachment,
-        file: file
+        file: file,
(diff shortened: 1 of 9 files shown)


amaia:
@ferblape I've removed the WIP from this one

ferblape (approved):
👍

ferblape:
Thanks @amaia. I have verified it works ok both in Safari and Firefox.

Issue #388, opened 2017-02-23
Load site corresponding to the URL when you access /admin


furilo:
If for example you type http://sencelles.gobify.net/admin you may end up in a site that is not Sencelles, an thus: 

- you may not be aware that you are not in the site you think you are
- when you realize it, or as a first action, you have to select the site you want to edit, which -I guess- can be easily inferred from the URL

PR #531, opened 2017-03-15
Load site from url in admin pages

amaia:
Connects to #388

### What does this PR do?

If the subdomain in the url is different than the current site it loads the site from the subdomain and redirects to the root admin page with that subdomain in the url. 

Paste an admin url with a site different from the current one and check that the site is changed in the admin page.
Check also if there are any weird redirections or inconsistent site from the page and the url.


ferblape (approved):
👍

Issue #359, opened 2017-02-16
Consultation Results report


furilo:
We are going to create a results report for the budget consultations. 

For the admin area:
- Entity officials (users in this context) should be able to export raw data in CSV. 
- We should provide some visualizations to let users understand results. Since this can be a huge task, lets define what we can do in a first iteration (it should be ready and working for March 5 aprox)

Data: 
- Participations (according to the interface we have already created)
- Participants (not all) will be segmented by sex, age, and hometown. In the reports we 

Initial ideas (just for discussion, not necessarily for v1):
- Participation overview: how many people, how are they (according to its segmentation), when they have participated (i.e.. 75% in the first day, etc)
- Opinions overview: opinions about each question (for Sencelles we'll probably have +30 - I have tried to convince them otherwise, but I failed). Enable filtering by segmentation, or plainly show results side by side for different (predetermined by us) demographics (ie. who young people thing vs. older people etc, male vs female...)
- Opinions analysis: % of positive|neutral|negative answers
- Equilibrium analysis: how many people tried to submit the consultation while being in deficit (this will take some effort to track, so definitively not for v1, although we may want to track it with Google Analytics just to check it by ourselves)

Also, this pontetially should also serve to make a public version of the report.

Let's add more ideas and narrow what we want to do for v1. We can design more and then only implement basic things, and leave other for the future.

cc @ferblape

PR #536, opened 2017-03-17
Consultation report

martgnz:
Connects to #359.

### What does this PR do?
This PR implements the new budget consultation report on the admin module.

It features statistics about the current consultations, a bar chart with participant's ages, and a table summarizing the response to each budget line proposal.

### Figures
- Total responses
- Gender breakdown
- Budget outcome summary
- Average age
- Location ranking
- Age report
- Budget line responses

### DIFF
--- a/app/assets/javascripts/admin/application.js
+++ b/app/assets/javascripts/admin/application.js
@@ -15,17 +15,6 @@
 
 
 $(document).on('turbolinks:load', function() {
-
-  if ($('#age_report').length) {
-    var vis_ageReport = new VisAgeReport('#age_report');
-    vis_ageReport.render();
-  };
-
-  if ($('#consultation-indicators').length) {
-    var vis_consultationIndicators = new VisIndicators('#consultation-indicators');
-    vis_consultationIndicators.render();
-  }
-
   if($('.air-datepicker').length){
     var start = new Date();
     start.setDate(start.getDate() + 1);
--- a/app/assets/javascripts/admin/application.js
+++ b/app/assets/javascripts/admin/application.js
@@ -16,6 +16,11 @@
 
 $(document).on('turbolinks:load', function() {
 
+  if ($('#consultation_report').length) {
+    var vis_ageReport = new VisAgeReport('#consultation_report');
+    vis_ageReport.render();
+  };
+
   if($('.air-datepicker').length){
     var start = new Date();
     start.setDate(start.getDate() + 1);
--- a/app/assets/javascripts/admin/application.js
+++ b/app/assets/javascripts/admin/application.js
@@ -21,6 +21,11 @@ $(document).on('turbolinks:load', function() {
     vis_ageReport.render();
   };
 
+  if ($('#consultation-indicators').length) {
+    var vis_consultationIndicators = new VisIndicators('#consultation-indicators');
+    vis_consultationIndicators.render();
+  }
+
   if($('.air-datepicker').length){
     var start = new Date();
     start.setDate(start.getDate() + 1);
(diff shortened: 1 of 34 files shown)


app/assets/javascripts/admin/application.js
```diff
@@ -16,6 +16,16 @@
 
 $(document).on('turbolinks:load', function() {
 
+  if ($('#age_report').length) {
```
furilo: We already have a admin/gobierto_budget_consultations/application.js - lets put this there.

app/assets/stylesheets/module-admin.scss
```diff
@@ -412,3 +412,141 @@ trix-toolbar .button_group button.icon, trix-toolbar .button_group input[type=bu
 	width: 2em;
 	font-size: 14px;
 }
+
+
+/* Consultations report */
+div[data-tab="consultation_report"] {
```
furilo: Tabs instead of spaces?

furilo: Look at the working integration, and you'll see that navigation is not with tabs but over full page reload, so this won't work. Just use a class name for the container.

app/assets/stylesheets/module-admin.scss
```diff
@@ -412,3 +412,141 @@ trix-toolbar .button_group button.icon, trix-toolbar .button_group input[type=bu
...
+			}
+		}
+		.figure-row {
+			padding-bottom: 0.25rem;
+			margin-bottom: 0.5rem;
+			border-bottom: 1px solid rgba(0,0,0,0.1);
+		}
+	}
+
+	#table_report {
```
furilo: This shouldn't go here. Look at comp-tables.scss - extend from there.

app/assets/stylesheets/module-admin.scss
```diff
@@ -412,3 +412,141 @@ trix-toolbar .button_group button.icon, trix-toolbar .button_group input[type=bu
...
+			.tick line {
+				stroke-dasharray: 2, 2;
+				stroke: $text_soft;
+			}
+		}
+		.bars rect {
+			fill: #00909E;
+		}
+	}
+	#consultation-indicators {
```
furilo: We are not reusing/extending existing styles from comp-indicator_widget?

app/assets/javascripts/admin/application.js
martgnz: I can't see that folder inside `javascripts/admin`. There's a `javascripts/gobierto_budget_consultations` folder, but it has the Vue app.

martgnz:
I extended the components and reorganized the JS and now there are few lines of custom styles in module-admin. I think this is ready for another review.

ferblape:
It's already integrated and ready to be reviewed in staging, i.e.: http://sencelles.gobify.net/admin/budgets/consultations/3/responses

(figures don't sum 100% because there are some wrong answers in staging)

I don't fully like the colors of the table, should we highlight the whole row on mouseover?

app/assets/stylesheets/comp-tables.scss
```diff
@@ -156,6 +156,9 @@ table.block_content {
 		color: transparentize($color_main, .5);
 		border: none;
 	}
+  th.center {
```
furilo, line 159: why this if we already have a .center class in populities?

furilo:
- [x] Print version is missing: Remove sidebar, make elements fit in the least possible space
- [x] Places don't fit in a laptop screen (see screenshot). Age distribution graphic would fit in less width; maybe we can create a 4 column grid, move the Place card to the second row, and have the age distribution in a 3 column width?
- [x] @ferblape: I would make the Report tab only appear once the consultation is finished, and it would be the first tab. 

captura de pantalla 2017-03-19 a las 11 06 33

martgnz:
The print version and the 4 column grid are ready now, I think. Could you check again, please?

Issue #464, opened 2017-03-03
Add a label to search forms


martgnz:
The search forms use the `placeholder` attribute to mark their action. We should use the `label` element too, as screen readers don't parse placeholders.

PR #540, opened 2017-03-21
Add more search labels

martgnz:
Connects to #464.

### What does this PR do?
This PR adds labels tag in the header's Search bar, the Budget module and the email box.

It also adds [a way](https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/) to hide content from the viewer, but not from a screen reader, inside Populities.


ferblape (approved):
👍

Issue #466, opened 2017-03-03
Use button elements instead of links


martgnz:
We should try to replace the links that act as buttons as much as possible. They are better supported by screen readers and keyboard actions become easier.

PR #541, opened 2017-03-21
Improve button declarations

martgnz:
Connects to #466.

### What does this PR do?
This PR changes the markup of nearly all the modules (and the subscription box) to add ARIA roles to the links which act as buttons.

### More info
It would be better to implement most of them as proper `<button>`'s. Unfortunately adding an anchor before of inside a button [isn't valid HTML5](https://stackoverflow.com/questions/6393827/can-i-nest-a-button-element-inside-an-a-using-html5).

At the end it's easier to take this approach (also [used by Bootstrap](https://getbootstrap.com/css/#buttons)).

Extra: improves the focus style on the Budget Consultations options.

### DIFF
--- a/app/assets/stylesheets/module-consultations.scss
+++ b/app/assets/stylesheets/module-consultations.scss
@@ -513,6 +513,11 @@ img.side_50 {
     padding: 0.75rem 1rem;
   }
 }
+.pct:focus {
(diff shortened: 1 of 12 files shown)


ferblape (approved):
👍

Issue #467, opened 2017-03-03
Add header to some tables


martgnz:
There are some tables that lack a `th` tag, as the one beneath the budget treemap in the budget module. We should add them, where possible.

PR #542, opened 2017-03-21
More table headers

martgnz:
Connects to #467.

### What does this PR do?
This PR adds more table headers in the Budget Module, improving the description of each column.

WIP because it needs integration from the translation side.


ferblape (approved):
👍

Issue #503, opened 2017-03-08
Make main menu accesible


martgnz:
Also, integrate keyboard navigation.

### Reference
- https://adobe-accessibility.github.io/Accessible-Mega-Menu/
- http://heydonworks.com/practical_aria_examples/

PR #543, opened 2017-03-21
Accesible menu

martgnz:
Connects to #503.

### What does this PR do?
This PR updates the main menu to make it accessible via ARIA roles. It also makes the switch from `menu` to `nav` in the collapsed and expanded navigation.

### DIFF
--- a/app/assets/javascripts/module-site_header.js
+++ b/app/assets/javascripts/module-site_header.js
@@ -1,43 +1,43 @@
 $(document).on('turbolinks:load', function() {
 
-	$('.nav_control, .close').click(function(e) {
-		e.preventDefault();
-		if( $('.menu.complete').css('display') == 'none' ) {
+  $('.nav_control, .close').click(function(e) {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/module-site_header.js
```diff
@@ -1,13 +1,18 @@
 $(document).on('turbolinks:load', function() {
 
-	$('.menu_control, .close').click(function(e) {
+	$('.nav_control, .close').click(function(e) {
 		e.preventDefault();
-		if( $('menu.complete').css('display') == 'none' ) {
-			$('menu.complete').velocity("transition.slideDownIn");
+		if( $('.menu.complete').css('display') == 'none' ) {
+      $('.nav').attr('aria-hidden', 'true');
```
ferblape: Are these tabs? Please, use spaces.

ferblape (approved):
Except for the tabs, the rest is OK!

PR #552, opened 2017-03-24
Add multitenancy & Pow clarifications to docs

amiedes:
Hello! I've been trying to understand how the different domains work and the Pow configuration and I thought it will be useful to add these clarifications to the docs, since the commands are not exactly the same as in the Docker setup.

### DIFF
--- a/docs/development-environment.md
+++ b/docs/development-environment.md
@@ -128,6 +128,26 @@ default: &default
   elastic_url: http://localhost:9200
 ```
 
+### Development top-level domain and port proxying
+
+Gobierto is a [_multi-tenant_](https://en.wikipedia.org/wiki/Multitenancy) application, designed to serve multiple *tenants* from the same instance of the application. A tenant is a group of users who share a common access with specific privileges to the software instance. In our case these tenants are the different cities using Gobierto.
+
+When running a single Gobierto rails application, you'll be able to access each of the available tenants throught a custom URL. The easiest way to deal with this is to use [Pow](http://pow.cx/)'s Port Proxying feature, as described on its [User's manual](http://pow.cx/manual.html#section_2.1.4).
+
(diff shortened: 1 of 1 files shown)


docs/development-environment.md
```diff
@@ -128,6 +128,26 @@ default: &default
...
+
+Gobierto is a [_multi-tenant_](https://en.wikipedia.org/wiki/Multitenancy) application, designed to serve multiple *tenants* from the same instance of the application. A tenant is a group of users who share a common access with specific privileges to the software instance. In our case these tenants are the different cities using Gobierto.
+
+When running a single Gobierto rails application, you'll be able to access each of the available tenants throught a custom URL. The easiest way to deal with this is to use [Pow](http://pow.cx/)'s Port Proxying feature, as described on its [User's manual](http://pow.cx/manual.html#section_2.1.4).
+
+Once Pow is installed, you have to configure it to work with Gobierto:
+
+```shell
+cd ~/.pow
+ln -s /path/to/gobierto/home
```
ferblape: Thank you @amiedes, I had this in my todo list but I didn't have time!

There are two ways of configuring Pow, and in this one you describe Pow acts as also as application server, so you don't need to run a Rails server with `rails s`. Drawback: a `debugger` statement doesn't stop execution.

There's in an alternative way which is just doing `echo 3000 > ~/.pow/gobierto`. In that case, you need to run `rails s` but gobierto will be visible in `http://gobierto.dev`.  I prefer this second way because the `debugger` works and is the setup. 

Test it if you have time, and we can discuss about it on Monday!

ferblape (changes requested):

docs/development-environment.md
amiedes: I didn't know about that 😮

Thanks @ferblape, I'll take a look!

ferblape (approved):
👍

Issue #544, opened 2017-03-22
Remove h1 tag from header layout and review the rest of headings


ferblape:
1. Remove `h1` tag from the header element in the layout

2. Review the rest of headings and add the `h1` when needed

PR #553, opened 2017-03-24
Review header levels

martgnz:
Connects to #544.

### What does this PR do?
This PR switches the first heading level from the city name to each section title (featured in the breadcrumb).

Extra: adds a missing hidden class to a table heading.

The header levels are section-focused now. You can check the structure using the [tota11y bookmarklet](https://khan.github.io/tota11y/) on each page.


ferblape (approved):
:+1:

PR #576, opened 2017-03-30
Don't require custom user records in responses created by admins

ferblape:
Unexpected

### What does this PR do?

This PR removes validations on custom user fields and user information in consultation responses  when created by an administrator


ferblape:
This is ready, @furilo

Issue #580, opened 2017-03-30
Extract mailer config to config file


amaia:
Set mailer configuration in config file and use from environments files to define the delivery method and necessary credentials

PR #583, opened 2017-03-31
Extract ActionMailer configuration

amaia:
Connects to #580

### What does this PR do?

Extracts the ActionMailer configuration to the config files and removes the setting from the `environments/*.rb` files. I also renames `config/initializers/amazon_ses.rb` to `config/initializers/action_mailer_settings.rb` to hold the more generic configuration, to support `:smtp` and `:ses` delivery methods.

Ansible roles updated in https://github.com/PopulateTools/populate-ansible/commit/8091e945d3a0c0d82573268ff18f6de5410fad2b


ferblape:
👍 Ansible executed in production, so le'ts release!

PR #587, opened 2017-03-31
Fix links for longer domains

amaia:
Unexpected

### What does this PR do?

The links were wrongly generated in the emails and some views in a gobierto installation using a base domain of more than 2 elements (go.dival.es vs. gobierto.es).

This PR changes the use of the url helpers to use the `:host` option instead of the `:domain` option, which doesn't depend on the configured tld length to work properly.

Check that the links in the emails and in the admin pages pointing to public pages are still correct for the current installation of gobierto.


ferblape:
👍 releasing!

Issue #593, opened 2017-04-03
Extract FileUploader adapter to a config setting


amaia:
Instead of specifying the adapter on each call to FileUploadService, extract it to a config setting, so it can be changed per installation. 

Possible values for now should be `:s3` and `:filesystem`.

PR #595, opened 2017-04-03
Extract FileUploadService adapter to config file

amaia:
Connects to #593

### What does this PR do?
Extracts the adapter for FileUploadService to the application config, but leaves `:s3` as the default.

Check that everything keeps working

Adds a new entry in `config/application.yml`:
```
  file_uploads_adapter: s3
```
the value can be `s3` or `filesystem`

It's not necessary to change Ansible roles because the default is `s3` as before, so nothing should break.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,4 @@ config/deploy/production.rb
 /coverage
 
 # Uploaded files
-/public/attachments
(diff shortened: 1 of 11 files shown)


ferblape (approved):
👍

PR #599, opened 2017-04-04
Compact smtp settings

amaia:
Connects to #580

### What does this PR do?

Compacts and symbolizes the keys to the smtp config introduced in #583 
Without this the keys for `mailer_smtp_settings` were being read as strings and adding them to `Rails.application.config.action_mailer.smtp_settings` didn't have any effect, as it didn't overwrite the default ones.


ferblape (approved):
👍

PR #602, opened 2017-04-04
586 - Sort past events backwards

### LINKED ISSUE #586: Agendas: past events should be ordered from most recent to least recent.
Now the oldest one appears on first place:

image

amiedes:
Fixes #586


ferblape:
@amiedes I have changed a couple of thinks:

- in the welcome controller just 10 comments were loaded, but in the rest of controllers loading events there was no limit
- if we are showing past events as fallback, don't show the past events link

Issue #560, opened 2017-03-27
People: Enable activation of module's subsections


furilo:
## In the Admin

* [ ] In the /admin/people/configuration: make the "Preferences" tab appear the first one
* [ ] New options block in /admin/people/configuration/settings
  **Secciones activas**
  - [ ] Cargos
  - [ ] Agendas
  - [ ] Blogs
  - [ ] Declaraciones

## In the front 

- If only one section is active: 
  - [ ] Menu main title: Only show the section name (not the module name). ie. **Inicio / Agendas** instead of **Inicio / Altos cargos y agendas / Agendas**
  - [ ] Menu main subsections: Don't show any
- Taking into account which sections are active: 
  - [ ] Show only them in main menu subsections, and in person sidebar

PR #609, opened 2017-04-06
Activate module subsections

amiedes:
Connects to #560 

### What does this PR do?

This PR implements the possibility to activate only certain submodules of GobiertoPeople: *Agendas*, *Officials*, *Blog* and *Statements*. This can be done from the `.../admin/people/configuration/settings` page.

The following changes were implemented:

* Certain components of the views are now placed inside conditionals so they can be displayed or hidden depending on the active submodules.
* Certain controller actions are only permitted to be executed if the corresponding submodule is active. Otherwise a redirection is made to `gobierto_people_root_path`.
* Settings and fixtures were added so all Gobierto People submodules are active by default.
* When only one Gobierto People submodule is active, its root path becomes the new Gobierto People welcome page. Also, navigation breadcrumbs will only show the submodule name instead of `Officials and Agendas / submodule`.
* The `.../admin/people/configuration` settings tab now appears first in the list and is selected by default.

Go to `.../admin/people/configuration/settings` and toggle the different submodules at your will. Then go to Gobierto as a regular user and check that only the desired submodules are being shown.

No

### DIFF
--- a/app/controllers/concerns/submodules_helper.rb
+++ b/app/controllers/concerns/submodules_helper.rb
@@ -1,0 +1,75 @@
+module SubmodulesHelper
+  extend ActiveSupport::Concern
+
+  included do
+    if respond_to?(:helper_method)
+      helper_method :active_submodules, :welcome_submodule_active?, :officials_submodule_active?, :agendas_submodule_active?, :blog_submodule_active?, :statements_submodule_active?, :submodule_path_for, :submodule_title_for, :submodule_controller_for
+    end
+  end
+
+  private
+
+  def available_submodules
+    {
+      officials: {
+        root_path: gobierto_people_people_path,
+        layout_title: t('gobierto_people.layouts.menu_subsections.people'),
+        controller_name: 'people'
+      },
+      agendas: {
+        root_path: gobierto_people_events_path,
+        layout_title: t('gobierto_people.layouts.menu_subsections.agendas'),
+        controller_name: 'person_events'
+      },
+      blog: {
+        root_path: gobierto_people_posts_path,
+        layout_title: t('gobierto_people.layouts.menu_subsections.blogs'),
+        controller_name: 'person_posts'
+      },
+      statements: {
+        root_path: gobierto_people_statements_path,
+        layout_title: t('gobierto_people.layouts.menu_subsections.statements'),
+        controller_name: 'person_statements'
+      }
+    }
+  end
+
+  def active_submodules
+    active_submodules = GobiertoPeople::Setting.where("key ~* ?", ".*_submodule_active").where(value: 'true', site: current_site)
+    active_submodules_names = active_submodules.pluck(:key).map { |key| key.gsub("_submodule_active", "") }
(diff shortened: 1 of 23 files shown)


app/controllers/concerns/submodules_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+        layout_title: t('gobierto_people.layouts.menu_subsections.statements'),
+        controller_name: 'person_statements'
+      }
+    }
+  end
+
+  def active_submodules
+    active_submodules = GobiertoPeople::Setting.where("key ~* ?", ".*_submodule_active").where(value: 'true', site: current_site)
+    active_submodules_names = active_submodules.pluck(:key).map { |key| key.gsub("_submodule_active", "") }
+    active_submodules_names & available_submodules.keys.map { |key| key.to_s }
```
ferblape: `map(&:to_s)`

app/controllers/concerns/submodules_helper.rb
```diff
@@ -0,0 +1,75 @@
+module SubmodulesHelper
+  extend ActiveSupport::Concern
+
+  included do
+    if respond_to?(:helper_method)
```
ferblape: Needed?

ferblape (approved):
This looks good to me! Please, update the description of the PR and release to staging!

amiedes:
@ferblape the last innocent commit broke the build. I think it's a Circle-CI's fault since tests pass on local. Is it possible that I get permissions to re-trigger tests without having to push again?

captura de pantalla 2017-04-07 a las 10 28 34

Should I deploy to staging anyways?

ferblape:
Yes, you can deploy to staging.

ferblape:
For this PR to be 100% complete, you need to include a migration script for existing sites. Something like this:

```ruby
Site.all.each do |site|
  if site.configuration.gobierto_people_enabled?
    GobiertoPeople::Setting.create site: site, key: "agendas_submodule_active", value: "true"
    GobiertoPeople::Setting.create site: site, key: "officials_submodule_active", value: "true"
    GobiertoPeople::Setting.create site: site, key: "blog_submodule_active", value: "true"
    GobiertoPeople::Setting.create site: site, key: "statements_submodule_active", value: "true"
  end
end
```

This script needs to be run after the deploy of this code, right?

app/controllers/concerns/submodules_helper.rb
```diff
@@ -0,0 +1,73 @@
+module SubmodulesHelper
+  extend ActiveSupport::Concern
+
+  included do
+    helper_method :active_submodules, :welcome_submodule_active?, :officials_submodule_active?, :agendas_submodule_active?, :blog_submodule_active?, :statements_submodule_active?, :submodule_path_for, :submodule_title_for, :submodule_controller_for
```
ferblape, line 5: Now I'm checking this I'm not 100% happy  of not having a scope to GobiertoPeople scope in the methods, but let's see how it looks like when another module needs to be splitted in submodules. It's ok for the time being.

ferblape (approved):
🚀

Issue #590, opened 2017-04-03
Implement a Rake task to import calendar events from Lotus Notes


ferblape:
We need to import calendar events from a Lotus Notes server and create Gobierto PersonEvent instances.

Lotus Notes offers a JSON API quite simple. Here's a link to the documentation: https://www-10.lotus.com/ldd/ddwiki.nsf/xpAPIViewer.xsp?lookupName=IBM+Domino+Access+Services+9.0.1#action=openDocument&res_title=Calendars_GET&content=apicontent

Here's a Ruby script that connects to the API endpoint and logins using Mechanize:

```ruby
require "mechanize"
require "json"

a = Mechanize.new { |agent|
  agent.user_agent_alias = 'Mac Safari'
}
a.agent.http.verify_mode = OpenSSL::SSL::VERIFY_NONE

a.get(API_URL) do |page|
  api = page.form_with(action: "/names.nsf?Login") do |form|
    form.Username = USERNAME
    form.Password = PASSWORD
  end.submit
  puts JSON.parse(api.body)
end
```

The API provides a different endpoint for each calendar person, so in Gobierto we are going to map an endpoint to a person.

From Lotus Notes events import:

- ID to perform syncrhonizations
- Event title
- Dates
- Notes, descriptions or any other interesting information

**All imported events should be created as draft, for an admin to accept them before being published**

The rake task should be site agnostic, therefore should be invocked with a parameter `SITE_ID`, meaning the ID of the site where events will be created.

The task will be invoked periodically in a cron task (maybe each hour) so it should deal with updates, and compare each event with the event from the source.

PR #612, opened 2017-04-11
590 - Lotus Notes API

amiedes:
Connects to #590 

### What does this PR do?

This PR adds the ability to synchronize person events from an IBM Notes calendar. Only public events are imported into Gobierto.

For the IBM Notes integration to work, the following steps are needed:

* In `admin/people/configuration/settings` select *IBM Notes* Calendar Integration
* Create the following settings for the site which is going to use this integration:

```ruby
GobiertoPeople::Setting.create! key: 'IBM_NOTES_ENDPOINT_USR', value: 'the_user', site: your_site
GobiertoPeople::Setting.create! key: 'IBM_NOTES_ENDPOINT_PWD', value: 'the_password', site: your_site
```
* (Optional) Associate a calendar URL for each of the people of the site. In the future, this will be done from the admin GUI, but right now this can only be done manually:

```ruby
c = GobiertoPeople::PersonLotusNotesCalendarConfiguration.new person_id: person_id
c.data = {endpoint: 'https://host.wadus.com/calendar_url' }
c.save!
```

The events synchronization is run from a Rake task:

```
bin/rails gobierto_people:sync_agendas
```


app/services/gobierto_people/agendas_integration.rb
```diff
@@ -0,0 +1,13 @@
+module GobiertoPeople
+  class AgendasIntegration
```
ferblape: What about renaming this class to `RemoteCalendars`?

app/services/gobierto_people/agendas_integration.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    def self.sync_all_agendas
```
ferblape: What about renaming this method to `sync`?

- we don't need agendas/calendars because it's implicit in the class name
- we don't need the all because is a method class and is implicit too

app/services/gobierto_people/agendas_integration.rb
```diff
@@ -0,0 +1,13 @@
...
+      Site.with_agendas_integration_enabled.each do |site|
+        calendar_integration = site.calendar_integration
+
+        site.people.with_calendar_configuration.each { |person| calendar_integration.sync_person_events(person) }
```
ferblape: A bit long for a one liner

lib/lotus_notes/api.rb
```diff
@@ -0,0 +1,30 @@
+require 'mechanize'
+
+module LotusNotes
+  class Api
+
+    def self.get_person_events(params)
+      response_page = make_request(params)
+      person_events = JSON.parse(response_page.body)["events"]
```
ferblape: You don't need the assignment. Also, deal with errors and take into account when the page hasn't events (return empty array).

lib/lotus_notes/api.rb
```diff
@@ -0,0 +1,30 @@
...
+    def self.make_request(params)
+      signin_page = agent.get params[:endpoint]
+      page = signin_page.form_with(action: "/names.nsf?Login") do |form|
+        form.Username = params[:username]
+        form.Password = params[:password]
+      end.submit
+    end
+
+    def self.agent
+      agent = Mechanize.new
```
ferblape: Memoize the `agent`.

lib/lotus_notes/api.rb
```diff
@@ -0,0 +1,30 @@
...
+      response_page = make_request(params)
+      person_events = JSON.parse(response_page.body)["events"]
+      person_events
+    end
+
+    private
+
+    def self.make_request(params)
+      signin_page = agent.get params[:endpoint]
+      page = signin_page.form_with(action: "/names.nsf?Login") do |form|
```
ferblape: You don't need the assignment to `page`.

app/models/gobierto_people/lotus_notes_calendar_integration.rb
```diff
@@ -0,0 +1,37 @@
+module GobiertoPeople
+  class LotusNotesCalendarIntegration
```
ferblape: model or service?

lib/lotus_notes/person_event.rb
```diff
@@ -0,0 +1,83 @@
+module LotusNotes
+  class PersonEvent
```
ferblape: This class does two things:

1 - wraps the results of the request into an object
2 - syncs with PersonEvent

Given that this is a lib class, it should be agnostic about PersonEvent, so my suggestion is remove the synchronization part from here and move it to another service class or to `LotusNotesCalendarIntegration`

Issue #508, opened 2017-03-10
People admin: enable load attachment to a content block item


furilo:
To an actual item, not the block

- [x] Markup
- [ ] Implementation

furilo:
@martgnz I need you to take a look at this because I can't figure it out!

furilo:
sorry @martgnz wrong issue

PR #654, opened 2017-04-21
508 - Attachment for content block

furilo:
Connects to #508 

### What does this PR do?

Provides de UI controls to

- upload a file attached to a row from a content block
- show the corresponding download file control in the content block in the front (warning: I've hardcoded the control, it should be properly integrated)


amiedes:
Don't merge, I just found some failing edge cases.

amiedes:
Hey @ferblape, the backend of this issue is ready. I'm not fully convinced about the two-step initialization for FormObjects:

```ruby
@person_form = PersonForm.new(id: params[:id], admin_id: current_admin.id, site_id: current_site.id)
@person_form.assign_attributes(person_params)
```
because every controller that instantiates form objects that have content block records with attachments must be aware of this. What do you think about correcting this after we make the transition to the "Attachment model"?

ferblape:
A few comments:

- [x] you need to add the validation of presence of site in the `PersonStatementForm`
- [x] When the content block record doesn't have an attachment don't show delete attachment checkbox
- [x] When a new attachment has been added, and the Save button clicked, show something in the content block record for the user to understand there's an attachment being uploaded
- [x] When a content block record is being edited and the attachment is marked to be removed, after the save button is clicked, remove the attachment from the content block record row
- [x] Remove the getter of `site_id`, so the user of the Form object is forced to set it in the initializer

And I agree with your comment, although I don't see an easy solution.

ferblape (approved):
+1

ferblape:
@furilo, please, review in staging, specially the part of deleting an existing attachment from a content block record.

furilo:
- [ ] Admin: There is a string (Attachment file) which is not getting translated: 
captura de pantalla 2017-06-02 a las 15 12 32

furilo:
Ready to merge once the string issue is fixed.

amiedes:
Done!

amiedes:
Wait, I'll need to fix the unused translation keys

ferblape:
I'll take care.

PR #655, opened 2017-04-21
621 - Fix calendar flacky test

amiedes:
Connects to #621 

### What does this PR do?

Change test year to past so it does not collide with events from fixtures


ferblape (approved):
Very pragmatic

Issue #628, opened 2017-04-18
Open data page should take into account enabled submodules


ferblape:
GobiertoPeople open data page should take into account which modules are enabled and display only those.

PR #657, opened 2017-04-21
628 - Don't show data download links for disabled submodules

amiedes:
Connects to #628 

### What does this PR do?

Hides download links in Gobierto exports section for disabled Gobierto People submodules


ferblape (approved):
👍

Issue #627, opened 2017-04-18
Add IBM Notes credentials to Gobierto People configuration screen


ferblape:
Right now, IBM Notes credentials are stared as GobiertoPeople::Setting. Move these credentials to the new GobiertoModuleSettings module, and add the proper controls to the GobiertoPeople > Configuration page.

After this issue has been completed, we should review if GobiertoPeople::Setting is used elsewhere and remove the model, the fixtures and the table.

amiedes:
Working on this issue I noticed an "edge case" which I consider important to solve. In the `GobiertoPeople::IbmNotes::CalendarIntegration#sync_person_events` method, after making the API call to get the events, we mark as *pending* all the future events which are still present in our DB but which are not being returned any more by the API, thinking that it's because they have been deleted from the IBM Notes GUI.

```ruby
# app/services/gobierto_people/ibm_notes/calendar_integration.rb
def self.sync_person_events(person)

  # ...

  received_events_ids = received_events.map { |event| event['id'] }

  person.events.synchronized_future_events.each do |future_event|
    unless received_events_ids.include?(future_event.external_id)
      future_event.pending!
    end
  end
end
```

But if the service is unaivailable or because the credentials are incorrect, the api call will return an empty array: 

```ruby
def self.get_person_events(params)
  response_page = make_request(params)
  begin
    JSON.parse(response_page.body)["events"]
  rescue JSON::ParserError
    []
  end
end
```

**causing all future events of this site to be marked as pending in our DB.**

The problem is that given how the authentication to the API works (using Mechanize), it's not very clear how to differentiate between the cases where the credentials are wrong, the service is down or there are no events (I think in this case a blank HTML page is rendered instead of returning an empty array of events in the JSON data).

Also, when credentials are incorrect, the HTTP response status code is `200`, so it's not possible to use status codes to reason about the result of the request.

One way to solve this is to make this distinction by using Mechanize again to parse the returned page and see if a blank page is returned or we've been redirected to the login form again (a "wrong credentials" message is displayed in this case). The problem with this approach is that it's *very specific* to the site using IBM Notes calendar integration.

Any advice on this, @ferblape ?

ferblape:
Good catch @amiedes! You are totally right, and we need to deal with those exceptions.

There are two scenarios we need to detect:

1. invalid credentials
2. invalid response body

This is what I suggest for the `IbmNotes::Api` class:

```ruby
module IbmNotes
  class InvalidCredentials < StandardError; end
[...]

PR #662, opened 2017-04-24
627 - Set IBM Notes credentials from admin interface.

amiedes:
Connects to #627 

### What does this PR do?

This PR adds the possibility for a site admin to set IBM Notes credentials from a GUI.

Also, credentials are now stored as `GobiertoModuleSettings` instead of
`GobiertoPeople::Setting`. This commit contains a migration to perform
the corresponding data transformations.

Also the exception catching / error handling for edge cases where the API server may be down, no events are returned or the provided credentials are wrong.

IBM Notes integration should keep working after this update.

No


ferblape:
Why is this branch not in master (and deployed to staging) and why is a WIP? Everything that is required to be reviewed should be in master, and everything that is a WIP should be explicit about why is WIP.

app/views/gobierto_admin/gobierto_people/configuration/settings/edit.html.erb
```diff
@@ -39,6 +39,17 @@
...
       </div>
+
+      <div class="form_item input_text">
+        <%= f.label :ibm_notes_usr, t(".ibm_notes_usr") %>
+        <%= f.text_field :ibm_notes_usr, placeholder: t(".placeholders.ibm_notes_usr") %>
+      </div>
+
+      <div class="form_item input_text">
+        <%= f.label :ibm_notes_pwd, t(".ibm_notes_pwd") %>
+        <%= f.text_field :ibm_notes_pwd, placeholder: t(".placeholders.ibm_notes_pwd") %>
```
ferblape: This should be a password_field, to be more private.

app/forms/gobierto_admin/gobierto_people/settings_form.rb
```diff
@@ -65,6 +75,14 @@ def save_settings
           settings_attributes.home_text_en = home_text_en
           settings_attributes.submodules_enabled = submodules_enabled.select{|m| m.present?}
           settings_attributes.calendar_integration = calendar_integration
+
```
ferblape: I think we should add a validation in the form: if the user selects IBM Notes integration but user and password are empty, we should display an error message that credentials are required.

app/views/gobierto_admin/gobierto_people/configuration/settings/edit.html.erb
```diff
@@ -39,6 +39,17 @@
...
+      </div>
+
```
ferblape: Are you going to add the javascript code to clean the form? Is that why it's a WIP?

ferblape (changes requested):
Code looks good. Please, add the changes I have requested and feel free to merge in master again and release to staging.

amiedes:
I marked this as WIP because wanted to add also the JS code and the solution for the issues exposed in #627. I didn't deploy to staging because I wanted you to take a look at the migration script previously.

I'll try to be explicit about this things from now on 👌

amiedes:
Deployed to staging 🚀 

The migration worked correctly but since it only deleted "deprecated" `GobiertoPeople::Setting` containing the key `IBM_NOTES_etc.`, I had to delete old `LOTUS_NOTES_etc.` settings manually from the console.

A few notes:

* I think the feedback provided by the error messages shown in the interface can be improved. Right now it only says "An error occured when saving this settings", but I couldn't guess how to customize it to say "Credentials can't be blank" etc.

ferblape (approved):

Issue #647, opened 2017-04-20
If only Agendas submodule is enabled, redirect to agendas when viewing person


furilo:
Now you reach http://santboi.gobify.net/cargos-y-agendas/cargos/13 when you click on a person's name. But this page will be mostly empty. I suggest we redirect to Agenda -> http://santboi.gobify.net/cargos-y-agendas/cargos/13/agenda/eventos

PR #663, opened 2017-04-24
Redirect person show to person events index when only agendas submodule is active

amiedes:
Connects to #647 

### What does this PR do?

Redirect person show to person events index when only agendas submodule is active.

The URLs specified in the issue should redirect to the same webpage (already deployed in staging).

No


ferblape (approved):
👍

PR #664, opened 2017-04-25
176 - Fix admin widget save stick in parent

### LINKED ISSUE #176: Admin UI > stick_in_parent is not working for the tool sidebar
The right column (with .stick_in_parent) should stick to top on scroll. Is not happening. No JS error. Maybe have to do with the flex box declarations of the parent, but removing them makes no effect.

furilo:
Fixes #176


ferblape:
👍

PR #690, opened 2017-05-04
Solve IBM Notes event ids problem

amiedes:
Connects to #678 

### What does this PR do?

This PR solves the current problems related to the wrong understanding of the IBM Events IDs and its relationship with recurrent events.

Now event information is retrieved from `GET calendar event <event_id>`, and the events collection endpoint is used merely as an intermediate step to get the `hrefs` of the events. 

Before making a new synchronization with the changes made in this PR, we need to delete the current events in the DB for the persons which are using the IBM Notes calendar integration. This can be achieved with:

```ruby
# (!) Make sure `Site.fifth` is the correct target site
site = Site.fifth

person_1 = site.people.first
person_2 = site.people.second

person_1.events.where("external_id IS NOT NULL").destroy_all
person_2.events.where("external_id IS NOT NULL").destroy_all
```

Events should now be synchronized correctly. Several instances of the same recurring event should behave for Gobierto as independent events.

No


ferblape:
@amiedes, take a look to the refactor and the new method names:

- basically names should be simpler and should be natural, nobody referes to links as hrefs, they are just links (href is an implementation detail)

- events_collection is redundant, is just event_collection or just events

- in class methods `private` keyword doesn't work, you need to use `private_class_method` keyword

Issue #637, opened 2017-04-19
Implement Google calendar integration


ferblape:
A few links of this integration:
- overview: https://developers.google.com/google-apps/calendar/concepts/
- authorization: https://developers.google.com/google-apps/calendar/auth
- sync: https://developers.google.com/google-apps/calendar/v3/sync
- push notifications: https://developers.google.com/google-apps/calendar/v3/push

PR #703, opened 2017-05-20
Google Calendar integration

ferblape:
Connects to #637, #705 and #704

### What does this PR do?

This PR introduces Google calendar integration to GobiertoPeople module.

It also changes the Person - Event relationship, because didn't consider in our first implementation how attendees could be invited to events which creator (organizer) is not part of the Gobierto module.

So, now PersonEvents controller shows the events the person is attending (instead of just the events she's organizing). As a consequence, each event created has at least one attendee, which is the organizer.

In the future, because of this decoupling we'll be able to add a Calendar model which doesn't need to belong to a person but to something more generic.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,9 +7,6 @@ HOST=gobierto.dev
 PORT=3000
 RAILS_MAX_THREADS=5
 
-# Rails test log level
-TEST_LOG_LEVEL=fatal
-
 # Poltergeist debug and inspector variables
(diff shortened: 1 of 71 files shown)


furilo:
I just added the integration one, but the notification system seems to have go nuts:

http://madrid.gobify.net/cargos-y-agendas/personas/alvaro-ortiz

captura de pantalla 2017-05-21 a las 9 03 52

Issue #706, opened 2017-05-21
Let the user select which calendar to sync


furilo:
In my testing of the GCal integration, I was expecting that it would let me choose which calendar to sync. It didn't let me, and it ended up importing all my events from all my calendar. We have to let the user choose which calendar to sync. Is this possible?

Ideally, to choose several with checkboxes. 

#703

PR #711, opened 2017-05-23
Pick calendars from Google Calendar

ferblape:
Connects to #706

### What does this PR do?

This PR allows a person and an admin to pick which calendars are synchronized from Google Calendar.


ferblape:
Maybe the interaction in the admin can be improved, because when you wan to remove the integration we are using a checkbox. 

This is because it's simpler to reuse an existing action in the controller than create a new one.

Issue #722, opened 2017-05-25
Integrate GobiertoAttachments in GobiertoCMS::Page resource


ferblape:
After #721 we are going to implement an UI to list and manage attachments of a resource

This is how the UI will look like:

Invision: https://projects.invisionapp.com/d/main#/console/11109218/234995489/preview

The UI will be implemented using Vue.js and the attachments API.

The first resource with attachments will be a CMS page.

PR #727, opened 2017-05-31
Gobierto Attachments UI

martgnz:
Connects to #722.

### What does this PR do?
This PR implements the Gobierto Attachments UI in Sandbox, as a small Vue app.

I'm rendering the partial at `sandbox/admin_users_user.html`.


app/assets/stylesheets/comp-modals.scss
```diff
@@ -31,4 +31,150 @@
 .mfp-close-btn-in .mfp-close:hover {
 	color: #FFF;
 	border-radius: 0;
-}
\ No newline at end of file
+}
+.modal-attachments {
```
furilo, line 35: modal-attachments o something more generic?

martgnz: Maybe modal-slim? The default modal has a different look (with more padding), so I didn't want to break styles used elsewhere.

ferblape:
This can be tested in staging with a few things still pending to discuss with @martgnz:

- [x] when there are no attachments, the height of the element is not enough
- [x] we need to discuss how to fix the markup in the edit attachment form
- [x] style an error message (try to upload the same attachment twice to get an error message)

furilo:
Regarding design: 

**Sidebar**

- [x] "Adjuntos:" Enclose in a `<small>`
- [x] Button: Let's not create a custom style, in any case lets add a generic button style in comp-forms.scss. We can simply use 'button light small'. Same for the button inside the 
- [x] Is a table needed to list the files? It's not tabular data, just a list.
- [x] List items: let's reduce the text size.
- [x] Do we need the padding for the items?
- [x] When editing a file, the modal is not properly displayed. The default `modal mfp-hide` classes are there? How are we building that modal?

**Add file modal**

- [x] Weird column widths. I would not use things like `.modal-attachments table tr td:nth-child(2), .modal-attachments table tr td:nth-child(3)` as you are doing. Let them flow. 
- [x] When dragging and dropping, styles don't change. We would need to show an #000 opacity .5 full screen overlay and some message?
- [x] @martgnz let's not customize the styles for the forms. What's the need?
- [x] After I have dropped a file, the height of the modal is less than the height of the content. Again, why are not we using the default system for modals that we already have in place? It's because of Vue?

furilo:
@ferblape what do you mean re: `we need to discuss how to fix the markup in the edit attachment form`? Do you mean the markup we generate when including a file in the main textarea?

Also: 

- [x] If I click on a file that is in the sidebar, the page reloads itself but nothing happens. Should we send the file to the textarea?

ferblape:
@furilo I meant to fix the edit form.  I guess this is a markup issue because there is no background color:

screen shot 2017-07-07 at 11 43 49

furilo:
This is because of the modal issues. How are we generating them? Does Vue allow to keep using jquery.magnific-popup ?

martgnz:
The modal is being generated as a vanilla div, like [this example](https://vuejs.org/v2/examples/modal.html). I am looking at a Vue + magnific-popup integration, but it's another layer of abstraction.

After looking at that example, wouldn't be better to fix the styles and reuse the ones we have instead of building the modal with magnific popup? (the modal works, the problem is the css).

martgnz:
@ferblape switched to MagnificPopup and the styles should be OK now. Let's review tomorrow.

ferblape (approved):
This is OK to me

furilo:
More things: 

- [x] I wouldn't make mandatory the title of the file. 
- [x] If an error gets generated and you keep doing things, the error keeps being there all the time. It should go away at some time.

furilo:
OK let's finish (ie merge) this iteration and create other issues for further improvements.

ferblape:
🚀

PR #734, opened 2017-06-03
Compatible migrations rails 5

ferblape:
Unexpected

### What does this PR do?

There were two issues with migrations:

- [ ] all of them need to be versioned - http://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html
- [ ] the format of the schema has changed


ferblape:
Check this @amiedes !

amiedes:
Cool

Issue #721, opened 2017-05-25
GobiertoAttachments model


ferblape:
We are going to implement a new module called `GobiertoAttachments`. This is a cross module that provides attachment features to any resource, such as CMS pages, People, Events... See [how to create a new module](https://github.com/PopulateTools/gobierto/blob/master/docs/developing-module.md)

The basic model of this module is `GobiertoAttachments::Attachment`, with the following attributes:

- `name`: string, not null
- `attachment_type`: string, not null. Ej: PDF, JPG (it can be extracted from the file extension)
- `file_size`: integer, bytes, not null
- `file_name`: string, not null
- `url`: string, not null
- `description`: text
- `current_version`: integer, not null

Relations:

- `versions`: old versions of the attachment (using [paper_trail](https://github.com/airblade/paper_trail)?)
- `attachables`: a has and belongs to many polymorphic relationship

Notes:

- The only change that produces a new version of an attachment is a change in the file. It means:
  - changing the name or the description of a attachment doesn't increases the version
  - changing the file **does** increase the version
  - to track these changes maybe we could store the md5 of the attached file

- It might make sense that this model uses the FileUploadService to upload attachments

PR #737, opened 2017-06-05
721 - GobiertoAttachments::Attachment model

amiedes:
Connects to #721 

### What does this PR do?

* Creates `Attachment` and `Attaching` models, to be used from now on in Gobierto as the default way of using attachments.
* Attachments now support versioning thanks to [paper_trail](https://github.com/airblade/paper_trail)

No


ferblape (approved):
👍

Issue #723, opened 2017-05-26
GobiertoAttachments API


ferblape:
After #721 and for #722 we need an API to integrate with the Vue component of the attachments.

The API will need to:

1. list the attachments of a given site (the last version of the attachment)
2. search attachments
3. upload a new attachment to the site
4. update an attachment
5. remove an attachment
6. associate an attachment to a resource
7. list the attachments of a resource
8. de-associate an attachment to a resource

This will be a private REST API. To authenticate the API we'll use a token, maybe a plain token, or maybe a JSON token.

PR #747, opened 2017-06-09
723 - Attachments admin API

amiedes:
Connects to #723 

### What does this PR do?

Creates simple API to manage attachments and attachings from admin windows.

**NOTE**: The `GobiertoAttachments::Attachment` Algolia index must be created in production.


test/integration/gobierto_admin/gobierto_attachments/api/attachments_test.rb
```diff
@@ -0,0 +1,247 @@
...
+
+        assert_raises ActiveRecord::RecordNotFound do
+          ::GobiertoAttachments::Attachment.find(pdf_attachment.id)
+        end
+      end
+
+      def test_attachments_destroy_error
+        login_admin_for_api(admin)
+
+        unexistent_id = 666
```
ferblape, line 263: :six::six::six:

ferblape:
Good work @amiedes! I have done a quite exhaustive testing with really bad ideas in my mind and everything worked as I expected. 

There are just a few things that need to be improved a bit:

- [x] When an Attachment is going to be created and already exists in the same site other with the same content (same md5) I think we should try to avoid duplicates and return something different to 200 OK. Maybe a validation error and include in the error message the link to the existing resource?

- [x] Capture this exception: https://rollbar.com/fernando/gobierto/items/437/

- [x] Capture this exception (missing file field in the body of the create request): https://rollbar.com/fernando/gobierto/items/436/

- [x] There's no action show in attachments! We should implement it

- [x] This request `http://santboi.gobify.net/admin/attachments/api/attachments?attachable_type=GobiertoCms::Page&attachable_id=8` returns a 200 OK and responds in HTML

app/controllers/gobierto_admin/gobierto_attachments/api/attachments_controller.rb
```diff
@@ -0,0 +1,97 @@
+module GobiertoAdmin
+  module GobiertoAttachments
+    module Api
+      class AttachmentsController < ::GobiertoAdmin::Api::BaseController
+
+        before_action :find_attachable, only: [:index]
+        before_action :find_attachment, only: [:update, :destroy]
+
+        def index
+          if params[:search_string]
```
ferblape: `attachments = if ... elsif .. else`, to avoid the assignment of the variable three times

app/controllers/gobierto_admin/gobierto_attachments/api/attachments_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+            each_serializer: default_serializer
+          )
+        end
+
+        def create
+          attachment = ::GobiertoAttachments::Attachment.new(
+            attachment_params.merge!(site: current_site, file: uploaded_file)
+          )
+
+          attachment.save!
```
ferblape: new + save! = create!, right?

app/controllers/gobierto_admin/gobierto_attachments/api/attachments_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+          attachable_type = params[:attachable_type]
+
+          if attachable_id && attachable_type && ::GobiertoAttachments.permitted_attachable_types.include?(attachable_type)
+            attachable_class = attachable_type.constantize
+            @attachable = attachable_class.find_by!(site: current_site, id: attachable_id)
+          end
+        end
+
+        def uploaded_file
+          @tmp_file = Tempfile.new('attachment_file')
```
ferblape: Why instance variable?

app/controllers/gobierto_admin/gobierto_attachments/api/attachments_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+          @tmp_file.binmode
+          file_content = Base64.decode64(attachment_params[:file])
```
ferblape: We should use `strict_decode64`

ferblape (changes requested):

app/models/gobierto_attachments/attachment.rb
```diff
@@ -1,17 +1,35 @@
...
+
     attr_accessor :file
 
     validates :site, :name, :file_size, :file_name, :file_digest, :url, :current_version, presence: true
 
+    validates :file_digest, uniqueness: {
+      scope: :site_id,
+      message: ->(object, data) do
+        url = object.site.attachments.find_by!(file_digest: object.file_digest).url
+        "not unique (already uploaded at #{url})."
```
ferblape: Translate this message

app/models/gobierto_attachments/attachment.rb
```diff
@@ -24,24 +42,35 @@ def extension
...
+      true
+    end
+
     private
 
     def update_file_attributes
       if file
-        new_digest = Attachment.file_digest(file)
+        if file.size > MAX_FILE_SIZE_IN_BYTES
+          errors.add(:base, "File exceeds max size")
```
ferblape: Translate this message, please

ferblape (changes requested):
We need to translate error messages
Also, we should translate Attachment attributes (file size, file name, file digest...)

amiedes:
@ferblape, does it make sense to make the translations for an admin API, and given that the only reason why the locale is available is because we haven't yet implemented a token-based authorization?

ferblape:
Well, the API should be available with any locale, no matter if it's an admin api. Imagine a customer building their website using Gobierto events, and they want the site to be in Catalan.

ferblape:
Also, consider I'm going to be using the API to display error messages to the administrators in the frontend. If the UI is in Spanish and the error message in English, the experience will be bad.

amiedes:
@ferblape Done!

ferblape (approved):

Issue #710, opened 2017-05-23
New GobiertoBudgets data endpoint


ferblape:
We need a new API endpoint with data for the GobiertoBudgets module.

This endpoint needs to return forecast data from **all available years** and **income / expense** in the following format:

```json
[
{
"id":"34",
"budget_category":"expense",
"level_1":"Producción de bienes públicos de carácter preferente",
"level_2":"Deporte",
"value":151965589.0,
"per_inhabitant":46.54,
"pct_section":0.03,
"pct_diff":0.0,
"year":2010
}
...
]
```

Issue #700, opened 2017-05-17
Bubbles for budget vis home


furilo:

furilo:
A couple of issues: 

- [x] The Turbolinks thing of duplicating the D3 if you use the browser history buttons (its tricky, it happens the second time you go back, not the first one)
- [x] If you click on the Debt bubble, you are redirected to another budget line (https://madrid.gobierto.es/presupuestos/partidas/1/2016/functional/G)

PR #748, opened 2017-06-16
Bubbles visualization

ferblape:
Connects to #700 and #710

### What does this PR do?

This PR implements a new budgets visualization for the homepage of Budgets module

After being released we need to run a task to calculate all the bubbles data for all the municipalities:

```
bin/rails gobierto_budgets:data:bubbles
```


furilo:
@martgnz some things: 

- [x] subtítulo ya no aplica: "Estos son los 5 principales ingresos y gastos de tu municipio."
- [x] falta leyenda para indicar qué significan los colores
- [x] quedamos en cambiar el cursor cuando estuvieses hover el timeline
- [x] when you click on a bubble, it should take you to the budget line
- [x] when you are hovering a bubble, I think we should explain what the up/down means. something like "mas/menos que el año anterior"

martgnz:
This is good to go IMHO! And already in staging 🚀

PR #751, opened 2017-06-20
Census build alternatives

apradillap:
Connects to Document number validation test

### What does this PR do?
- I have tried to eliminate complexity in the method **build_alternatives**
- First, I passed rubocop, then replaced the ifs with case
- I know the change of gsub by tr, decreases performance (https://github.com/JuanitoFatas/fast-ruby#stringgsub-vs-stringtr-code) but thanks to this, I found a bug. For example, "X05104959V" returned:
` 
["X05104959V", "X05104959", "XX51X4959V", "XX51X4959", "005104959V", "005104959", "514959V", "514959", "X05104959", "X05104959V", "X05104959", "005104959V", "005104959", "05104959V", "05104959"]
`
when the right thing had to be
`
 ["X05104959V", "X05104959V", "X05104959V", "X05104959V", "X05104959", "X05104959", "X05104959", "X05104959"] 
`

- I think it would be possible to extract part of the code to a method but the different options makes it inadvisable.

At first I saw it simple but when you get deep to understand it, I think being would be one of the most optimal solutions.

`
bin/rails test test/repositories/census_repository_test.rb
`

No


amiedes:
Thanks @apradillap! 🙌

Issue #772, opened 2017-06-29
Add budgets to global search


furilo:
In order to index Elasticsearch budget lines in Algolia, we're first going to implement the following:

* Add a `destroy` method in `BudgetLine`, which will:
  1. Remove the record from Algolia (use `AlgoliaClient`)
  2. Remove the record from ES
* Add a `save` method in `BudgetLine`, which will:
  1. Index the record in ES
  2. Index the record in Algolia (use `AlgoliaClient`)

Also:

* Create attribute accessors for all columns
* Store the ES mapping, ES index name and ES index type inside `BudgetLine` class methods

¿Maybe use `ActiveModel`?

PR #788, opened 2017-07-07
772 - Permits searching for budget lines in Algolia

amiedes:
Connects to #772 

### What does this PR do?

* Permits indexing instances of `BudgetLine` in Algolia.
* Refactors `BudgetLine` model, moving Elasticsearch-specific behavior to `budget_line_elasticsearch_helpers.rb` and Algolia-specific behavior to `budget_line_algolia_helpers.rb`.
* Provides a Rake task to destroy and reindex all of the `BudgetLine` records indexed in Algolia for a specific site, and reindexes them again.

Budget lines should be searchable by name and description.

### DIFF
--- a/app/assets/javascripts/module-search.js
+++ b/app/assets/javascripts/module-search.js
@@ -44,7 +44,11 @@ $(document).on('turbolinks:load', function() {
       case 'GobiertoCms::Page':
         return I18n.t("layouts.search.page_item");
       case 'GobiertoBudgets::BudgetLine':
-        return I18n.t("layouts.search.budget_line_item");
+        if (d['kind'] === 'I') {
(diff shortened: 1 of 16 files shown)


ferblape:
@amiedes I think we should add if the budget line is a expense or an income for the shake of clarity. It could be "Partida presupuestaria ingreso" and "Partida presupuestaria gasto"

Look at this case in staging:

screen shot 2017-07-10 at 18 54 15

amiedes:
@ferblape  @furilo 👍  BTW, do you want me to also add the budget line level to the search results? This way we can differentiate between budget lines with the same name but different level.

amiedes:
@ferblape  🚀

ferblape:
👍

Issue #792, opened 2017-07-10
Header borders don't show up on Chrome


martgnz:
Seems to be related with an upgrade. It works with Firefox and Safari.

PR #793, opened 2017-07-10
Fix header border color on Chrome

martgnz:
Connects to #792 

### What does this PR do?
This PR fixes the header border color, which wasn't showing up on latest Chrome.

screen shot 2017-07-10 at 14 52 33

I just transformed the rgba color to hex with [this converter](http://tdekoning.github.io/rgba-converter/).


ferblape:
👍

PR #798, opened 2017-07-13
Don't reduce years on timeline

martgnz:
### What does this PR do?
This PR stops reducing the years available on the timeline when switching to a past year.

Go to the budgets homepage, select a past year with the dropdown and you'll find the new year active, but the timeline is not reduced.


ferblape:
🚀

Issue #754, opened 2017-06-23
CMS / Collections


furilo:
We have to define the features for the next revision of the CMS, focusing on how it matches with the Participation requirements. Things we'll have (let's see afterwards _when_ we do it)

- global news section for participation
- news section inside a group
- probably it makes sense to have a global news section for the site
- an specific module could also be asking for a news section

This takes us to the idea of content collections or buckets. How I think it could work: 

- the basic model for a content piece is the same no matter where it is (be it from a participation group news section or the site global one): 
  - title, content, common editor, attachments, revision history if/when we implement it, placement in menus (when we have it), tags
- each of those sections could be a collection. We'd have a model for collections, which would have references to the individual content pieces it holds, and to the place/section it belongs. ie. The news section for a group/process would be a collection; the general news section of the participation module would be another; the general news section from the site would be another; etc.)
- we'll be aggregating pieces from different collections. Examples: 
  - In the general news section of the participation module we would be showing news from that collection, as well as from the collections from the different groups the module contains. 
  - In a theme page (where we are listing all groups/processes belonging to a group), we would aggregate news from the collections from the different groups the theme contains. Same with agendas and updates.

I think we should only go this way if we truly manage to keep things simple, if not this will be a nightmare soon ;) So lets put an effort into making the simplest scheme possible.

Once we agree on the model, we should think on the interface for admin, ie. where are we providing the UI to add news; if we have a central screen to manage news, or if we provide a news section inside each module/group/etc.

![captura de pantalla 2017-06-23 a las 15 02 11](https://user-images.githubusercontent.com/24740/27483338-404ee32c-5825-11e7-8ad1-a60b5b76769d.png)

Providing a central location will make development simpler, at a cost in the UX; and viceversa. So lets think of a simple way of providing a news section admin UI for each collection in its context. 

This scheme could also share characteristics with what we may do with agendas #753 

cc @ferblape

furilo:
In branch 752-participation-admin-html / Test HTML: https://github.com/PopulateTools/gobierto/pull/785/commits/2b6c64e7efee92a36d53a7c09ab4462a6afff326 

What: 

* [CMS home](http://madrid.gobierto.dev/sandbox/admin_cms): List last edits + collections
* [Collection](http://madrid.gobierto.dev/sandbox/admin_cms): 
  * List pages from that collection (same list as in the CMS now)
  * Collection configuration/new: 
    * Context: where the collection is located (Site would be for general pages, for example)
[...]

PR #799, opened 2017-07-13
[WIP] CMS Collection

apradillap:
Connects to #754


ferblape (approved):
👍 looks good to me!

amiedes (approved):

ferblape (changes requested):
Just a couple of small issues:

screen shot 2017-07-21 at 16 46 23


1 - could you translate the attributes of the collection? (title, slug, container, item type). We use to do this in `config/locales/<module name>/models/<locale>.yml`, for example in GobiertoBudgetsConsultations

2 - why in the container dropdown hasn't any option? Maybe the sites are not yet included in the available containers?

apradillap:
@ferblape I translated ymls relation with collections and issues that I forgot in the previus PR. 
In addition, when selecting a container you can already select sites or issues.

I deployed in staging after several circle failures 😓

ferblape (approved):

PR #805, opened 2017-07-19
Use 2017 data

ferblape:
### What does this PR do?

This PR activates 2017 data


ferblape:
We need to perform a few checks in staging before releasing it.

Issue #753, opened 2017-06-22
Global Agendas prototype


furilo:
We have to think if we are going to adapt current agendas to make them more flexible or create a separate entity, and how it affects the backend.

furilo:
I've added the prototype for the global agendas section in the admin https://github.com/PopulateTools/gobierto/pull/785/commits/cba2928cd61d45546489e203213c498d7f6898ce

**Sidebar**: I've reordered it a little bit: moving CMS to the second block, thus leaving CMS, Agendas, Docs: 

![captura de pantalla 2017-07-07 a las 11 59 58](https://user-images.githubusercontent.com/24740/27953388-e4ba32b8-630b-11e7-8658-f66cac5068f6.png)


[**Agendas** home](http://madrid.gobierto.dev/sandbox/admin_agendas) in the admin: I've followed the same scheme as in CMS #754: last activity around events, and access to agendas. We'll be able to access standalone agendas, or agendas belonging to a module (the Agenda of a person, or the Agenda of a participation group). 

In this last case, we have to decide if we provide the navigation context of agendas or the destination module (in fact, we should also apply the same decision for CMS). I think I balance myself towards providing the destination context, with a link to go back to global agendas if you come from there. Thoughts? This would look like: 

http://madrid.gobierto.dev/sandbox/admin_people_person_agenda?from_global_agendas=true

![captura de pantalla 2017-07-07 a las 12 34 22](https://user-images.githubusercontent.com/24740/27954520-a38032ac-6310-11e7-97c6-c8e6e3147450.png)

(we can iterate on the actual design of that back button)

ferblape:
Given an agenda, or a CMS collection we can know the nearest parent. For example, if an agenda belongs to a person, the nearest parent is the person, and the farthest will be the module the site.

Based on that nearest parent we can build a breadcrumb with the navigation, so 👍 to that double navigation.

furilo:
Mmm... but we probably don't want to offer the double navigation always, only when you are coming from global agendas.

PR #832, opened 2017-07-31
Extract agendas and events to GobiertoAgendas module, integrated with new Collections

ferblape:
Connects to #753

### What does this PR do?

This PR refactors GobiertoPeople::PersonEvent class (and related classes) to a more generic classes, implementing the new GobiertoAgendas module.  This module uses the new GobiertoCommon::Collection abstraction.

It also implements the admin UI to create and manage events in collections.


app/models/gobierto_people/person.rb
```diff
@@ -58,5 +60,28 @@ def attributes_for_slug
       [name]
     end
 
+    def events_collection
+      @events_collection ||= GobiertoCommon::Collection.find_by container: self, item_type: 'GobiertoCalendars::Event'
+    end
+
+    def events
```
amiedes, line 67: @ferblape Random comment: ¿Is it a good idea to memoize this? I haven't checked this, but I'm thinking of a situation when after doing `person.reload` the `@events` variable still holds the previous value because it's not an attribute [see the docs](https://apidock.com/rails/ActiveRecord/Base/reload) about the reload method.

ferblape: Good point, I think you are right!

Issue #815, opened 2017-07-25
Implement Group events and event pages


ferblape:
Sandbox agenda (collection): http://madrid.gobierto.dev/sandbox/participation_agenda.html
Sandbox event: http://madrid.gobierto.dev/sandbox/participation_event.html

The idea is that an agenda can belong to a process and a theme, or event to the whole module. So there will be different ways to access agendas. For the moment lest just focus on the agenda of a group.

PR #856, opened 2017-08-04
815 - Group events

amiedes:
Connects to #815 

### What does this PR do?

Implements super-dummy version or process events #index and #show


ferblape (approved):
👍

Issue #900, opened 2017-08-29
You can't create the same Issue in two different sites


furilo:
Right now if you have an issue Sport, which a url "sport" in one site, you can't use that url for another issue in another site. You should be able to do that, since sites are totally independent. 

cc @apradillap

apradillap:
Correct!! The validation is erroneous

```
validates :slug, uniqueness: :site_id
change by
validates :slug, uniqueness: { scope: :site }
```

I add it to the Sprint

PR #901, opened 2017-08-29
create the same issue in two different sites

apradillap:
Connects to #900


ferblape:
I don't need a git blame to know that I have written those validations 🥇

Issue #914, opened 2017-09-01
Improvements for documents management inside a process


furilo:
- [x] Can we autoextract the title from the filename and have that as the title of the document if the user doesn't provide a title?
- [x] In the tab listing documents, add a column "Date" (created)
- [x] In the tab listing documents, add a column with the filename
- [x] In the listing docs tab, the spanish literal says "Nueva" for new document, it should be "Nuevo"
- [x] In the edit document screen, put the save button in the right column, the same as in other editing pages

cc @apradillap

PR #939, opened 2017-09-11
Improvements document management

apradillap:
Connects to #914 

- [x] Save a document without name in a collection.
- [x] Go to the documents view of a collection and you'll see the literal "New" in Spanish and the filename column. I have deleted the description column because I don't think it will add much to this view, since we still have to insert the public link of the document.
- [x] Edit a document


ferblape (approved):
:+1: documents now looks better.

Issue #935, opened 2017-09-08
Participation polls front


amiedes:
Tasks:

- [x] Backend code
- [x] Tests
- [x] AJAX call to submit poll - Callbacks for success/error
- [x] When clicking on (X), redirecto to polls#index
- [x] Show share in social networks slide after successfully submitting poll
- [x] Create new layout without header and footer
- [x] Check user is logged in before letting him answer a poll
- [x] If user is not logged in, show login modal
- [x] Link poll fron admin ("View poll" link)
- [x] Disable next question button until user answers question
- [x] Validations at form object level to check all questions were answered
- [x] Let answers to open questions be blank
- [x] In polls index, deactivate link if already participated in poll

Pending:

- [ ] CSS for disabled next question button
- [x] View poll links in admin/process/process_id/polls table

For the future:

- [ ] Save poll state for when an error occurs
- [ ] Check user has right permissions to answer a poll (census integration pending)
- [ ] Button for going back between the questions

Pending edge cases:

- [x] Polls with 1 question are buggy (refactor JS also)
- [x] Check links from admin (preview tokens etc.)

Issue #833, opened 2017-08-01
HTML for polls front and back


furilo:
## Admin

Polls list for a group: http://madrid.gobierto.dev/sandbox/admin_part_group_polls
Poll edit: http://madrid.gobierto.dev/sandbox/admin_part_group_polls_poll_edit

## Home 

http://madrid.gobierto.dev/sandbox/participation_group_polls

## Poll 

Launch from http://madrid.gobierto.dev/sandbox/participation_group_polls

Elements to integrate:

- Poll title
- Question title
- Question type:
  - I have added a control to select answer type (single or multiple) in <h4>Tipo de pregunta</h4>
  - If it's a single choice answer, if you select any option, the rest should be unchecked
  - If it's a multiple choice answer, you can check various options. Also, the div.help should be shown
- Question progress: 
  - I've skecthed a very basic JS to provide animation and button interaction, it should be properly integrated
- Progress map
  - move .active through li's as question advances
- Last screen: 
  - remove .poll_progress_map and .poll_actions

amiedes:
## Order of tasks:

* [x] Sketch models and DB tables
* [x] Tests for admin polls index
* [x] Implement admin polls index
* [x] Write tests for `Poll`
* [x] Implement `Poll` create / update
* [x] Write tests for `PollQuestion`
* [ ] Implement `PollQuestion`
* [ ]  Write tests for `PollAnswerTemplate`
* [ ] Implement `PollAnswerTemplate`
* [x] Write tests for frontend
* [ ]  Implement frontend

PR #943, opened 2017-09-12
Admin polls

amiedes:
Connects to #833 

### What does this PR do?

Creates the DB and models infrastructure por the poll system, as well as the admin GUI for creating/updating them.

**Poll creation**

1. Go to [http://madrid.gobierto.dev/admin/participation/processes](http://madrid.gobierto.dev/admin/participation/processes)
2. Edit a process that has polls stage active (or activate it for a existing one)
3. Inside polls tab, click "New"
4. Fill title, description and start & end date
5. **Add new question (A)**
  a. Fill title
  b. Select "open answer"
6. **Add new question (B)**
  a. Fill title
  b. Select "fixed answer"
      1. Add new answer (A)
      2. Add new answer (B)
      3. Add new answer (C)
      4. Delete answer A (and check that after submit it wasn't created)
      5. Reorder answer C to be before B (and check after submit it's on the first place)
7. **Add new question (C)**
  a. Fill title
  b. Select "open answer"
8. Delete question A (and check later it was never created)
9. Reorder question C to be before B (and check after update it's on the right position)

No

### Pending work

The frontend for users answering to polls is still pending, work will continue in issue #935. Also some minor details of the admin GUI are pending, since I need #935 to be finished to do it (like the link to "view poll")


ferblape (changes requested):
@amiedes great work! works very well and code looks good, despite of the complexity.

I only have 3 minor change requests:

1. When editing a new question, it'd be great that the title has the focus when the popup is open. You can do this with Javascript.

screen shot 2017-09-14 at 16 17 46

2. Same with an answer

screen shot 2017-09-14 at 16 17 51

3. It'd be good that when you create a new question and you click Cancel, don't create an empty row. Do you think it would be easy?

screen shot 2017-09-14 at 16 18 01

And a bug: when you edit a poll, there's an issue with simple calendar and turbolinks. I think we solved it in other parts of the app by disabling Turbolinks. Maybe we should do the same for this page.

amiedes:
❤️

Issue #841, opened 2017-08-02
Integrate the new login/signup modal window


ferblape:
Sandbox: http://madrid.gobierto.dev/sandbox/participation.html (click in any of the poll options)

The idea is to reuse the existing HTML forms, but in an Ajax way.

ferblape:
This is done, as @apradillap wrote:

Plain HTML link:

`link_to new_user_sessions_path, class: "pure-menu-link"`

Modal window link:

`link_to '', new_user_sessions_path(open_modal: true), class: "pure-menu-link open_remote_modal"`

PR #947, opened 2017-09-13
Modal login signup

apradillap:
Connects to #841 

### What does this PR do?

It offers the possibility to use the login with a modal window.

We will have some links where we will use as before:
`link_to new_user_sessions_path, class: "pure-menu-link"
`
But if we want open modal window:
`link_to '', new_user_sessions_path(open_modal: true), class: "pure-menu-link open_remote_modal"
`

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -79,6 +79,12 @@ $(document).on('turbolinks:load', function() {
 
   // Modal windows
   $('.open_modal').magnificPopup({
+    type: 'inline',
+    removalDelay: 300,
+    mainClass: 'mfp-fade'
+  });
+
+  $('.open_remote_modal').magnificPopup({
     type: 'ajax',
     removalDelay: 300,
     mainClass: 'mfp-fade'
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/application.js.erb
```diff
@@ -79,7 +79,7 @@ $(document).on('turbolinks:load', function() {
 
   // Modal windows
   $('.open_modal').magnificPopup({
-    type: 'inline',
+    type: 'ajax',
```
ferblape: Careful, changing the behaviour here breaks other parts of the code which rely on this magnificPopup, for example, editing a stage of a process:

screen shot 2017-09-14 at 15 52 50

I'd suggest to have to handlers:

- `$('.open_modal')` opens an inline modal
- `$('.open_remote_modal)` opens a remote modal, using AJAX

ferblape (changes requested):

app/assets/javascripts/application.js.erb
apradillap: Sorry 😓  , I didn't see that part. I saw that it was used in this file (app/Views/gobierto_budgets/budgets/index_old.html.erb)  but I thought it was deprecated. I added open_remote_modal to the application javascript because it was only for admin application javascript.
`
link_to new_user_sessions_path(open_modal: true), class: "pure-menu-link open_remote_modal"
`
Thanks!!

ferblape (approved):
🚀

Issue #971, opened 2017-09-25
Implement progress map


furilo:
https://github.com/PopulateTools/issues/issues/70

ferblape:
The code for this issue is in #972 

@apradillap feel free to prioritize it in the sprint.

PR #972, opened 2017-09-25
971 process progress map

furilo:
Connects to #971, https://github.com/PopulateTools/issues/issues/70

### What does this PR do?

Partial markup for the Process Progress Map. 

It lacks the dots for each phase and the tooltip on hover seen in https://github.com/PopulateTools/issues/issues/70

When clicking on a dot, it should scrollTo (use Velocity function to have a smooth scroll) to id="timeline"


amiedes:
Implementation is ready in staging. The tooltip still needs to be styled as in [here](https://user-images.githubusercontent.com/4933/30172843-343eb64a-93f6-11e7-97ff-0d38511ac076.png).

Also there is something messed up with the styles of the buttons in staging (in local works properly):

captura de pantalla 2017-10-13 a las 15 28 34

ferblape (approved):
👍

Issue #918, opened 2017-09-04
Polish UX of process/group creation


amiedes:
In #804 I left a few things unfinished, let's do the following improvements: 

- [x] Add Save/Cancel buttons in the process modal
- [x] Missing translations (stage translation in flash message)
- [x] If no slug is provided, it will be autofilled in the server
- [x] Both processes and groups can have stages
- [x] Stages are not compulsory, nor for processes nor for groups.
- [x] The first time a process/group is created, the stages form will not appear. After submitting the form a flash will appear ("Process/group created, now you can add stages") and the stages form will be visible.
- [x] The "activate stage" checkbox its just a flag, it does **not** delete information in the server. When a stage is deactivated, the information is kept in the DB just in case it gets activated again in the future.
- [x] Flag to ignore/use datepicker values
- [x] Make sure stage title is compulsory but description is optional.
- [x] Text input placeholders are missing.
- [x] Add link to the front process (as we do in other sections of Gobierto)
- [x] Fix the following flow:
  1. Create process with 1 stage
  2. Add new stages, but dont fill in stages info, click save
  3. Checkboxes are emptied, and "success" message appears, but stages are not saved.
- [x] Create data migration to create all missing stages for existing processes and groups

ferblape:
Could you add also a link to view the process in the flash message as we do in other sections of Gobierto?

![screen shot 2017-09-20 at 08 40 01](https://user-images.githubusercontent.com/17616/30630100-67a02184-9ddf-11e7-9ebd-9be8f0f701bb.png)

PR #973, opened 2017-09-26
918 - Polish process creation UX

amiedes:
Connects to #918 

### What does this PR do?

Implements a series of changes so creating a new process/group is less restrictive and more user friendly (see details below).

**IMPORTANT:** from now on, when a process is created all of its stages records are created, but with the flat `active: true`. This PR contains a data migration to update all the previously created process stages.

When creating a process from the admin, the following things should have changed:

* When you first create a process / group, the part of the form dedicated to create stages is hidden.
* After creating the process, you should be able to add new stages.
* As long as a stage is not activated, validations shouldn't avoid it from being saved.
* Starts / ends is no longer compulsory for processes. There is a checkbox to toggle this.

Nope


ferblape (approved):
👍 Very nice!

PR #68, opened 2016-11-21
WIP - don't merge yet - Budget consultations html

furilo:
### What does this PR do?

The HTML for the front part of the consultations module:

- http://madrid.gobierto.dev/sandbox/consultation
  - all the "Cómo funciona" and beyond would be what the admin inputs in the admin UI)
  - while we have only one consultation, we would redirect the user to this page. in the future we'll create a consultations home, with active, past, etc.
- http://madrid.gobierto.dev/sandbox/consultation_budget-line
- http://madrid.gobierto.dev/sandbox/consultation_confirmation
- http://madrid.gobierto.dev/sandbox/consultation_thanks

Also, user login/signup and activation page: 

- http://madrid.gobierto.dev/sandbox/user_signin
- http://madrid.gobierto.dev/sandbox/user_confirmation

Pending: 

- [ ] User verification screen
- [ ] User account page

### How should this be manually tested?

PR #976, opened 2017-09-28
68 - Fullcalendar view for GobiertoPeople person events index

amiedes:
Connects to [#68](https://github.com/PopulateTools/issues/issues/68)

### What does this PR do?

* Implements [Fullcalendar](http://fullcalendar.io) for gobierto people person agenda.
* Previous event listing is kept under the button *Lista*.
* *Past Events* filter under the *List* was causing some issues because past events weren't being retrieved using AJAX and page reloading was causing some Turbolinks issues.

  Turbolinks had to be disabled anyways afterwars, but since now we have the Fullcalendar view and the *Past Events* view does not make sense anymore, this filter has been hidden using a `<% pending do %>` block.  

image

No


ferblape (approved):
👍 looks good to me!

ferblape:
@furilo please review and confirm if it's fine to merge in master. You can test it here: http://santboi.gobify.net/agendas/francisco-gutierrez?locale=ca

Maybe I'd adjust the colors of the buttons when hovered:

screen shot 2017-10-01 at 08 48 11

furilo:
I'm seeing one issue: If you go to the list view, and the go back to any of the other views, the events are not loading ok. If you browse again to any calendar views, it gets fixed. 

screencast 2017-10-02 09-55-32

ferblape:
After investing a couple of hours trying to debug the error and trying different solutions to refresh the events heights once the container is shown again, it looks like there's no easy way to fix this. 

The solution we have come to is to use a min-height for events. Given that it's not hapening always, it's a good compromise.

PR #990, opened 2017-10-04
988 - Fix initialization of pages with a single datepicker

### LINKED ISSUE #988: Can't select date for a goods declaration
In Person > Declarations > New, when trying to select a date for the declaration, I click on the the date selector but it doesn't do anything. No JS error fires.

And if you try to save the declaration, you can't because it doesn't have a date.

amiedes:
Fixes #988


amiedes:
@ferblape The datepicker initialization code inside `application.js` really needs to be rewritten in a more clear way, but for the time being I'd leave it like this.

test/services/gobierto_budget_consultations/csv_exporter_test.rb
```diff
@@ -16,8 +16,8 @@ def test_export
     csv = @subject.export(consultation)
     assert_equal csv, <<~CSV
       id,age,gender,location,question,answer
-      692345489,35,male,,Pavimentación de vías públicas,-5
-      692345489,35,male,,Inversión en Instalaciones Deportivas,5
+      692345489,36,male,,Pavimentación de vías públicas,-5
```
ferblape: We prefer to fix it in a more resilient way.

Issue #989, opened 2017-10-04
Allow an admin user to edit only specific People


furilo:
- [x] Markup
- [x] Implementation (define more tasks if needed)
  - [x] Don't show the module list for Manager users. (Right now, the Permissions for modules list is show for both Manager and Regular users. Manager users have access to every module, so that option don't make sense)
  - [x] If you change the user type in the right side from Regular to Manager or viceversa, hide/show the Permissions/Sites list in the main column
  - [x] When a Regular user is selected, if Gobierto People is selected, show the list of users ([HTML](https://github.com/PopulateTools/gobierto/pull/991/files#diff-50c43ef47884fdc641e60607aec70659R66))
  - [x] By default, the first option "All" should be selected. If you click on any individual user, that option should get de-selected in client side.
  - [x] If one or various were selected, but the got de-selected, when the count is 0, the option "All" should get selected again

amiedes:
Tasks:

- [x] JS frontend
- [x] Backoffice implementation
- [x] Integrate the *All* button
- [x] Remove debug comments and annotations
- [x] Disable links in people index when no permissions exist. Ensure this is not permitted at controller level either (I'll use the `Policy` here)
- [x] PersonPolicy tests
- [x] Include the `AuthorizableResourceTestModule` in all `gobierto_admin/gobierto_people/tests`

amiedes:
@furilo I wouldn't do this one:

![captura de pantalla 2017-10-10 a las 13 36 35](https://user-images.githubusercontent.com/9287468/31384651-1e15a994-adc0-11e7-9fa2-cc8d1b2c0f92.png)

It doesn't add too much and maybe we end up having some *"intermediate" admin states* where no people permissions have been granted yet.

furilo:
Mmm... OK.

PR #991, opened 2017-10-04
Enable that an admin user only can edit specific people

furilo:
Connects to #989 

### What does this PR do?

- [x] Provide the markup 
- [ ] Implements

- [ ] Markup available in http://madrid.gobierto.dev/sandbox/admin_users_user


furilo:
A last request: 

- [ ] If a user can't edit all users, she shouldn't be able to reorder them. Just don't show the control to reorder.

PR #997, opened 2017-10-06
New task to check existing indexes and create the new ones

ferblape:
Connects to [#94](https://github.com/PopulateTools/issues/issues/94)

### What does this PR do?

This PR adds a new task to be executed during the deploy. This task checks the searchable models in Algolia and checks if the index exists. If it doesn't, the script will create it.


ferblape:
cc @amiedes @apradillap

Issue #1016, opened 2017-10-16
Edit page: Add sections UI


furilo:
- When editing a page, an admin can select a section for the page from a dropdown
- When a section has been selected, a dropdown appear with its first level to the page can be set as a son of one of this, or as a first level page
- If the user selects a parent, another dropdown will appear with its sons, repeating previous point
---
- [x] Markup #1017 
- [ ] Implementation

PR #1017, opened 2017-10-16
UI for including a page in a section from the edit page screen

furilo:
Connects to #1016 

### What does this PR do?

- [X] Provides the markup for including a page in a existing section
- [x] Implements it

- Add section from page creation
- Update section from page update
- Creation of new section
- Add section without parent
- Add section of a child
- Delete section

After adding the page to the section you can check the tree of the section.

Consideration: When a page belongs to a section several times, only the first section item is shown in the page edition.

### Pending (VUE)

- [x] When you create a new section, the multi-language titles displayed in the window appear. I thought it was relation with turbolinks but I deactivated them and it continue. Also, when you create a section from /admin/cms/pages, the multi-language section options don't appear.


apradillap:
This is ready to be reviewed. 😄

furilo:
- Should we add an event log that a page has been put into a section? 
- [ ] Right now it's not obvious to add a page from the CMS page. Let's add a new column in Collections, with the fa.plus-circle, which would take you to Collection>New page
- [ ] It's not from this PR, but let's do it: in http://getafe.gobify.net/admin/cms/sections/5 - the order of "last edited pages" is wrong

app/assets/javascripts/admin/app/controllers/gobierto_cms_controller.js
```diff
@@ -73,6 +73,11 @@ this.GobiertoAdmin.GobiertoCmsController = (function() {
                 'globalized-form-container': 'true'
               }
             }
+          },
+          callbacks: {
+            ajaxContentAdded: function() {
+              window.GobiertoAdmin.globalized_forms_component.handleGlobalizedForm();
```
apradillap: Viewed 🕵️

ferblape:
I've fixed the state of the checkbox when the user clicks and the translated attributes in the form:

screen shot 2017-11-16 at 16 59 19

Just a couple of things, if you need help with the JS ping me:

- [ ] When the user unchecks the checkbox, the page should be removed from the section
- [ ] When the form is loaded and the page belongs to a section the checkbox should be checked
- [ ] If I click multiple times the checkbox the list witht he elements of the section grows and grows
- [ ] tests!! With the javascript drivel shouldn't be difficult to test this

apradillap:
@ferblape 

- [x] add a new column in Collections, with the fa.plus-circle, which would take you to Collection>New page. **Extra:**  I have also added this option to the collections of documents and events to homogenize.
- [x] When the user unchecks the checkbox, the page should be removed from the section
- [x] When the form is loaded and the page belongs to a section the checkbox should be checked
- [x] If I click multiple times the checkbox the list witht he elements of the section grows and grows
- [x] tests!! With the javascript drivel shouldn't be difficult to test this. **Extra**, I also added test for the creation of section.
- [x] It's not from this PR, but let's do it: in http://getafe.gobify.net/admin/cms/sections/5 - the order of "last edited pages" is wrong.
- [x] **Extra:** I detected and fixed that the collections of pages and news didn't display correctly the items in the collection.

Álvaro asked Should we add an event log that a page has been put into a section? We save these events. They are not saved when updating the position of a page in a section, I saw it excessive.

ferblape:
👍 released!

Issue #1051, opened 2017-10-30
Fix calendar integration picker on Firefox


amiedes:
There's an issue with Firefox that provokes the calendar integration picker to only display the current integration, instead of all of them:

![captura de pantalla 2017-10-30 a las 18 07 03](https://user-images.githubusercontent.com/9287468/32184641-2cd1e24e-bd9d-11e7-980f-8dfb6d05241f.png)

Issue #1046, opened 2017-10-29
500 when browsing to / for a site with only Participation activated


furilo:
If you browse to site.domain.com/participacion it's ok, but if you go to

site.domain.com it gives 500

apradillap:
@furilo , I've been seeing that it doesn't give an error because the Participation Module is activated. One way to fix the error quickly is to remove the modules from the site and first select one of the modules that do have root_path, for example: Gobierto Budgets, Gobierto Indicators, Gobierto People or Gobierto Participation. To solve it correctly, **root_path** should be defined for the modules that currently do not have:
- [ ] Gobierto Development
- [ ] Gobierto Budget Consultations
- [ ] Gobierto CMS

The other modules have:
- Gobierto Budgets: /presupuestos/resumen/2017
- Gobierto Indicators: /indicadores
- Gobierto People: /cargos-y-agendas
- Gobierto Participation: /participacion

furilo:
Let's neutralize this problem with #927

Issue #927, opened 2017-09-07
Define the default site homepage from the site settings


furilo:
- An admin will be able to choose which module act as the global site homepage
- An admin will be able to choose the CMS Module, which will show another selector showing Sections. The initial page from the choosen Section will act as the global site homepage is choosen

PR #1062, opened 2017-11-02
Select homepage

apradillap:
Connects to #1046 

### What does this PR do?

An admin will be able to choose which module act as the global site homepage. The first task has been done because the functionality is in production and the section part for the front is not yet implemented (https://github.com/PopulateTools/gobierto/issues/927)

Editing a site. I have followed the same casuistry as when selecting a default locale where if the selected locale is not among the allowed ones, it also adds it to the allowed, so, if a module is selected as homepage and it is not among the selected modules, it is also selected.

### Execute
This task selects in the order in which the homepages were selected to define the homepage.
`
rake data:migrate
`

Fernando, in order to have controlled the modules that can be selected as root_paht I put it in the application.yml, next to where modules are added but we have to modify the application.yml in the servers with these values.

In application.yml:

```
  site_modules_with_root_path:
    -
      name: Gobierto Budgets
      namespace: GobiertoBudgets
    -
      name: Gobierto People
      namespace: GobiertoPeople
    -
      name: Gobierto Participation
      namespace: GobiertoParticipation
    -
      name: Gobierto Indicators
      namespace: GobiertoIndicators
```


apradillap:
PendingTasks:
- [x]  Add root_path to pending modules
- [x]  Add documentation

furilo:
I think we should control a little bit better the interactions between start page option and activated modules. Suggestions: 

- [x] Move the Start page select after the Activated modules
- [x] Control with JS that if I mark a module as activated or not, the corresponding option appears/disappears in the combo

apradillap:
Pending to put in ansible the configuration of the application.yml, if not, in each deployment in staging the possibility of proving it is lost.

furilo:
Good for me. I've included styles and html to fix #1051 (not 1050 as I referenced on the commit message)

apradillap:
:wrench: ansible updated, executed and reviewed in production and staging.

ferblape (approved):
👍

Issue #1057, opened 2017-10-31
Enable to publish news for participation module


furilo:
- In Collection context a context "Module: Participation" should be available
- In Collection type: a type "News" should be available
- When a Process is created, the collection type should be News
- Update to type News all the existing processess CMS collections

Related #1056

PR #1066, opened 2017-11-02
Enable publish news

apradillap:
Connects to #1057 

### What does this PR do?

- In Collection context a context "Module: Participation" should be available
- In Collection type: a type "News" should be available
- When a Process is created, the collection type should be News
- Update to type News all the existing processess CMS collections

### Task

`
rake data:migrate
`


apradillap:
captura de pantalla 2017-11-02 a las 17 48 54
@furilo News out of the module

ferblape:
We might need this script too:

```ruby
Site.all.each do |s|
  s.collections.where(item_type: "GobiertoCms::News").each do |c|
    c.collection_items.where(item_type: 'GobiertoCms::Page').each do |i|
      i.item_type = 'GobiertoCms::News'
      i.save!
    end
  end
end
```

Issue #1047, opened 2017-10-29
Take into account Issues order when listing them in the front


furilo:
I have reordered issues in /admin/issues, but the order is not reflected in the front (ie. participation home)

Review all places where we are listing issues in the front, and take into account its order (anyhow, this should have just to be done in one place, right?)

**Update:** But it seems it doesn't. 

In /participacion/agendas you get the wrong order, in /participacion/noticias you get the right one.

PR #1072, opened 2017-11-03
Define order for current_site.issues has_many with the scope sorted

apradillap:
Connects to #1047

### What does this PR do?

Correct that it always shows ordered issues in the site. The solution has been simple, in the relationship of issues with site is added the scope of ordering, so it is only in one place of the site

In the participation module, in participation home, in the list of issues, in the filters for news or events, etc.


furilo:
Good for me!

PR #1077, opened 2017-11-06
Alcobendas districts map

Crashillo:
### What does this PR do?
Add a SVG districts map


furilo:
I'm closing this because we are putting the map code directly in the template (web admin ui) of the corrsponding site.

Issue #1084, opened 2017-11-07
Admin > When viewing a process, remove "Processes" from the breadcrumb


furilo:
In the Processes part of the breadcrumb we are linking to /admin/participation/processes, which is a leftover (and wrong). I think the only place from where it is called is the breadcrumb when you are seeing a process or an element from a process. We can get rid of the "Processes" part of the breadcrumb.

PR #1086, opened 2017-11-08
Remove Processes form the admin breadcrumb

apradillap:
Connects to #1084 

### What does this PR do?

- We get rid of the "Processes" part of the breadcrumb.
- Extra: Correct breadcrumb for process information

Editing a process from admin


ferblape:
👎  I still can see the link in the admin:

http://participacion.gobify.net/admin/participation/processes/7/edit

screen shot 2017-11-09 at 08 22 17

ferblape:
Also, if that action is not linked and won't be used we should remove it.

Issue #1060, opened 2017-11-01
Collection type and container should be unique per site


ferblape:
There can't be collections of the same type and container in a site, because will break things. 

1. Add a unique index
2. Add a validation

PR #1126, opened 2017-11-16
Collection unique per site

apradillap:
Connects to #1060 

### What does this PR do?

- [x] Add a unique index
- [x] Add a validation

I had to delete a test relation with the person collection creation because we have a lot of dependences with some test and the other paths required more change.

Try to create a collection with a container that already has another collection.

### Clean collections

I had to clean up duplicate staging collections from console:

Site 4
16: 4
21: 0 -> Delete

Site 7
17: 1 Wadus -> Delete
43: 32

Site 10
86: 3
106: 1 -> Fake example
107: 1 -> Fake example

Site 12
88: 0  -> Delete
96: 0 -> Delete
108: 1

Process 2
49: 4
59: 0 -> Delete

Process 9
68: 0 -> Delete
122: 0 -> Delete
123: 0

### DIFF
--- a/test/integration/gobierto_admin/gobierto_calendars/collections_test.rb
+++ b/test/integration/gobierto_admin/gobierto_calendars/collections_test.rb
@@ -70,6 +70,42 @@ module GobiertoAdmin
           end
         end
       end
+
+      def test_person_events_collection
+        with_javascript do
+          with_signed_in_admin(admin) do
+            with_current_site(site) do
+              visit @path
+              click_link "Richard calendar"
+
+              assert has_selector?("h1", text: person.name)
+              assert has_link?("< Back to agendas")
+
+              within ".sub_filter ul", match: :first do
+                assert has_selector?(
+                  ".all-events-filter",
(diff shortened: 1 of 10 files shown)


test/integration/gobierto_admin/gobierto_calendars/collections_test.rb
```diff
@@ -70,62 +70,6 @@ def test_create_collection
           end
         end
       end
-
-      def test_create_collection_for_person
```
ferblape: Hey, what was wrong with this test?

ferblape (approved):
LGTM, but I restored the test that you removed. Calendar collection has a special layout in the admin, and that test was checking that.

test/integration/gobierto_admin/gobierto_calendars/collections_test.rb
apradillap: Excuse me. I'm sorry, it's true that it doesn't fail, it'd fail me for something else.

Issue #1085, opened 2017-11-08
Attachments on news/pages of a process


ferblape:
Should be included in the Process attachments collection automatically.

Related to https://github.com/PopulateTools/issues/issues/149

PR #1152, opened 2017-11-20
Attachments on news/pages of process

apradillap:
Connects to #1085

### What does this PR do?

When you add an attachment to a new (**Extra:** also to an event) it will appear in the document section of the process.

You can attach a document to a page and it will appear in the documents part of a process. It will also appear at the Issue and Scope level

### Comments

I had to correct part of the append method of Collection because:

- When a news item was related to an Issue, it wasn't saved correctly item_type
- The association of collection_items with scopes wasn't being saved
- I've corrected fixtures
- I deleted part of the append method (collection) because it was not implemented correctly.


apradillap:
I've been passing the tests on the site with id 12. I have a question, should documents associated with pages appear in the documents section of the collection to which the page is associated? I would say no because these attachments associated with page are not associated with a collection and to show the documents of the collections we should now change the way we extract them through collection items.

furilo:
> I would say no because these attachments associated with page are not
associated with a collection

Should they be?

ferblape (changes requested):
Code looks good. Just one thing: event attachments are not displayed in documents page:

[This event](http://participacion.gobify.net/admin/gobierto_calendars/events/7736/edit?collection_id=66) has an attachment which I can't see here: http://participacion.gobify.net/participacion/documentos

apradillap:
@furilo I commented with Fernando my doubts about the attachments and were resolved. Remember (https://github.com/PopulateTools/issues/issues/149) that since the event attachments are going to change from a string field to Attachment object, it will be necessary to add a filter when an event is updated, add the attachments it has to the collection as Fernando commented in https://github.com/PopulateTools/gobierto/pull/1152

Issue #974, opened 2017-09-27
Amend Sluggable module


ferblape:
[Sluggable module](https://github.com/PopulateTools/gobierto/blob/master/app/models/concerns/gobierto_common/sluggable.rb) works fine, but it can be improved a bit:

1. The callback that sets the slug should be `before_validation`. This way, it will work not only after creating but after updating

2. The module should include two validations:
  - validate the presence of the slug
  - validate the uniqueness of the slug in the scope of the site

amiedes:
I have a concern about this concern :trollface:.

If I remember right, the reason why I used `before_create` was because the slug shouldn't be `null` (there is a DB constraint for this). This way, when a user doesn't specify a slug in the interface, it gets auto-assigned when the record is created for the first time.

So, a `before_validation` is not strictly needed, but adding it won't change the current behavior.

ferblape:
After discussing with @amiedes I have updated the issue.

apradillap:
Keep in mind when we do the issue that as it has an after_create, in the case, objects like PersonPost the method attributes_for_slug

```
    def attributes_for_slug
      [created_at. strftime ("%F"), title]
    end
```

It has an attribute that we haven't in the before_validation and we will have to omit the creation_at for the slug no? cc @ferblape

ferblape:
Good point, let's us Time.now.

PR #1154, opened 2017-11-20
Fix slug indexes and sync 2 days back for agendas

amiedes:
Connects to [#156 - Sync 2 days back](https://github.com/PopulateTools/issues/issues/156) and [#974 - Amend sluggable module](https://github.com/PopulateTools/gobierto/issues/974)

### What does this PR do?

* Syncs events since 2 days back for all agendas integrations
* Fixes many slug DB indexes so they are unique in the scope of the site (fixes Rollbar related to duplicated slug)

Example: create an event in a synced calendar 1 day ago, manually run synchronization and check the event gets added/updated.

You can manually run a calendar synchronization doing something like this:

```ruby
site = Site.find_by(domain: "domain.gobify.net")
person = site.people.first
site.calendar_integration.sync_person_events(person)
```

### Important notes related to migrations

This has been tested locally with both DB data from staging and production, so it is expected to succeed without problems, but we should keep an eye on this.

There were many attachment records in staging with a blank slug. The migration scripts assigns slugs for this records.

No

### DIFF
--- a/app/services/gobierto_people/google_calendar/calendar_integration.rb
+++ b/app/services/gobierto_people/google_calendar/calendar_integration.rb
@@ -51,7 +51,7 @@ module GobiertoPeople
       end
 
       def sync_calendar_events( calendar)
-        response = service.list_events(calendar.id, always_include_email: true, time_min: Time.now.iso8601)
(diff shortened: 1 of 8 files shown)


ferblape:
You haven't upload any code on the Sluggable module, right?

PR #1161, opened 2017-11-21
Sanitize slugs

apradillap:
Closes https://github.com/PopulateTools/issues/issues/150

### What does this PR do?

When you add a slug with spaces in the creation of any object, it creates the slug correctly.

Create an object that requires slug, for example, processes, pages

### DIFF
--- a/app/models/concerns/gobierto_common/sluggable.rb
+++ b/app/models/concerns/gobierto_common/sluggable.rb
@@ -11,10 +11,13 @@ module GobiertoCommon
     private
 
     def set_slug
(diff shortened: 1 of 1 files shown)


furilo:
It seems if you have the process already created, and change the slug, we are not sanitizing it in that moment.

apradillap:
Exactly @furilo, there is an issue to change the slug, not only in the edition (https://github.com/PopulateTools/gobierto/issues/974) so I commented that it was only in creation

furilo:
ok so let´s merge this

PR #1162, opened 2017-11-21
Wrong layout for a pool

apradillap:
Closes https://github.com/PopulateTools/issues/issues/152

### What does this PR do?

After a refactor in the controllers, pools layout wasn't correct.

http://alcobendas.gobify.net/participacion/p/Presupuestos%202017/aportaciones/en-que-nos-gastamos-450-000-en-el-distrito-norte

### DIFF
--- a/README_EN.md
+++ b/README_EN.md
@@ -23,8 +23,8 @@ Gobierto is being built in the open by [Populate](http://populate.tools), a prod
 
 More info:
 
-* Main site (spanish): [gobierto.es](http://gobierto.es)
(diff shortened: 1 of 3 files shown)


ferblape (changes requested):
Hey @apradillap there's another conditional in the footer of the layout that you didn't replace.

Also, could you extract a method in the application helper named `full_layout?` with this condition for the moment to avoid RY?

Also, in that file could you replace http://gobierto.es with http**s**://gobierto.es?

Thanks

ferblape (approved):

PR #1163, opened 2017-11-21
Force turbolinks editing a process

apradillap:
Closes https://github.com/PopulateTools/issues/issues/151

### What does this PR do?

In all calls to edit_admin_participation_process_path I've added 

`
data: { turbolinks: false } 
`

We have to remember that every time we send a link to a page with datepicker you have to send:

`
data: { turbolinks: false } 
`

Editing a process


ferblape (approved):
👍

PR #1173, opened 2017-11-22
No activity for contributors

apradillap:
Closes https://github.com/PopulateTools/issues/issues/153

### What does this PR do?

- Add public activity when you create a contribution.
- I corrected activity scopes

Create a contribution and you can see the activity as Participation or within a process


app/controllers/gobierto_participation/activities_controller.rb
```diff
@@ -23,7 +23,7 @@ def find_issue
     def find_participation_activities_issue(issue)
       ActivityCollectionDecorator.new(Activity.no_admin
                                               .in_site(current_site)
-                                              .in_container(issue)
+                                              .in_process(issue.processes)
```
ferblape, line 26: 👍

ferblape (approved):
👍

Issue #1150, opened 2017-11-20
Allow select a page as the home of a site


furilo:
Following #1062 we should enable that an admin can select a page or a section as the home of a site. For Getafe we need a page, but can do two things now: 

- Add a new option in the Start Page select: GobiertoCMS
- If GobiertoCMS is selected, show a second select, with the list of sections and then the list of pages
  - when we have tons of pages, we may implement something like this https://select2.org/dropdown to make it easier to select an specific page

Example of select: 

```
Section A
Section B
Section C
---
Pages (non-selectable option)
Page A
Page B
Page C
...
```

apradillap:
Seen, I think what you're proposing is a very good solution. 👍

PR #1174, opened 2017-11-22
Allow select page/section as homepage

apradillap:
Closes https://github.com/PopulateTools/gobierto/issues/1150

### What does this PR do?

Add https://select2.org/dropdown to select page or section as the home page

Edit stye

### Pending

- [x] Now the sections shown are from all sites, I have to add the scope
- [x] Display the new field name correctly

### Styles

- [x] Select does not display correctly when nothing is selected, it looks small


ferblape:
Should we avoid to unpublish a page which is used as the homepage?

apradillap:
It should be forbidden unpublish a page (homepage). I have doubts, how do we show it?

ferblape:
@furilo now the home renders the page or the section without redirecting.

ferblape (approved):
👍

ferblape:
Ok @apradillap let's show a flash message if the user tries to unpublish the page used as homepage.

ferblape (approved):
👍

Issue #1083, opened 2017-11-07
Small fixes in admin > pools > pool


furilo:
- [ ] Column State: currently we are showing a lock + "published". Let's remove this column until we develop the feature of moderating an item
- [ ] "Valoraciones": Let's show a column for each value (with title attribute so we show the full label): 
  - "++"
  - "+"
  - "0"
  - "-"
- [ ] Link "View item" correctly to the front (if we don't already have the permalink for an item inside a pool, open a new issue for that)
- [ ] Substitute the "New" button (it doesn't make sense in this context) with "View" (and a link target blank to the front)

PR #1178, opened 2017-11-23
1083 admin contributor containers small fixes

entantoencuanto:
Related with #1083

### What does this PR do?
- [x] Hides Column State
- [x] "Valoraciones": Let's show a column for each value (with title attribute so we show the full label): 
screen shot 2017-11-23 at 15 44 37

- [x] Links "View item" correctly to the front: The link is the same link to the pool with a hash and the slug of the contribution.
- [x] Substitutes the "New" button (it doesn't make sense in this context) with "View" (and a link target blank to the front)
screen shot 2017-11-23 at 15 44 48

Visit http://madrid.gobify.net/admin/participation/processes/27/contribution_containers/12

No


ferblape (approved):
👍

Issue #1168, opened 2017-11-22
Show polls in partials


furilo:
* An admin can add a helper method in a view (now in a view; in the short term in the template editor as well) to show a poll
* If the admin provides an ID, we select that specific poll
* If no ID is provided, we select between all the open and public polls
* If no ID is provided and the user is logged in, we select between all the open, public and not answered polls by the user
* The method would use the partial we currently have in the home page.
* If it's a multiquestion poll, the method would 
  * show the first question
  * the button would read "Continuar" instead of Send
  * it would take you to the poll page
* We'll have to do something equivalent for featuring pools, in case it make sense to think about that now
----

It could be something like: 

```
show_poll(poll_id: 123)
```

Being poll_id an optional parameter.

PR #1180, opened 2017-11-24
Template editor

furilo:
Starts with #1169 

### What does this PR do?

- [x] Basic markup in the sandbox for the template editor
- [x] Integration: for the moment we'll be able to edit gobierto_participation/welcome/index.html.erb, also using #1168 

Main points:

- Created gobierto_core folder where I have placed the Template and SiteTemplate models. In the future, we will move Issue, Activity, etc.
- Codemirror editor with https://codemirror.net/demo/folding.html plugins
- Translated app/views/gobierto_participation/welcome/index.html.erb to app/views/gobierto_participation/welcome/index.liquid
- New Liquid Tags: render_partial
- New Liquid Filters: liquid_i18n and image_filter
- Tree structure of template directories
- For new templates, replace html.erb with liquid and add it by console to the Templates model.

Browse to http://madrid.gobierto.dev/sandbox/admin_sites_site_templates

### Future

- Preview button


test/fixtures/gobierto_core/site_templates.yml
```diff
@@ -0,0 +1,4 @@
+madrid_participation_index:
```
ferblape, line 1: Could we add a test to check that this template is rendered instead of the regular view?

lib/liquid/gobierto_common/tags/render_partial.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class RenderPartial < Liquid::Tag
+  def initialize(tag_name, partial_name, options)
+    super
+    @partial_name = partial_name.gsub(/\s|"|'/, "")
+  end
+
+  def render(context)
+    context.registers[:controller].dup.render(partial: @partial_name.strip)
```
ferblape: Move the `.strip` to the initialize, so it'll be stripped always.

lib/liquid/gobierto_common/filters/image_filter.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
```
ferblape, line 1: Add this new tag to docs/liquid-templates

app/controllers/gobierto_admin/gobierto_core/site_templates_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+          :site_id,
+          :template_id
+        )
+      end
+
+      def ignored_site_templates_attributes
+        %w(created_at updated_at)
+      end
+
+      def find_site_template
```
ferblape, line 64: Filter by current_site

app/controllers/gobierto_admin/gobierto_core/templates_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        templates = ::GobiertoCore::Template.all
+        @dir = {}
+
+        templates.each do |paths|
+          merge_paths @dir, paths.template_path.split("/")
+        end
+      end
+
+      def form
```
ferblape: Are you sure this action is necessary? Right now does nothing if I'm not wrong.

ferblape (changes requested):
Also, could you add a data migration that creates the first Template?

apradillap:
@ferblape, I corrected all your changes requested

ferblape (changes requested):
Thanks for the changes, but I need to ask you one more thing: I have been able to write ERB tags and do queries to the database:

```
<% current_site = Site.last %>
<%= current_site.domain %>
```

We should strip ERB tags to avoid this security issue.

BTW, in the default template you are using ERB in the styles: `<%= image_url('illustrations/oldlady.svg') %>`

apradillap:
@ferblape ready!:
- Clean default participation welcome index template without erb tags
- Strip ERB tags in templates. [http://participacion.gobify.net/admin/gobierto_core/templates](http://participacion.gobify.net/admin/gobierto_core/templates). Note: as page level also makes a render_liquid, if you try to put ERB tags, it shows them as text, doesn't execute them.

ferblape (approved):
👍

PR #1184, opened 2017-11-27
Show polls in partials

### LINKED ISSUE #1168: Show polls in partials

apradillap:
Closes #1168 

### What does this PR do?

An admin can add a liquid tag show_poll in a template (/admin/gobierto_core/templates)

`
{% show_poll(poll_id: 353021530) %}
`
show_poll has these characteristics:
- If the admin provides an ID, we select that specific poll
- **If the admin provides an ID and it's answered, we select between all the open, public and not answered polls by the user (extra)**
- If no ID is provided, we select between all the open and public polls
- If no ID is provided and the user is logged in, we select between all the open, public and not answered polls by the user
- The method would use the partial we currently have in the Participation home page.
-Show only the first question, the options and next button
- If you are not logged in, after logging in you can continue to solve the poll.

- From /participation you can solve the poll.
- To test how it behaves according to the type of question, it is very easy to change the order of the survey questions.

### Styles

- [ ] The participate button when we hover the mouse over it, you don't see correctly. It's broken since we modified the follow_button in the header.
- [ ] Also when the question is multiple-choice, the options are not displayed correctly.

captura de pantalla 2017-12-12 a las 17 44 24


app/helpers/application_helper.rb
```diff
@@ -62,4 +62,32 @@ def filetype_icon(attachment)
     html = "<i class='fa fa-file" + fontawesome_filetype + "-o'></i>"
     html.html_safe
   end
+
+  def show_poll(poll_id = nil)
+    render partial: "shared/polls", locals: { poll_id: poll_id }
+  end
+
+  def next_poll(poll_id = nil)
+    poll = GobiertoParticipation::Poll.find(poll_id) if poll_id
```
ferblape: Filter by current_site

app/helpers/application_helper.rb
```diff
@@ -62,4 +62,32 @@ def filetype_icon(attachment)
     html = "<i class='fa fa-file" + fontawesome_filetype + "-o'></i>"
     html.html_safe
   end
+
+  def show_poll(poll_id = nil)
+    render partial: "shared/polls", locals: { poll_id: poll_id }
```
ferblape: I'd check here here also that the poll is from the current site.

If the poll id is wrong I would return `render nothing: true`

app/helpers/application_helper.rb
```diff
@@ -62,4 +62,32 @@ def filetype_icon(attachment)
...
+  def next_poll(poll_id = nil)
+    poll = GobiertoParticipation::Poll.find(poll_id) if poll_id
+    answerable_polls = GobiertoParticipation::Poll.by_site(current_site).answerable.order(ends_at: :asc)
+    answerable_polls_by_user = answerable_polls.select { |p| p.answerable_by?(current_user) } if current_user
+
+    if current_user
+      if poll
+        if poll.answerable_by?(current_user)
+          poll
+        else
```
ferblape: Use `elsif` to save a couple of lines.

app/helpers/application_helper.rb
```diff
@@ -62,4 +62,32 @@ def filetype_icon(attachment)
...
+          if answerable_polls_by_user
+            answerable_polls_by_user.first
+          end
+        end
+      else
```
ferblape: Use `elsif`

app/helpers/application_helper.rb
```diff
@@ -62,4 +62,32 @@ def filetype_icon(attachment)
...
   end
+
+  def show_poll(poll_id = nil)
+    render partial: "shared/polls", locals: { poll_id: poll_id }
+  end
+
+  def next_poll(poll_id = nil)
+    poll = GobiertoParticipation::Poll.find(poll_id) if poll_id
+    answerable_polls = GobiertoParticipation::Poll.by_site(current_site).answerable.order(ends_at: :asc)
+    answerable_polls_by_user = answerable_polls.select { |p| p.answerable_by?(current_user) } if current_user
```
ferblape: Use `.detect` to avoid doing `.first` later.

app/views/shared/_polls.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="p_1 filled_block content_block" style="background-image: url(<%= image_url("illustrations/emojis.svg") %>)">
+  <h2><%= t("shared.polls.opinion") %></h2>
+  <% if next_poll.is_a?(::GobiertoParticipation::Poll) %>
```
ferblape: `next_poll(poll_id)`, right?

ferblape: In that case to avoid calling next_poll two times assign it to a variable `poll = next_poll(poll_id)` and check if poll is present.

ferblape (changes requested):

furilo:
- [ ] You are showing the title of the poll, the idea was to show the title of the question +  the options (as per the sandbox). Although maybe it makes sense to show both... Although we may quickly ran out of space. Lets for the moment include poll title, first question + options. 
- [ ] If I'm logged out, click on participate, the modal appears, I do login, you take me to the home again, forcing me to repeat the action I just did.

furilo:
- [ ] If you don't specify an id, you get Liquid error: internal - We should be able to just call show_poll() and it should show randomly any of the available polls

apradillap:
Ok, I'Il add that option

furilo:
Well, it was something that was working before, right?

apradillap:
Before the migration to Liquid it worked, I fix it

apradillap:
@furilo I've added the possibility that show_poll is called without parameters. Deployed in staging

app/views/gobierto_participation/welcome/index.liquid
```diff
@@ -9,9 +9,7 @@
       </div>
 
       <div class="pure-u-1 pure-u-md-8-24">
-        {% comment %}
-          show_poll()
-        {% endcomment %}
+        {% show_poll(poll_id: 353021530) %}
```
ferblape: Could you remove the hardcoded id?

ferblape (changes requested):
LGTM! Just remove the hardcoded ID and feel free to merge and release.

PR #1194, opened 2017-11-30
Show table with all docs

### LINKED ISSUE #1193: Show table with all docs
When uploading docs to pages, they are not added to any collection. So it's not possible to browse all uploaded documents. When adding pages with attachments to a site, I'm finding difficulties to retrieve them, so as a temp and quick fix I'd add a table showing all docs in /admin/attachments/file_attachments, below what we have now. 

Columns: 

- Edit icon
- Title (if exists)
  filename (in a second line in the same without path)
- Created at
- Updated at

In master we don't have yet a public view for a doc, right? When we have, we'd include a public view icon

apradillap:
Closes #1193 

### What does this PR do?

Show table with all docs (in/out collection) in /admin/attachments/file_attachments

/admin/attachments/file_attachments


furilo:
Dale

Issue #1132, opened 2017-11-17
Offer budget-exectuion data in a reusable format in /datos


furilo:
- [x] Decide data format. Somehting like: 
  - ID
  - code
  - year
  - area: (economic, functional, custom)
  - income/expense
  - name
  - description
  - value_budget_initial
  - value_budget_modified
  - value_execution
  - level
  - parent_code
  - updated_at
  - municipality_id
- [ ] Include links to /data section (@furilo)
  - Local menu?
  - Besides tables?

PR #1200, opened 2017-12-01
1132 budget execution data

entantoencuanto:
Retated to #1132

### What does this PR do?
Generates data to be exported in json and csv formats, grouping budget lines information from several indexes under the columns/attributes `value_budget_initial`, `budget_column_
`value_budget_modified`, `value_budget_execution`.

Visit http://madrid.gobify.net/datos

No

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -370,6 +370,19 @@ html.turbolinks-progress-bar::before {
         margin: .5em 0;
       }
     }
+    .subblock {
+    border-top: none;
+    padding: 0 0 2em 0;
+    margin: 0 0 2em 0;
+    }
+    .data_subitem {
+      margin: 0 0 .5em 0;
+      h3 {
+        display: inline-block;
+        vertical-align: middle;
+        margin: 0;
+      }
+    }
     h3 {
       display: inline-block;
       vertical-align: middle;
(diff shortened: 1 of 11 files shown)


ferblape:
Pending things:

- I forgot to mention that you should filter and get only those which custom_code and functional_code is nil
- updated at date. The value is read from the Activity event, if nil should be the beginning of the year
- I'd add a year field
- it's kind of weird that some budget lines don't have initial value but do have execution. Could you double check those?

ferblape (changes requested):
Pending things:

- I forgot to mention that you should filter and get only those which custom_code and functional_code is nil
- updated at date. The value is read from the Activity event, if nil should be the beginning of the year
- I'd add a year field
- it's kind of weird that some budget lines don't have initial value but do have execution. Could you double check those?

entantoencuanto:
I've been exploring data, and for some levels there is execution information but not initial value. For example the 22 code for budget lines, for Income in 2015, in place with id = 28079 and economic area:

* At the first level, with 22 code, there are initial and execution values.
* At the next level there is only 220 code, and only execution value, coincident with the parent 22 level execution value.
* At the deeper level, there are 220-00, 220-01, ... , 220-06 budget lines, with both initial and execution values. The sum on each category (initial/execution) is coincident with the same parent value.

In the controller:

* The search on index view of budgets looks for data on index_forecast (the index for initial value), and then looks for the rest of data on the other indexes, so always there is shown data with an initial value.
* In the show action of an specific budget line, the method looks for related data on the same default index_forecast, so, for example, trying to show information of a 220-00 code of 2015 (http://madrid.gobify.net/presupuestos/partidas/220-00/2015/economic/I) the controller fails trying to get the 220 parent code information.

The export action of controller takes data from all indexes and join it for each id formed with the place id, the year, the budget line code, the kind (expense, income) and budget area. If there is only execution data for an id, the corresponding row will have empty the column for initial. The action can be changed to export rows with initial data for a given id only.

PR #1207, opened 2017-12-03
Position for section items is not working properly

### LINKED ISSUE #1206: Position for section items is not working properly
The order in the section tree is different from the front

apradillap:
Closes #1206

### What does this PR do?

Alvaro told me that he didn't follow the same order of the tree as in the front, he was absolutely right.

Compare section tree and front section tree order


furilo:
I've been playing around and can't break it ;) @ferblape all yours.

PR #1212, opened 2017-12-04
Fix IBM Notes cal integration when no future events

amiedes:
Fixes [this Rollbar](https://rollbar.com/Populate/gobierto/items/648/)


ferblape (approved):
👍

PR #1213, opened 2017-12-04
Fix levels parameter for list_children_pages

### LINKED ISSUE #1203: Levels parameter for list_children_pages is not working properly
In https://gobiernoabierto.getafe.es/s/portal-de-transparencia/publicidad-activa for example: 

- if you don't specify the parameter, two sections show children (but more levels have children and are not being shown)
- if you put 0, a third section show children
- if you put 1, same as first case
- if you put 2, the sections that didn't show anything when 1 show chidren, but those who did, stop doing it
...

Other issues: 

- [ ] We should put a relative URL (don't know how Rails usually do this), or if we include the domain, take into account the https
- [x] The order of the children is not taking into account the section info. I guess that's because we are looking at the collection instead of the section... Since we are probably doing some changes to the interaction of collections-sections, we can leave this aside for the moment.

apradillap:
Closes #1203 

### What does this PR do?

Review list_children_pages tag, wasn't working properly.

My test:

[http://participacion.gobify.net/s/participacion/ayuda-global-del-sitio](http://participacion.gobify.net/s/participacion/ayuda-global-del-sitio)

The order of the children is fixed in https://github.com/PopulateTools/gobierto/pull/1207

{% list_children_pages page_slug %} (by default 1 level)
{% list_children_pages page_slug | levels: 0 %}
{% list_children_pages page_slug | levels: 1 %}
{% list_children_pages page_slug | levels: 2 %}
...

### DIFF
--- a/app/helpers/gobierto_cms/page_helper.rb
+++ b/app/helpers/gobierto_cms/page_helper.rb
@@ -1,17 +1,20 @@
+# frozen_string_literal: true
+
 module GobiertoCms
   module PageHelper
     def section_tag(nodes, viewables)
-      html = ""
+      html = []
       html << "<ul>"
       nodes.each do |node|
-        html << "<li>" + link_to(node.item.title, node.item.to_url(section: true, host: current_site.domain))
+        html << "<li>" + link_to(node.item.title, node.item.to_url(section: true,
+                                                                   host: current_site.domain))
         if node.children.any? && !(viewables & node.children).empty?
           html << section_tag(node.children, viewables)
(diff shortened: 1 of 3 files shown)


furilo (approved):

furilo:
For next PRs, where pertinent, please provide a live example of the feature

ie. 
http://getafe.gobify.net/admin/cms/pages/16/edit?collection_id=43
http://getafe.gobify.net/s/portal-de-transparencia/informacion-del-portal

PR #1269, opened 2017-12-19
190 id number storage

entantoencuanto:
Closes [#190](https://github.com/PopulateTools/issues/issues/190)

### What does this PR do?
Allows to store an encrypted ID number as a new user verification type. The model makes use of a key identifier which value is expected to be defined in application secrets.

Define in `config/secrets.yml` a new secret using RSA public key algorithm for the value. To create a valid key use, for example:
```ruby
OpenSSL::PKey::RSA.new(2048).public_key.to_pem
```
You can create a new id number verification calling:
```ruby
verification = User::Verification::IdNumber.new(encryption_key: key_secret_identifier, id_number: id_number)
```

Once saved, `verification` will only store the `id_number` encrypted with the previously defined key. This value is accessible calling `verification.encrypted_id_number`

Not directly, but to create valid User::Verification::IdNumber instances at least a public key using RSA public key algorithm should be defined in `config/secrets.yml` and used in encryption_key attribute.


ferblape (changes requested):
@entantoencuanto secrets.yml is commited and part of the repository. Sensitive information is stored in the environments file. 

So could you update the PR with the following changes?

- update and commit secrets.yml referencing an environment variable
- update .env.example with the new variable
- update the environment variables document in docs/

PR #1277, opened 2017-12-26
Ruby 2.5.0

ferblape:
### What does this PR do?

This PR upgrades Ruby version to 2.5.0

Check everything works in staging. 

(we'll test it in staging for one week or two, and then we'll release it to production)


ferblape:
A rollbar related with TLS and Microsoft Exchange gem: https://rollbar.com/Populate/gobierto/items/688/

PR #1284, opened 2017-12-28
Mobile menu

### LINKED ISSUE #1217: Improve mobile menu
@Crashillo Sketch file shared in Dropbox (look at the page "To HTML")

Marvel reference: https://marvelapp.com/12047afg/

Crashillo:
Closes #1217 

### What does this PR do?
Adapt header & menus to mobile resolutions

### Pending

- [x] Get the current submodule name
- [x] Render, in rails, the `secondary_nav`, even you're in other module (it's handled by frontend)
- [x] Once fixed, split header main files into partial files


ferblape:
This is ready to review in staging, except for the secondary navigation in the Participation module, which I prefer to discuss next Monday.

I have also a couple of questions:

1. should we hide the submodule navigation when there are not submodules? Example in: http://getafe.gobify.net/s/portal-de-transparencia/informacion-del-portal

screen shot 2018-01-05 at 15 53 55

2. also in the submodule navigation, when there's no current option (example, in the home of participation http://participacion.gobify.net/participacion) should we add a placeholder?

menu

3. I have removed the dropdown in the Login/Signup option. I think it's correct but we can add it back again:

screen shot 2018-01-05 at 15 55 55

Crashillo:
> 1. should we hide the submodule navigation when there are not submodules? Example in: http://getafe.gobify.net/s/portal-de-transparencia/informacion-del-portal

Hide submenu is the best choice, I'd say.

> 2. also in the submodule navigation, when there's no current option (example, in the home of participation http://participacion.gobify.net/participacion) should we add a placeholder?

I'd suggest to add a new label called `home` or so. Right now, the only way to go back, if you're in a different submodule, would be tapping the module name? A `home` label would repeat the options to go back, but I believe is more intuitive.

ferblape:
I'm adapting the menu for Gobierto plans to the mobile form.

ferblape:
Merged because we need some commits here and because it's better than we have now. If new issues appear, we'll create separated issues.

Issue #1267, opened 2017-12-19
Gobierto authentication/authorization strategies


ferblape:
Gobierto needs to integrate with different authentication/authorization strategies, such as OAuth, SAML, or a SSO.

The objective of this issue is to define the minimal changes in the code required to introduce a first strategy, with the following requirements.

- strategies code will live outside Gobierto, except those that are very generic becuase implement standar protocols, i.e OAuth. In this case, the strategy should be completly defined using parameters, such as URLs, keys, tokens

- the current login/signup will remain as an strategy integrated in the core of the project. Maybe in the future, for clarity, it'll make sense to extract it from the core

- strategies code will be added in deploy time to `vendor/auth-strategies/` folder

- a basic strategy will be implemented in `vendor/auth-strategies/` as a reference/template and for testing purposes

- the name and the availability of the strategy will be defined in `application.yml`

- a strategy will contain:
  - a `validations/` folder with the required validations
  - a `views/` folder with the templates
  - a `form/` folder for the new form objects
  - a `config/locales/` folder with the translations
  - tests
  - a README
  - a configuration file

- Gobierto should detect and extend its settings to include the folders of the strategies

- this process should be documented in `doc/`

PR #1286, opened 2017-12-28
1267 gobierto auth strategies

entantoencuanto:
Related with #1267 

### What does this PR do?
Allows the use of alternative mechanisms for users authentication / authorization in site.

By default the site has the email/password mechanism, but from admin site administration there are different authentication modules which can be enabled. If one of them is enabled, the application uses an alternative session form or redirects to another site whose responsibility is the authorization of the user. The content of the alternative session form is provided by the auth strategy. The controller uses an object form provided by the auth strategy module, and it contains the logic to create a valid user session. The auth strategy files must be stored under vendor/auth_strategies/ An example of the structure of an strategy can be reviewed in vendor/auth_strategies/null_session/

Enable "Null Session" from site admin configuration and try to make sign in ;)

- [x] new entry in `config/application.yml`?
These lines has been added to application:
```
  auth_modules:
    -
      name: null_strategy
      description: Null Strategy
      session_form: NullStrategySessionForm
      default: false
```
Further modules can be added there and they will be available in the admin site configuration form


entantoencuanto:
It appears that with the last commit the problems with staging environment eager loading are solved. Now the namespace associated with custom authentication is available and the Null Strategy works as expected.

entantoencuanto:
Some changes has been added:
* For the links in application to actions only allowed to authenticated users which open via AJAX a login form, instead of make a redirect when the authentication strategy is defined outside, a message with a link to authentication provider is shown.
* After registration, the password fields can be disabled by the auth strategy on confirmation and settings forms. If none of the auth strategies enabled for a site include `password_enabled: true` in `application.yml` password fields won't be shown in forms and related validations won't be performed.

ferblape:
@entantoencuanto we've just realized that the strategies are available in all the sites. For staging that's fine, but in production we should limit this, and the strategy should be only visible in the sites configured in the YAML file of the strategy. What do you think?

entantoencuanto:
I can add an attribute to configuration with a white list of the sites a strategy is available

entantoencuanto:
Added a configuration option in `application.yml` for auth strategies named `domains`:

* If not present or with nil value the strategy will be available for all sites.
* To restrict the availability of the auth strategy for a group of sites the value of this attribute must be an array containing their domains.

ferblape:
Please, review this flash message:

screen shot 2018-01-16 at 11 55 38

ferblape:
I also found a translation missing:

screen shot 2018-01-16 at 12 04 46

entantoencuanto:
The missing translation issue has been included in #1343

PR #1287, opened 2017-12-28
1265 verification filter

### LINKED ISSUE #1265: Enable tools verification level check
When assigning a tool to a phase of a process the verification level is required to be defined, but the setting is not yet used when citiziens use the tool.

Implement the corresponding controller filters in polls and pools that read the verification level setting and perform the corresponding check (email/session verification or census verification for the moment).

entantoencuanto:
Closes #1265

### What does this PR do?
Defines a model concern to manage a `visibility_user_level` enum defined for some tools of Gobierto Participation module. This concern provides a method to check if a user has permissions over a resource depending on visibility level. By default, there are only 2 levels:

* `registered`: The user must exist in database and confirmed.
* `verified`: The user must be verified in the site the resource belongs to.

More levels can be created for a resource by defining them in the `visibility_user_level` enum and defining in User model or decorator an instance method called `verified_in_level_with_site?` which accepts the `visibility_user_level` name and the site.

This concern is used with `GobiertoParticipation::Poll` and `GobiertoParticipation::ContributionContainer` in controllers to filter creation, edition or deletion of depending resources, as poll answers, contributions, or votes / comments on contributions. The filter is stored in `app/controllers/concerns/user/verification_helper.rb` and defines the raised error and redirection for visibility_user_level defined on the resource if the user does not pass the validation. In this way if some resource requires an specific validation the user will be redirected to the mechanism associated with de validation, instead of receive a generic error message.

From admin part define a poll or contribution_container in the corresponding stage of a process on a site and change settings in the edition form, in the USERS select. Then visit the page not logged in, with a confirmed user but not verified in the site and with a confirmed and verified in the site

No

### DIFF
--- a/app/controllers/concerns/user/verification_helper.rb
+++ b/app/controllers/concerns/user/verification_helper.rb
@@ -3,6 +3,12 @@ module User::VerificationHelper
 
   private
 
+  def check_visibility_level(resource, user)
+    unless resource.visibility_level_allowed_for?(user)
+      raise_visibility_forbidden(resource, user)
+    end
+  end
+
   def verify_user_in!(site)
     raise_user_not_verified(site) unless user_verified_in?(site)
   end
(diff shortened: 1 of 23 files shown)


ferblape:
Tested in staging works like charm 👍

PR #1302, opened 2018-01-08
Soft delete

### LINKED ISSUE #1128: Archive elements as a way to soft-delete
I propose that we implement an archive function that would serve as a soft-delete. 

- In the front: Archived elements would behave as deleted elements. That is, they should not appear in any public place.
- In the admin: we would create a new classification for Archived elements. If you archive an element, it would go to this page; you would be able to recover (un-archive) an element and it would come back to normal.

This is an inventory of entities I would implement this for: 

## General

- [x] Pages
- [x] Events
- [x] Documents

## Participation

- [x] Processes
  - [x] Polls
  - [x] ContributionContainers

Please @apradillap give it a thought of what more elements we should take into account.

---

For the front-end, let's do the following: 

- in list tables, add a column at the end of the table, where we put the icon trash, with class "tipsit ", and title: "Archive this element"
- when clicking on trash, we show this JS alert: "Are you sure you want to archive this element? It won't show on the front end neither on lists like this. You will be able to recover it"
- After the table, if archived elements exist, add: 
```
<div class="m_v_1 right">
<a href="">Archived elements</a>
</div>
```
- when clicking on "archived elements", show a table with the same columns but without linking the elements, except for the first one (edit icon), and a last column that reads "Recover element", with class "tipsit" and title: "If you recover an element, it will be listed again"

apradillap:
Closes #1128 

### What does this PR do?

Add archiving possibility to:

- Page
- Attachment
- File
- Poll
- Contribution Container

These objects will have an archived_at field showing the date when it was archived.

Additionally:

- Fixed some bugs with the Page and News division.
- Fixed some bugs relation with collection pages in processes.

When archiving any object you will see a table below where you can retrieve the object.


ferblape (approved):
👍 good job!

PR #1326, opened 2018-01-12
Phases tool content

### LINKED ISSUE #1260: Phases / Tool content
- [ ] I've updated the view for the "content" tool for a phase: http://madrid.gobierto.dev/sandbox/admin_part_group_info (in 1234-modify-info-about-phases). The idea is to just show a select2 to be able to select a page. When a page is selected, we would offer a link to directly edit it. If the user needs to create a new page, we provide a link to do it.

imagen

- [ ] I'm seeing that we are using the news template to show this: http://newalcobendas.gobify.net/participacion/p/mejoremos-la-ciudad/noticias/acerca-de - We should use a new template with a simple layout, not making references to news nor pages. So maybe create a views/gobierto_participation/shared/pages/_information.html.erb copying an actual page, make this tool use it, and then I can finish the HTML/CSS details. We just need to change the template, the rest (menus, etc) is OK.

apradillap:
Closes #1260 

### What does this PR do?

- Added  help text to associate a page with a stage.
- To reuse the contoller in the Process and use a different view for the phase helper, I pass a parameter through which I show the partial: 

`app/views/gobierto_participation/shared/pages/_information.html.erb
`

Example: http://participacion.gobify.net/participacion/p/creo-un-proceso


furilo:
If you browse to any of the links and come back, the form is duplicated: 

captura de pantalla 2018-01-12 a las 14 20 03

apradillap:
Fixed @furilo I saw that forms with select2 duplicate with back button on browser in rails 5:  https://stackoverflow.com/questions/41442531/forms-with-select2-are-duplicated-when-clicking-back-button-on-browser-in-rails

apradillap:
Review in Safari

ferblape:
I still can see the duplicated form when clicking the back button (in Firefox). 

BTW it's a bit confusing for the admin to be out of the phases editor when editing a page. Should we try to add a breadcrumb, or redirect to the phases editor again after saving the page?

apradillap:
@ferblape I fixed the duplicity of select2 for Safari and Firefox and I have added breadcrumb to return to the editing of phases when the page is not a Process new because with the tabs it is easy to return to the editing of phases if is a Process New

ferblape (changes requested):
This conditions a bit a bit complex and fragile. What if we add a parameter in the "Edit this page" link to detect that we are editing a page of a process? 

The parameter could be something like "stage_id", and use it to bould the back link.

ferblape:
@apradillap I got this rollbar when creating a page in staging https://rollbar.com/Populate/gobierto/items/726/

apradillap:
ok, You're right @ferblape , I'm going to modify it.

apradillap:
@ferblape I already simplified it and corrected the rollbar error

Issue #1312, opened 2018-01-09
Scope users in sites


ferblape:
User model was designed to be available in all sites, and using the field `source_site_id` we know the first site where she signed up.

We are going to change this approach and link users to sites. Therefore we need:

- add a `site_id` field, using the value from `source_site_id`
- remove `source_site_id` field
- scope the validations to the `site_id`, so we could have multiple users with the same email in different sites
- user this scope in the Session / Signup controllers
- update Form objects
- update tests and fixtures, taking care of the users in the current site
- update docs/user-namespace.md
- amend admin users list to fetch only the users in that site

entantoencuanto:
Deployed in staging. The part of confirmation token and password recover must be reviewed, it seems that there are some cross-site errors calling to a site with the token of another.

PR #1327, opened 2018-01-12
Scope users in sites

entantoencuanto:
Related with #1312

### What does this PR do?
* Renames the source_site attribute of users with site.
* Changes the association between `User` and `Site` adding a `has_many :users, dependent: :nullify`.
* Removes the unique database index for `email` on users and creates a new one unique for `email` and `site_id`.
* Defines a validation of uniqueness of `email` scoped to site in User model.
* In `sessions_controller` a user will only be able to sign in in their site. This is implemented by including the site in the `User::SessionsForm`.
* Refactor tests taking into account the site and includes some test:
  * Verifies a user can't use email/password to sign in in wrong site.
  * Verifies the list of users in admin/users only displays users of the site.
  * Test that `User::SessionsForm` must include the site to be valid.

Register with the same email in different sites, try to sign in with the credentials of one site in other. As admin user take a look to the list of users of a site.

No

### DIFF
--- a/app/controllers/concerns/gobierto_admin/site_session_helper.rb
+++ b/app/controllers/concerns/gobierto_admin/site_session_helper.rb
@@ -8,8 +8,8 @@ module GobiertoAdmin
       @current_site ||= begin
         if session[:admin_site_id] && (matched_site = Site.find_by(id: session[:admin_site_id]))
           SiteDecorator.new(matched_site)
-        else
-          SiteDecorator.new(managed_sites.first) if managed_sites.present?
+        elsif managed_sites.present?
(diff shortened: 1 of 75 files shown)


ferblape (changes requested):
I have created a new user in burjassot.gobify.net and automatically I got logged in dival.gobify.net and the rest of gobify.net domains. You should limit the cookie domain to the site domain for users

entantoencuanto:
Last commits implements the requested change. The users session won't remain between different sites, also admins sessions.

Also, some commits has been added to fix #1342 and some redirection errors signing out from admin session.

ferblape (approved):
🚀

PR #1338, opened 2018-01-15
Allow collections of the same container if item_type is Page

### LINKED ISSUE #1322: Allow collections of the same container if item_type is Page
Modify validation if the collection is a page there is no uniqueness

https://github.com/PopulateTools/gobierto/blob/master/app/models/gobierto_common/collection.rb#L15

### LINKED ISSUE #1332: some help/direction needed for this newbie
I am trying to install/test gobierto app. Can I please get some direction? I did docker compose up and  I see Rails page when I load the sites.. also I see 5 users in the db but not sure how I can get to login on /admin.
screen shot 2018-01-12 at 5 47 35 pm

I am using docker for mac for testing the app.

apradillap:
Closes #1322 

### What does this PR do?

Modify validation if the collection is a page there is no uniqueness

For example, you can create several collection with item_type News

### DIFF
--- a/app/models/gobierto_common/collection.rb
+++ b/app/models/gobierto_common/collection.rb
@@ -14,11 +14,21 @@ module GobiertoCommon
 
     validates :site, :title, :item_type, presence: true
     validates :slug, uniqueness: { scope: :site }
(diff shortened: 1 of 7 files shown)


ferblape (approved):
👍

PR #1347, opened 2018-01-17
Error with archived elements

### LINKED ISSUE #1340: Error detecting existing elements already archived
This is an issue in `GobiertoCalendars::Event` but potentially can affect to all the classes that support soft-delete.

Before creating and event, the process checks in the database if the event exists with this query:

```sql
SELECT  1 AS one FROM "gc_events" WHERE "gc_events"."archived_at" IS NULL AND "gc_events"."site_id" = 9 AND "gc_events"."slug" = $1 LIMIT $2  [["slug", "2018-01-15-despatx-amb-el-primer-tinent-d-alcaldia"], ["LIMIT", 1]]
```

Notice the `archived_at IS NULL` statement. If there's an element archived this query won't detect it, and the INSERT query will fail 💥:

```sql
INSERT INTO "gc_events" ("starts_at", "ends_at", "state", "created_at", "updated_at", "title_translations", "description_translations", "external_id", "site_id", "slug", "collection_id") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) RETURNING "id"  [["starts_at", "2018-01-15 11:00:00"], ["ends_at", "2018-01-15 12:00:00"], ["state", 1], ["created_at", "2018-01-16 14:26:31.360783"], ["updated_at", "2018-01-16 14:26:31.360783"], ["title_translations", "{\"es\":\"@ Despatx amb el primer tinent d'alcaldia\"}"], ["description_translations", "{\"es\":null}"], ["external_id", "F4DEE6203892A72DC125821200401D31-Lotus_Notes_Generated"], ["site_id", 9], ["slug", "2018-01-15-despatx-amb-el-primer-tinent-d-alcaldia"], ["collection_id", 168]]
```

Rollbar: https://rollbar.com/Populate/gobierto/items/725/

apradillap:
Closes #1340 

### What does this PR do?

When events are synchronized and there is an archived event with same slug, it gives an error when making the insert.

The after_destroy I was able to put it in the concern but not the callback after_restore.

Besides, as there is an index I have to put another slug for the archived item, adding "-slug-archived"

Error doesn't repeat https://rollbar.com/Populate/gobierto/items/725/?item_page=0&#instances

You can archive an event (see the slug from edition) and then restore the event.

### DIFF
--- a/app/models/concerns/gobierto_common/sluggable.rb
+++ b/app/models/concerns/gobierto_common/sluggable.rb
@@ -12,7 +12,7 @@ module GobiertoCommon
     private
 
     def set_slug
-      if slug.present? && !slug.include?("-slug-archived")
+      if slug.present? && !slug.include?("archived-")
         self.slug = self.slug.tr("_", " ").parameterize
         return
       end
@@ -39,7 +39,7 @@ module GobiertoCommon
 
     def add_archived_to_slug
       unless destroyed?
-        update_attribute(:slug, slug + "-slug-archived")
+        update_attribute(:slug, "archived-" + id.to_s)
       end
     end
   end
--- a/app/models/concerns/gobierto_common/sluggable.rb
+++ b/app/models/concerns/gobierto_common/sluggable.rb
@@ -6,12 +6,13 @@ module GobiertoCommon
 
     included do
       before_create :set_slug
+      after_destroy :add_archived_to_slug
     end
 
     private
 
     def set_slug
-      if slug.present?
+      if slug.present? && !slug.include?("-slug-archived")
         self.slug = self.slug.tr("_", " ").parameterize
         return
       end
@@ -34,6 +35,12 @@ module GobiertoCommon
       end
 
       self.slug = new_slug
+    end
+
+    def add_archived_to_slug
+      unless destroyed?
+        update_attribute(:slug, slug + "-slug-archived")
+      end
     end
   end
 end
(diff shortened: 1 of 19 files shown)


ferblape:
Now we got a new rollbar when archiving elements already archived:

https://rollbar.com/Populate/gobierto/items/748/

app/models/concerns/gobierto_common/sluggable.rb
```diff
@@ -35,5 +36,11 @@ def set_slug
 
       self.slug = new_slug
     end
+
+    def add_archived_to_slug
+      unless destroyed?
+        update_attribute(:slug, slug + "-slug-archived")
```
ferblape: I'd change the slug format to something like "archived-<superlong random string>", to avoid the rollbar I mentioned in the comments

app/models/gobierto_attachments/attachment.rb
```diff
@@ -50,6 +50,7 @@ class Attachment < ApplicationRecord
 
     after_create :add_item_to_collection
     before_validation :update_file_attributes
+    after_restore :restore_slug
```
ferblape: Why no set_slug instead of a new method that calls set_slug?

ferblape: Why no set_slug instead of a new method that calls set_slug?

ferblape (changes requested):
👍 just avoid the issue with the repeated slugs and we got it.

PR #1348, opened 2018-01-18
Remove filtering text option

### LINKED ISSUE #1339: Calendar filtering rules option to remove the filtering text
Add an option to the calendar filtering rules to remove the text used to filter from the event. It could be a boolean attribute.

For example, if somebody creates a rule that imports only events that contain the string "[public]", and the rule activates this option, the events shouldn't contain that string.

ferblape:
Closes #1339

### What does this PR do?

This PR adds a new option to filtering rules to remove the text used to filter in the processed event.

It's not finished since I need some help with the CSS @furilo @Crashillo:

screen shot 2018-01-18 at 15 06 06

1. Create a rule
2. Activate the new option
3. Check the text is removed from the title or the description


Crashillo:
Styles fixed. Take a look at the conflict.

PR #1363, opened 2018-01-22
Use same sync_range for all kinds of events and for all different integrations

amiedes:
### What does this PR do?

Uses same sync range for recurrent and non-recurrent events and for all different calendar integrations, since this *"uncoherent"* behavior was creating confussion for some users.

Now events older than 5 days should not be imported / updated

No


ferblape (approved):
Good!

PR #1366, opened 2018-01-23
Scope event search on person when syncing calendars

amiedes:
### What does this PR do?

Fixes a bug that caused when two people in the same site have the same calendar configured, only one of them would retain the events at the end of the synchronization (because events in each person had the same `external_id`, but the search was not scoped on person events).

Add the same IBM Notes calendar integration for two people in the same site, sync, and at the end both should have the corresponding events.

No


ferblape (approved):
👍

PR #1394, opened 2018-02-01
1240 - Simpler URL for docs

### LINKED ISSUE #1240: URL for docs
@furilo commented on [Thu Dec 07 2017](https://github.com/PopulateTools/issues/issues/184)

For Getafe we won't make it, but maybe it makes sense to have a kind of wrapper for the absolute URL for a doc. When we list the absolute URL, we wouldn't show the actual path (ie. https://gobierto-populate-production.s3.eu-west-1.amazonaws.com/site-14/attachments/attachment-f017a860-1ce4-4f16-bacb-a67e64d31f59/presupuesto_completo_12_0_9899_1.pdf) but something like: 

https://gobiernoabierto.getafe.es/docs/329 

(or a more obscure identifier if want to avoid easy scrapping)

Now, this would just be a redir to the original path. It would avoid having full paths in a page, which is ugly. Also, it may solve https://github.com/PopulateTools/gobierto/issues/733

In the future, this would allow us things like: 

- a counter for downloads
- a viewer for docs, where we can add UI to see previous versions (if/when we have that)
- tools to fav a doc, or track it
- and the most important thing: changing a doc location (also related with https://github.com/PopulateTools/gobierto/issues/733)

All without worriyng about changing anything

amiedes:
Closes #1240 

### What does this PR do?

Provides al alternative (and simpler) URL for accessing attachment files.

When accessing documents directly, they should be displayed inside an iframe and the URL should match */docs/\<id\>*. Test this:

* With this image: [http://madrid.gobify.net/docs/33](http://madrid.gobify.net/docs/33)
* With this pdf: [http://madrid.gobify.net/docs/87](http://madrid.gobify.net/docs/87)

When rendered inside a CMS page, images should be displayed correctly. Test it in this page: [http://madrid.gobify.net/paginas/pagina-con-archivos](http://madrid.gobify.net/paginas/pagina-con-archivos). Also, when clicking on document link in a page, like this one:

imagen

the result URL should match */docs/\<id\>* and the file will be displayed inside an iframe.

When using the "Include attachment" widget in the admin and clicking on "Copy link", a human readable link should be given to the user (*wadus/docs/123*). Test it here: [http://madrid.gobify.net/admin/cms/pages/97/edit?collection_id=13](http://madrid.gobify.net/admin/cms/pages/97/edit?collection_id=13)

No


furilo:
To avoid opening a URL like https://gobierto-populate-staging.s3-eu-west-1.amazonaws.com/site-7/attachments/attachment-7ec0a83f-47db-4ce6-a9fa-dba8b14c10ab/GETAFE_Plano.pdf for browser-viewable resources, such as PDFs, which is one the things we want to avoid, I suggest that the /docs/ view generates an iframe, so in the browser we'll keep seeing the nice URL but the PDF is opened and viewable.

ferblape:
We are publishing open data, right? Why do we need to hide the URLs? 

Also, I'm not sure about obscuring the id with that formula. The idea is to make the scrapping more difficult, but using this formula the bad guys just need an id and can iterate on the following.

furilo:
> Why do we need to hide the URLs?

Because if someone is visiting gobiernoabierto.getafe.es and then end up
having a very long URL of gobierto.aws.etc it might be confusing (why is
the doc being downloaded from this domain if I was in another completely
different?) and the URL if ugly.

Also, I'm not sure about obscuring the id with that formula. The idea is to
> make the scrapping more difficult, but using this formula the bad guys just
> need an id and can iterate on the following.
>

Yes, I also think it's not needed. I said it in the other thread.

ferblape:
This PR requires to install `libmagic-dev` in the servers.

app/controllers/gobierto_attachments/attachments_controller.rb
```diff
@@ -0,0 +1,32 @@
+module GobiertoAttachments
+  class AttachmentsController < GobiertoAttachments::ApplicationController
+
+    before_action :load_attachment
+    layout false
+
+    def show
+      if @attachment.file_size < 100.megabytes
+        tempfile = open(@attachment.url)
```
ferblape: FYI: `OpenURI#open` doesn't return a `Tmpfile` always: http://winstonyw.com/2013/10/02/openuris_open_tempfile_and_stringio/

ferblape:
I think we got it with [this commit](https://github.com/PopulateTools/gobierto/pull/1394/commits/1f6ffa011425954d75ef7c0104e30582b8f9460e). Please @amiedes:

- check the different scenarios in case I forgot something
- add the pending tests
- update the PR description

amiedes:
I think everything is working as expected. Thanks for the help @ferblape.

The PR description contains links to test several documents.

ferblape (approved):

PR #1410, opened 2018-02-06
Renaming a slug in edition

### LINKED ISSUE #1404: Renaming a process slug should not accept an empty value
If you have a process, edit it and remove the value for slug, it gets saved as empty, so it breaks. If you "empty" the slug, it should be treated as when you are creating the resource and you don't enter any value.

apradillap:
Closes #1404 

### What does this PR do?

- With this PR slugs can be renamed in the edition
- I cleaned a repeated function and adapted the slug attributes of some models that had no attributes before_validation

You can create or edit slugs of pages, processes, issues, etc.
Available in staging

### DIFF
--- a/app/models/concerns/gobierto_common/sluggable.rb
+++ b/app/models/concerns/gobierto_common/sluggable.rb
@@ -5,7 +5,7 @@ module GobiertoCommon
     extend ActiveSupport::Concern
 
     included do
-      before_create :set_slug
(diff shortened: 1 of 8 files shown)


ferblape:
How are you testing this change?

ferblape (changes requested):
Tests, 🙏

ferblape (approved):

Issue #1373, opened 2018-01-25
Participation module subscriptions


ferblape:
(Follows #1372)

Amend the user subscriptions section to include the items from the Participation module:

1. include the module in the Area
2. include the list of processes in the items column (the one in the right)
3. include areas and scopes in the items column 

Extra ball:

1. when the "whole site" option is checked, disable with JS the rest of elements
2. when the "whole site" options is unchecked enable the rest of elements

PR #1411, opened 2018-02-06
1373 subscriptions edition

entantoencuanto:
Related with #1373

### What does this PR do?
* Adds a global participation area item and individual processes subscription items to user subscription preferences form
* When a user checks "Whole site" or a module the dependent checkboxes are marked and disabled
* If a user marks, for example, whole site and saves, only the site subscription is saved, and the other checked options are ignored, since "whole site" option has priority over the rest of subscriptions. The same occurs with module subscriptions.
* If a user unmarks "Whole site" or a module, if there was a previous subscription on a dependent item, the item appears as marked, because the subscription remains, and now will be used.

#### Pending:
* ~~Include individual `GobiertoCommon::Scope` and `Issue` instances in right subscription column~~

Visit http://madrid.gobify.net/user/subscriptions as signed in user

No

### DIFF
--- a/app/assets/javascripts/user/app/controllers/user_subscription_preferences_controller.js
+++ b/app/assets/javascripts/user/app/controllers/user_subscription_preferences_controller.js
@@ -3,24 +3,60 @@ this.User.SubscriptionPreferencesController = (function() {
 
   SubscriptionPreferencesController.prototype.index = function(){
     _handleSiteSelected();
+    _handleAreaSelected('gobierto_budget_consultations');
+    _handleAreaSelected('gobierto_people');
+    _handleAreaSelected('gobierto_participation');
   };
 
   function _handleSiteSelected(){
     $('#user_subscription_preferences_site_to_subscribe').on('change', function(e){
       if($(this).is(':checked') === false){
-        _uncheckAllOptions();
+        _uncheckAllModuleOptions();
       } else {
-        _checkAllOoptions();
+
+        _checkAllModuleOptions();
       }
     });
   }
 
-  function _checkAllOoptions(){
-    $('input[data-selected]').prop('checked', 'checked');
+  function _handleAreaSelected(area_name){
+    $('#user_subscription_preferences_modules_'.concat(area_name)).on('change', function(e){
+      if($(this).is(':checked') === false){
+        _uncheckAllAreaOptions(area_name);
+      } else {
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/user/app/controllers/user_subscription_preferences_controller.js
```diff
@@ -3,24 +3,60 @@ this.User.SubscriptionPreferencesController = (function() {
 
   SubscriptionPreferencesController.prototype.index = function(){
     _handleSiteSelected();
+    _handleAreaSelected('gobierto_budget_consultations');
```
ferblape, line 6: Having the three modules hardcoded doesn't scale well in the future.  I propose to add a special attribute (a `data-` or a `rel=`) and loop through them.

app/models/site_configuration.rb
```diff
@@ -22,7 +22,7 @@ class SiteConfiguration
 
   DEFAULT_LOGO_PATH = "sites/logo-default.png".freeze
 
-  MODULES_WITH_NOTIFICATONS = ["GobiertoPeople", "GobiertoBudgetConsultations"]
+  MODULES_WITH_NOTIFICATONS = ["GobiertoPeople", "GobiertoBudgetConsultations", "GobiertoParticipation"]
```
ferblape, line 25: Could you add a reference to this constant in https://github.com/PopulateTools/gobierto/blob/master/docs/developing-module.md ?

app/views/user/subscriptions/index.html.erb
```diff
@@ -42,16 +42,20 @@
...
-                <%= f.collection_check_boxes(:modules, @user_notification_modules, :first, :first) do |b| %>
-                  <div class="option ">
-                    <%= b.check_box({"data-selected" => f.object.modules.include?(b.value)}) %>
-                    <%= b.label do %>
-                      <span></span>
-                      <%= t(".modules.#{b.text}") %>
-                    <% end %>
-                  </div>
-                <% end %>
+              <% if @user_notification_modules.any? %>
```
ferblape, line 45: Maybe in the future we could divide this in partials and declare the partials in each module, as we do with the partials that implement the navigation.

ferblape (changes requested):
Code looks good and works fine. 

I have added a couple of improvements to make this more maintenable in the future.

PR #261, opened 2017-01-20
Display a Site logo placeholder if there isn't one uploaded yet

danguita:
Connects to #251.

### What does this PR do?

This implements a variant to the "logo" Site configuration option which fallbacks to the default one, which acts as a placeholder, if it isn't uploaded already.

### How should this be manually tested?

- [x] Just check that the placeholder logo is displayed when the Admin hasn't uploaded one yet.

screen shot 2017-01-20 at 8 19 21 am

PR #1419, opened 2018-02-08
261 events timeline limit

entantoencuanto:
related with [#261](/PopulateTools/issues/issues/261)

### What does this PR do?
* Limits to 5 the activities list appearing in index of scopes and issues
* Removes the hour present in events lists

Visit An issue or scope index page and check the second column for the first point and the first column for the second

No


ferblape:
👍 works as expected

PR #1439, opened 2018-02-12
Take into account groups

### LINKED ISSUE #1400: Take into account groups when printing "Sobre este proceso"
When showing a group, we are not discerning if it's a group or process when printing "Sobre este proceso": 

imagen

Take into account if its a group.

apradillap:
Closes #1400 

### What does this PR do?

Initially the issue concerned changing the text: "About this process" taking into account groups in show process/group.

I extended this bug to fix it:

- Follow button (front)
- Group/process edition messages (tabs and alerts)
- In addition, I have modified this text because it doesn't makes much sense if the options are process and **group**. I've left Tipy
captura de pantalla 2018-02-12 a las 12 48 52

- Show process/group
- Edit process/group
- Follow process/group


ferblape (approved):
👍

PR #1440, opened 2018-02-12
Use scopes for retrieving collection items

### LINKED ISSUE #1354: New API for retrieving elements belonging to a collection
As seen [here](https://github.com/PopulateTools/gobierto/blob/cf628a75730df8b145c60cb5a04b3c619ece7fb4/app/models/gobierto_cms/page.rb#L70-L101) the API for retrieving pages etc. from a collection is has become a bit confusing, we should find a clearer way of performing these operations. ¿Maybe took a a look at [this](https://github.com/PopulateTools/gobierto/blob/cf628a75730df8b145c60cb5a04b3c619ece7fb4/app/models/gobierto_common/collection.rb#L33-L51) also?

amiedes:
Partially closes #1354 

At the end I only did this little changes because it was more complex than what I thought at the beginning.

### DIFF
--- a/app/models/gobierto_attachments/attachment.rb
+++ b/app/models/gobierto_attachments/attachment.rb
@@ -71,17 +71,17 @@ module GobiertoAttachments
     end
 
     def self.file_attachments_in_collections(site)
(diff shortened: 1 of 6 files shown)


ferblape (approved):
👍 Scopes FTW!

PR #1462, opened 2018-02-19
View all documents/events link in participation/processes

apradillap:
Closes https://github.com/PopulateTools/issues/issues/270

### What does this PR do?

- In show/event/attachment (process): see al events of process
- In index/event/attachment (process): see al events of participation
- Fix url of events/attachments if belong to a process

**Extra:**

- I have placed the link at the bottom
- I have reviewed the filters by issue and their respective links
- I have added pending tests for Participation
- I've unified the texts
- I've unified the filtering by issue

Navigating through the events and documents of Participation/Processes.

### Pending

- [ ] The documents part would be pending

### DIFF
--- a/app/controllers/gobierto_participation/events_controller.rb
+++ b/app/controllers/gobierto_participation/events_controller.rb
@@ -44,10 +44,7 @@ module GobiertoParticipation
     end
 
     def container_events
-      module_events = GobiertoCalendars::Event.events_in_collections_and_container_type(current_site, "GobiertoParticipation")
-      processes_events = GobiertoCalendars::Event.events_in_collections_and_container_type(current_site, "GobiertoParticipation::Process")
-
-      @container_events = module_events.merge(processes_events)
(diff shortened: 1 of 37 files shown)


ferblape:
👍 What was proposed in the issue it's implemented successfully, but using these pages I've found a few issues that we could tackle in this PR:

### Documents / attachments

1. When you are inside [the documents of a process](http://newalcobendas.gobify.net/participacion/p/remodelacion-parque-navarra/documentos?issue_id=comercio), we should remove the list of issues, because it only makes sense when you are in the documents of the whole module.

### Calendars

1. Same thing, in the Agenda of a process/group, remove the list of issues, because doesn't make sense

2. The calendar navigation is broken. I think it's because you are using the same instance variable to render the events in the calendar and in the list, and they should be different. The calendar should always render all the events, to allow the user to navigate

3. In the event page, could we remove "Documents" if there aren't documents attached? It's a bit confussing

PR #1465, opened 2018-02-19
View past events in participation/processes

apradillap:
Closes https://github.com/PopulateTools/issues/issues/271

## :v: What does this PR do?

- Add link **View past events**
- Add link **View current events**

You can create past or future events to see how they appear on Participation and on processes.

### Before this PR

captura de pantalla 2018-02-19 a las 18 57 10

### After this PR

captura de pantalla 2018-02-19 a las 18 52 13

captura de pantalla 2018-02-19 a las 18 52 31


app/views/gobierto_participation/events/index.html.erb
```diff
@@ -32,12 +32,30 @@
...
+          <% unless @past_events %>
+            <div>
+              <small><%= link_to t('gobierto_participation.shared.past_events'), gobierto_participation_past_participation_events_path %></small>
+            </div>
+          <% end %>
         </div>
       </div>
 
 
       <div class="pure-u-1 pure-u-lg-2-3">
```
furilo: why all these divs?  just put a `<p><a href="">View current events</a></p>`

ferblape:
I'm adding this comment to this PR:

👍 What was proposed in the issue it's implemented successfully, but using these pages I've found a few issues that we could tackle in this PR:

### Documents / attachments

1. When you are inside [the documents of a process](http://newalcobendas.gobify.net/participacion/p/remodelacion-parque-navarra/documentos?issue_id=comercio), we should remove the list of issues, because it only makes sense when you are in the documents of the whole module.

### Calendars

1. Same thing, in the Agenda of a process/group, remove the list of issues, because doesn't make sense

2. The calendar navigation is broken. I think it's because you are using the same instance variable to render the events in the calendar and in the list, and they should be different. The calendar should always render all the events, to allow the user to navigate

3. In the event page, could we remove "Documents" if there aren't documents attached? It's a bit confussing

apradillap:
Fixed requests @ferblape

furilo:
Please merge and deploy this

PR #1469, opened 2018-02-20
improve performance loops

Crashillo:
## :v: What does this PR do?
Improves strongly DC charts performance. Reduces page load to less than 10s, most of them spent bringing the heavy CSV file.


Crashillo:
Please, deploy to check whether IE issues has gone or not.

PR #1472, opened 2018-02-21
Review gobierto-indicators

### LINKED ISSUE #1467: Responsive review
- [x] Margins, paddings, etc.
- [x] If you have scrolled to the middle of the list or further, and click on an element, scroll position is kept, so if your destination content if shorter than the position you are in, you won't see the content

Crashillo:
## :v: What does this PR do?

- [x] Make tree responsive. Some extra UX fixes. Closes #1467 
- [x] Add permalinks
- [x] Open detail on click the full banner, not just the link
- [ ] GCI Indicators unique index cc/ @apradillap 

## 📷 Screeshots

> Tree:

imagen

> Detail:

imagen


Crashillo:
Hi, @apradillap, to make it work properly the permalinks, it requires an unique ID number from GCI indicators, in other words, do not reset the id counter (as ITA & Infoparticipa)

imagen

Crashillo:
We have a situation here.

In order to create a permalink, we need to set a proper reference. That's an id. ITA & Infoparticipa works like a charm, because `id` is always fulfilled and rendered in the view. GCI isn't, instead.

I've tried a CSS-way to setup dinamically the ids (`counter` prop.), but it allows me not to get the `id` number from JS, and, due to it's a tree, when you enter in a single view, this `id` is reset. Dead end.

The problem with GCI is that the data isn't the same every year. They're filtered, instead. That's means the `id` is "wrong", so the permalink reference is incorrect.

Any light?

apradillap:
Hi, @Crashillo, I added an unique ID number on GCI indicators.

Remember:

1. Update populate-data-indicators
2. From gobierto, execute:

`bin/rails runner .../import_gci.rb your_local_domain`

captura de pantalla 2018-02-27 a las 17 19 39

PR #1491, opened 2018-03-05
Improve performance plans json

### LINKED ISSUE #1487: Improve performance of plans JSON
http://esplugues.gobify.net/planes/pam/2016.json takes 4 secons in the server, and it should take less than 500 ms.

Doing a quick analysis, it performs +2500 queries to the database. Maybe that's a good point to start.

apradillap:
Closes #1487 

## :v: What does this PR do?

Improved plan load performance. Previously the load of the plans was greater than 3 seconds, now about 1 second. It launched many queries of the calculations of the progress and the uid.

They've been cached in a database: uid and progress.

Go to a plan page

## :eyes: Screenshots
No

## :shipit: Populate-data-indicators

A couple of tasks have been added that are displayed when importing the PAM and that need to be executed. Example:

rake gobierto_plans:category:progress_cache["pam-2016-2019"]
rake gobierto_plans:category:uid_cache["pam-2016-2019"]


ferblape:
👍 much better now

PR #1499, opened 2018-03-07
Hound environment setup

Crashillo:
## :v: What does this PR do?
In order to make Hound work on gobierto, setup all files

### Linters
- [x] SCSS
> Not all rules has been fixed. Some are still there to proceed later by Hound
- [ ] Javascripts
> Only module admin
- [ ] Ruby


ferblape (approved):
👍

PR #1567, opened 2018-04-04
Fix I18n fallback declaration

ferblape:
Closes #301

## :v: What does this PR do?

This PR fixes how I18n fallbacks are declared in the new I18n gem version.

A test has been added to check the application is properly configured.


ferblape:
Discarded

PR #1582, opened 2018-04-10
Force JSON url to have a format parameter

### LINKED ISSUE #1578: Plans / JSON response when browsing back to a plan
In Chrome (Mac+Windows) if you browse to a plan: https://portalobert.esplugues.cat/planes/pam/2016 - then browse to any link, and then hit back in the browser, you get the JSON response instead of the web page.

ferblape:
Closes #1578 

## :v: What does this PR do?

This PR adds a format attribute to the JSON to avoid the browser cache both urls as if they were different.

1. Go to the plan
2. Visit another link
3. Click back button


furilo:
works !

PR #1607, opened 2018-04-20
1602 bubbles generation

### LINKED ISSUE #1602: Different bubble results depending on where the code is executed
There are two ways of load the bubbles:

- execute the Rake task: `bin/rails gobierto_budgets:data:bubbles_sites`
- use the `GobiertoBudgets::BudgetLinesImporter`

Both use the same call to the same class but return different results. Investigate why and fix it.

entantoencuanto:
Closes #1602

## :v: What does this PR do?
Fixes the differences in generation of bubbles depending of the used command. With this changes the `GobiertoBudgets::BudgetLinesImporter` model generates bubbles from existing data instead of existing translations in current locale.

The bubbles generation task uses `GobiertoBudgets::BudgetLinesImporter` which was taking the codes to create the bubbles from translations available in current `I18n.locale`. Creating each bubble, the model changes the locale setting it to `:es` and ending with `:ca`. So if we was generating the bubbles with `GobiertoBudgets::BudgetLinesImporter` without a previously locale set, the bubbles were generated with the translated lines on default locale `:es`. If we was generating the bubbles from the task, except for the first site, the other ones were calling `GobiertoBudgets::BudgetLinesImporter` with :ca: locale.

execute the Rake task: `bin/rails gobierto_budgets:data:bubbles_sites` and use the GobiertoBudgets::BudgetLinesImporter from console and compare the results

### Before this PR
screen shot 2018-04-20 at 17 08 14
screen shot 2018-04-20 at 17 10 04

### After this PR
screen shot 2018-04-20 at 17 24 14

No

### DIFF
--- a/app/models/gobierto_budgets/data/bubbles.rb
+++ b/app/models/gobierto_budgets/data/bubbles.rb
@@ -27,31 +27,29 @@ module GobiertoBudgets
       end
 
       def build_data_file
-        base_conditions = { site: site, kind: GobiertoBudgets::BudgetLine::EXPENSE, area_name: GobiertoBudgets::FunctionalArea.area_name, level: 2 }
-        expense_categories.each do |code, name|
-          fill_data_for(code, name, base_conditions, "expense")
+        expense_lines.group_by(&:code).each do |code, lines|
+          fill_data_for(code, lines, "expense")
         end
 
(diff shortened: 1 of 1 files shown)


ferblape:
👀 !

Issue #292, opened 2017-01-31
Add the text from the home page as a configuration setting


ferblape:
The text in the homepage of Gobierto People module should be editable. Add a settings page in the module configuration section with a textarea for this setting.

PR #1614, opened 2018-04-25
292 - Skip IBM Notes requests to recurrent event instances out of sync range

amiedes:
Closes [#292](https://github.com/PopulateTools/issues/issues/292)

## :v: What does this PR do?

Two of the people we have using IBM Notes configured had several recurrent events that expanded during a really log period (i.e. had lots of event instances).

The initial API request made to retrieve these people events was scoped on the sync range we have, but when retrieving the recurrent event instances we were not filtering anymore so we were doint *lots* of requests that took more than 40 minutes to finish for each person.

I've modified the code so we don't do HTTP request to event instances we already know are out of the sync range.


ferblape (approved):
Let's see how it works

PR #1615, opened 2018-04-25
Ui issues with organizations

entantoencuanto:
## :v: What does this PR do?
Makes some improvements to site personalization:
* Extracts `javascript_hook` on footer layout. In this way the hook can be used in liquid templates.
* Replaces all references to "Ayuntamiento" (except budget lines guide, this part will be redrafted) using `site.organization_name` instead. To do this, some inflections has been defined to use organization_name in context with the grammatical particles correctly gathered. For example, The if the organization name is "universidad", in the expression "Name of the university" the last part should be evaluated in different locales as:
  * `es`: "... de la universidad"
  * `ca`: "... de l'universidad"
  * `en`: "... of the universidad"

  The grammatical gender of the name can be set on each locale using the site attribute "Other configuration variables" and three genders are available: "f" for feminine, "m" for masculine and "n" for neuter. If none is set, neuter is assumed. For example, "Other configuration variables" can be set as:
```yaml
organization_name_gender:
  es: "f"
  ca: "m"
```

Configure a site setting `organization_name` on organization field and set grammatical genders for each locale if you wish and visit budgets path

### Before this PR
screen shot 2018-04-25 at 23 28 03
screen shot 2018-04-25 at 23 28 15

### After this PR
screen shot 2018-04-25 at 23 31 50
screen shot 2018-04-25 at 23 31 35

No

### DIFF
--- a/app/controllers/gobierto_budget_consultations/application_controller.rb
+++ b/app/controllers/gobierto_budget_consultations/application_controller.rb
@@ -7,6 +7,9 @@ module GobiertoBudgetConsultations
 
     before_action { module_enabled!(current_site, "GobiertoBudgetConsultations") }
 
+    def set_current_site
+      @site = SiteDecorator.new(current_site)
+    end
     protected
 
     def check_not_responded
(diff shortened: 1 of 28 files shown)


ferblape (approved):
😁

Issue #315, opened 2017-02-10
CMS Module - first iteration


furilo:
Sandbox: 

- [x] Section index http://madrid.gobierto.dev/sandbox/admin_cms
- [x] Edit page http://madrid.gobierto.dev/sandbox/admin_cms_page_text

(in 268-integrate-design-gobierto-people)

For a further iteration, we would create the menú editor, in which we create a hierarchy to which we assign existing pages.

PR #1622, opened 2018-04-27
315 - Fix Algolia search in GobiertoCms

amiedes:
Closes [#315](https://github.com/PopulateTools/issues/issues/315)

## :v: What does this PR do?

* Includes the search Javascript code in `GobiertoCms` module
* Uses `GobiertoCms` layout when rendering a page from the `MetaWelcomeController`. I think I'm not breaking anything with:

https://github.com/PopulateTools/gobierto/blob/883b614407ab56d5d3333b0f8d5e5339066b9fba/app/controllers/meta_welcome_controller.rb#L33

but please check.

Just go to http://getafe.gobify.net and check the search is working

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,7 @@ class ApplicationController < ActionController::Base
   rescue_from ActionController::UnknownFormat, with: :render_404
 
   helper_method :helpers, :load_current_module_sub_sections, :current_site, :current_module,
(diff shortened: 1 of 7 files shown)


furilo:
- [ ] Not neccesarily related with this PR, but in the past we commented about showing first the results from the CMS module, before the ones from the budgets module, in the autosuggest.

amiedes:
Done! BTW I've reverted the changes in #1619 so we're sure this does not break anything

PR #1629, opened 2018-05-03
Update poltergeist version

ferblape:
Unexpected

## :v: What does this PR do?

This PR updates to the latest Poltergeist version.

Just check the tests are green.


ferblape:
@amiedes could you review the tests that fail? They should be easy to fix.

PR #1632, opened 2018-05-03
1449 - Show regular GobiertoBudgets search when Algoliasearch is not configured

### LINKED ISSUE #1449: Fallback from Algolia search
Algolia account shouldn't be mandatory in the budgets module.

When Algolia is not present, use the ES search implemented in the budgets explorer.

amiedes:
Closes #1449 

To test in staging you'll have to empty one of the algolia env variables and restart the server. In development it's easier to test then:

1. Comment any of the algolia env variables
2. If you go to a page in the budgets module, you should be shown the budget line search
3. If you go to a page in any other module, the search box should be hidden


ferblape:
I'm getting this error in a site without Algolia configured:

`uncaught exception: AlgoliaSearchError: Please provide an application ID. Usage: algoliasearch(applicationID, apiKey, opts)`

I think you forgot to check in the JS block that configures algolia in the layout.

PR #1645, opened 2018-05-11
Call to next filter if current one has no data

Crashillo:
## :v: What does this PR do?
Automatically call to the next filter if the last one, currently *last 3 months* filter, has no data. If so, an alert is prompted.


ferblape:
Why the `alert`?

app/javascript/budgets/modules/invoices_controller.js
```diff
@@ -67,6 +67,20 @@ window.GobiertoBudgets.InvoicesController = (function() {
...
         data = _.filter(csv, _callback(filter));
 
+        if (!data.length) {
+          // disable empty filter
+          $('#invoices-filters button[data-toggle=' + filter + ']').prop('disabled', true)
+
+          // if there's no data, get all available filters and trigger a new one
+          let filters = [];
+          $('#invoices-filters button[data-toggle]').each(function() { filters.push($(this).attr('data-toggle')) });
+          let previousFilter = (filters.indexOf(filter) > 0) ? filters[filters.indexOf(filter) - 1] : alert('Any filter has data.')
```
ferblape: You should use I18n here in the alert message.

ferblape (changes requested):

app/javascript/budgets/modules/invoices_controller.js
Crashillo: Oh! You're right! I forgot it, look, even I created the translations:
https://github.com/PopulateTools/gobierto/pull/1645/files#diff-c1ce275193fd9589720bb8618e801eb2R498

ferblape:
To me this is ready to release

furilo:
go

PR #1650, opened 2018-05-16
Cleaner validations for translated attributes. Abstract form objects …

### LINKED ISSUE #1538: Differentiate compulsory / optional fields in fields with dynamic validations
Related with #960 and #1513

Differentiate compulsory / optional fields in fields where the validations are not standar AR validations but dynamic validations defined in the model.

There are just a few missing models, i.e Events.

amiedes:
## :v: What does this PR do?

Closes #1538 - but also does some overall improvements:

* Creates `TranslatedAttributePresenceValidator` to check for attribute presence at FormObject level. The standard Rails presence validator did not work because an empty JSON field (i.e. `{}`) is always present.

  Usage:

  ```ruby
  validates :title_translations, translated_attribute_presence: true
  ```

* Creates `TranslatedAttributeLengthValidator` to check for the length of translated attributes. The standard Rails length validator did not work because it didn't check each of the translations sepparately. This validator tells explicitly which locale is too long, because now it happened that we could have the default shown locale with empty text but the error say "X is too long", which is confusing:

  image

  Usage:

  ```ruby
  validates :title_translations, translated_attribute_length: { maximum: 140 }
  ```

* Creates a `attribute_indication_tag` helper to display required and max_length tips in form labels. It supports passing the length value, or a reference to the form builder (and it'll get the length value from the model)

  Usage:

  ```erb
  <%= attribute_indication_tag required: true %>
  <%= attribute_indication_tag max_length: 80 %>
  <%= attribute_indication_tag required: true, max_length: 80 %>
  <%= attribute_indication_tag required: true, max_length: { f: f, attr: :cta_text_translations } %>
  ```

* Creates `BaseForm` class to group common behavior between all form objects
* Adds missing attribute indications for some form labels
* Moves some user-faced validations from models to form level. 

## How to test this

The scope of this issue is a bit broad, but a good place to test is in the new process for stage: http://madrid.gobify.net/admin/participation/processes/27/process_stages/new

No

## Pending work

- [x] Fix missing locales


app/forms/base_form.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class BaseForm
+
+  include ActiveModel::Model
+  include GobiertoCommon::AttributeLengthValidatable
+
+  protected
+
+  def promote_errors(errors_hash)
```
ferblape, line 10: 🎆

ferblape (approved):
Good work

PR #1671, opened 2018-05-29
DC Charts title alignment

Crashillo:
## :v: What does this PR do?
Align texts properly, no matter which browser.

## :eyes: Screenshots
Before
imagen
After
imagen


ferblape:
Ok to me

PR #1723, opened 2018-06-25
Fix url buildable concern

amiedes:


amiedes:
@ferblape do you see any problems in doing this?

ferblape:
Nop, go go go

PR #1728, opened 2018-06-27
Remove some unnecessary columns

ferblape:
Improvement

## :v: What does this PR do?

This PR uses a new option in Populate Data to exclude unnecesary columns from the response dataset and save brandwith, in the case of invoices around the 40%.


ferblape:
FYI @Crashillo

PR #1742, opened 2018-07-03
399 person submodules redirect

entantoencuanto:
Related with #399

## :v: What does this PR do?
Redirects from person show controller to person events, gifts, invitations or trips when the person only have data of one type.

This redirect is only enabled when the site has defined a default date range for gobierto_people module and the redirection is determined based on the available resources for the current date filter. 

Set a date range for site and visit the show page of a person. Play with start_date and end_date params to have only events, gifts, invitations or trips.

No


ferblape (approved):

ferblape:
Works as expected. Merging!

PR #1751, opened 2018-07-04
Better axis length calculation

Crashillo:
## :v: What does this PR do?
Improves the function to estimate how many x-axis there are, based on Dates, in the provided data

### DIFF
--- a/app/javascript/visualizations/modules/punchcard.js
+++ b/app/javascript/visualizations/modules/punchcard.js
@@ -1,4 +1,4 @@
-import { d3 } from 'shared'
+import { d3, moment } from 'shared'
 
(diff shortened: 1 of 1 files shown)


ferblape:
Do you have screenshots of the before / after or an example of URL where I can check?

Crashillo:
It's related with an issue in the review. Current status may be found: http://gencat.gobify.net/grupos-de-interes/unio-sindical-obrera-de-catalunya?start_date=2016-01-08
Former calculation didn't take into account the difference between elements. That is, if your dates are `[January, March, June]` you don't have 3 axis, you have 6, instead (from Jan to Jun)

ferblape:
Looks much better, thanks!

PR #1761, opened 2018-07-06
408 - Microsoft Exchange: request events in batches to avoid timeouts

amiedes:
Closes [#408](https://github.com/PopulateTools/issues/issues/408)

## :v: What does this PR do?

Modifies Microsoft Exchange Integration to request detailed event information in batches to avoid timeouts with the remote server.

In staging you can't due to SSL problems. In local you can configure the conflictive calendars and run the manual sync from the UI to check it completes successfully (I've done it for both of them and everything looks OK).

No


ferblape (approved):
I'm not 100% convinced but let's try

PR #1775, opened 2018-07-13
Plan UI fixes

ferblape:
Improvement

## :v: What does this PR do?

- Removes old ES2015 duplicated code
- Removes this weird animation at level 0:

animation-level0

- Enables click in the arrows and fixes the cursor:

click-arrow

- Adjust columns width to avoid two lines

screen shot 2018-07-13 at 09 39 46

Probably the CSS rules I'm using are not the best, feel free to review them or refactor them in another issue.

### DIFF
--- a/app/assets/stylesheets/module-planification.scss
+++ b/app/assets/stylesheets/module-planification.scss
@@ -146,10 +146,15 @@ $white: $color_main_negative;
       :first-child {
         flex: 0;
       }
+
+      .fa.cursor-pointer {
+        cursor: pointer;
+      }
     }
 
     h3 {
       margin: 0 0 0.5em;
+      cursor: default;
 
       a {
         display: block;
@@ -159,10 +164,10 @@ $white: $color_main_negative;
         &::first-letter {
           text-transform: uppercase;
         }
-      }
 
-      &:hover {
-        opacity: 0.75;
+        &:hover {
+          opacity: 0.75;
+        }
       }
     }
 
--- a/app/assets/stylesheets/module-planification.scss
+++ b/app/assets/stylesheets/module-planification.scss
@@ -139,6 +139,13 @@ $white: $color_main_negative;
       text-align: right;
     }
 
+    .flex-20 {
+      flex-basis: 20%;
+    }
+    .flex-10 {
+      flex-basis: 10%;
+    }
+
     .node-title {
       flex-basis: 65%;
       display: flex;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/module-planification.scss
```diff
@@ -139,17 +139,29 @@ $white: $color_main_negative;
...
+
     .node-title {
       flex-basis: 65%;
       display: flex;
 
       :first-child {
         flex: 0;
       }
+
+      .fa.cursor-pointer {
```
Crashillo: While `.cursor-pointer` is within `.node-title` (nothing else gonna be changed), avoid using `.fa` class, since it belongs to FontAwesome

app/views/gobierto_plans/plan_types/show.html.erb
```diff
@@ -252,12 +252,12 @@
 <!-- node list template -->
 <script type="text/x-template" id="node-list-template">
   <div class="node-list">
-    <div class="node-title">
-      <div><i class="fa" :class="[(isOpen) ? 'fa-caret-down' : 'fa-caret-right']"></i></div>
-      <h3><a @click.stop="setActive" data-turbolinks="false">{{ model.attributes.title | translate }}</a></h3>
+    <div class="node-title" @click.stop="setActive">
+      <div><i class="fa cursor-pointer" :class="[(isOpen) ? 'fa-caret-down' : 'fa-caret-right']"></i></div>
+      <h3><a href="javascript:void(0)">{{ model.attributes.title | translate }}</a></h3>
```
Crashillo: Is this actually necessary? Maybe `href="#"` or removing the attribute?

app/assets/stylesheets/module-planification.scss
```diff
@@ -139,17 +139,29 @@ $white: $color_main_negative;
       text-align: right;
     }
 
+    .flex-20 {
```
Crashillo: This helpers are quite static. And confusing.
1. It you wanna use helper classes, they should be created with a mixin, to be shared, and to create a lot. In this case, for instance, a loop from 0 through 10
2. The name `flex-10` could be refer to `flex: 10;`, but we have `flex-basis`, instead. Rename it to `flex-basis-10`, `f-b-10`, `flex-b-10`... something like

Crashillo (changes requested):

ferblape:
Thanks @Crashillo

PR #1782, opened 2018-07-16
111 - Anonymize poll answers

amiedes:
Closes [111](https://github.com/PopulateTools/issues/issues/111)

## :v: What does this PR do?

Anonymizes poll answers and includes encrypted demographic information (+ custom records) within each answer DB record.

* You can answer a poll from the UI - http://madrid.gobify.net/participacion/p/proceso-imagenes-encuestas/encuestas/38/answers/new
* The results page continues to work correctly (data was migrated): http://al*******s.gobify.net/admin/participation/processes/18/polls/10/answers
* You can check the migrated data in staging with something like:

```ruby
puts JSON.pretty_generate ::GobiertoParticipation::PollAnswer.pluck(:encrypted_meta).map { |meta| SecretAttribute.decrypt(meta) }
```

Nope

### DIFF
--- a/app/forms/gobierto_participation/processes/poll_answer_form.rb
+++ b/app/forms/gobierto_participation/processes/poll_answer_form.rb
@@ -58,14 +58,8 @@ module GobiertoParticipation
 
           next if question.nil? || user.nil? || answer_attrs_wrapper[:answers_attributes].nil?
 
-          answer_attributes = {
-            poll_id: question.poll.id,
-            question_id: question.id,
-            user_id: user.id
-          }
-
           answer_attrs_wrapper[:answers_attributes].each do |_, answer_attrs|
-            @answers_attributes << answer_attributes.merge(
+            @answers_attributes << answer_attributes(question, user).merge(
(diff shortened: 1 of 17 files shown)


ferblape:
I think we need to think it a bit, because with this solution the answer is anonymous but I can obtain the user in less than 1 second:

```ruby
u = User.all.select { |u| SecretAttribute.hmac(u.id) == GobiertoParticipation::PollAnswer.first.user_id_hmac }
=> [#<User id: 26, email: "...", referrer_entity: "">]
```

And if an attacker gets the answers table is quite probable she gets the users table too, and given we use integer and autoincremental ids it's quite easy to do the matching.

I'm thinking if we need to use the non deterministic encryption and store in an auxiliary table if the user has answered or not a poll. Ideas?

ferblape:
After discussing this with Alberto:

- we are implementing this encryption with the same principles that in GobiertoParticipation, but with a new more secure algorithm

- my attack would only succeded if the attacker gets the Rails secret key. In my case it was easy because I was in a Rails console. Without Rails secret key the encryption is secure

PR #1802, opened 2018-07-24
Population pyramid

Crashillo:
## :v: What does this PR do?
It adds an new chart in observatory section, feed from populate-data
- [x] Improved data-reloading behaviour


Crashillo:
Former population chart has been replaced by the new one

ferblape:
Releasing!

PR #1832, opened 2018-08-03
Datepicker issues

### LINKED ISSUE #1811: Events date picker issues
If you go to GobiertoPeople > Person > Agendas and edit an event:

- [ ] ~it shows always the current time + the offset of the time zone. You can check this with any event in staging. If you edit and save the event it'll be recorded with a new time of + 2 hours, and so on~

- [x] when you pick a day, the date picker closes. If you want to edit the hours too you need to open it again

- [x] could we make the input editable so no need to use the date picker? could we make the hours input editable?

- [ ] it'd be great to link (bind) start and end dates for events and not for processes phases. So maybe we could think on a parameter that if active links the dates, otherwise it doesn't

Crashillo:
Closes #1811 

## :v: What does this PR do?
Fixed the following glitchs:
- [ ] ~time saved is offseting from selection~ solved via RnR
- [x] remove autoClose always
- [x] input editable
- [x] binded dates via parameter
- [x] update 2nd datepicker (end date) when date is before 1st datepicker (start date)


ferblape:
👎  I'm editing an event from an agenda and I can't see dates are binded

ferblape (changes requested):
I'm creating a new event in an agenda and dates are not binded

Crashillo:
I was checking the behaviours on staging, let me know your thoughts. These are the casuistries I applied:

1. Click on `start_date` datepicker
     1. If `start_date > end_date`, automatically `end_date` displays `start_date + 1h` (This is implemented)
     2. If not (`start_date < end_date`), then do nothing. You may want to have a longer event. 
2. Click on `end_date` datepicker
     1. If `start_date > end_date`, then do nothing. Should it set as `start_date = end_date - 1h`?
     2. If not, neither do anything. You may want to have a longer event.

IMHO, I should apply only the feature `2.i`. Otherwise, you won't allow the user set the dates as they want (datepickers will trigger each other)

Crashillo:
Feature applied & deployed

PR #1834, opened 2018-08-03
1808 protect terms and vocabularies with permissions

### LINKED ISSUE #1808: Protect vocabularies and terms controllers to regular users
This is a follow-up of #1806, and depends on #1799

> The "Vocabularies" tabs is hidden when there are no permissions, but protecting controller actions with filters dependes on pull #1799. Also the integration tests for this case are pending.

entantoencuanto:
Closes #1808

## :v: What does this PR do?
Includes a filter to check if an admin has permissions to edit vocabularies and terms and redirects him in case it doesn't.

Update an admin to be regular and don't check the Vocabularies checkbox. An admin with this profile shouldn't be available to reach `/admin/vocabularies` path.

No


ferblape (approved):
👍

PR #1846, opened 2018-08-08
Plan enhancements

### LINKED ISSUE #1830: Add permalinks to plans
- each time the user visits a node or a category the url contains a reference using a hash `#...`
- url can be shared. A url with a reference to a node or category will open the plan explorer showing that item
- the back button should work

### LINKED ISSUE #1831: Styles issues
When the mouse is over a project the cursor should be a pointer instead of `text`

screen shot 2018-08-03 at 11 15 22

Crashillo:
Closes #1830 & Closes #1831 

## :v: What does this PR do?
- Includes a set of permalinks features to allow the user share an active. (#1830)
- Set cursor to point on table view (#1831)
- Code clarification: Split into partials, comment code, update eslint rules


ferblape (changes requested):
Two things:

- permalinks are ugly, could we think in different things, like the UUID?
- browser back button doesn't work

Crashillo:
@ferblape I was doing some research&test about the UUIDs, and I didn't find any easy way that worths it. 
First off, they require an extra library to work with (I think it's excessive add a library just for renaming url's).
Moreover, the feature has to be directional (encode/decode). Currently, it's using the Base64 encoded `uid` field. This `uid` is the real field to make it work all the permalinks. If we decided to use another field, we need to get the `uid` anyways, so we must parse that url string, look for it in the JSON tree, get the uid, and apply it.

I'll suggest to keep the `uid` in the url for easier processing purposes. No matter if you append something, encoded or not, to the final url, but keeping the `uid` there in some way.

Crashillo:
Requests satisfied & deployed

PR #1854, opened 2018-08-13
Fix subscriptions when referer is not available

amiedes:
Fixes rollbar [#1202](https://rollbar.com/Populate/gobierto/items/1202/occurrences/49905012625/)

## What does this PR do

When creating a new subscription, if the referer is not available it redirects to the home page.

### DIFF
--- a/app/controllers/meta_welcome_controller.rb
+++ b/app/controllers/meta_welcome_controller.rb
@@ -7,7 +7,7 @@ class MetaWelcomeController < ApplicationController
     render_404 and return if current_site.nil?
 
     if current_site.configuration.home_page != "GobiertoCms"
-      redirect_to eval("#{current_site.configuration.home_page.underscore}_root_path")
+      redirect_to(current_site.root_path)
     else
       # This code allows the homepage to render a page without redirecting to the page url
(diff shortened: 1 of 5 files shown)


app/controllers/user/subscriptions_controller.rb
```diff
@@ -50,8 +50,7 @@ def create
             sign_in_path: new_user_sessions_path(host: current_site.domain)
           )
         end
-
-        redirect_to request.referrer
+        redirect_to(request.referrer || root_path)
```
ferblape: There's a new way to do this!

https://blog.bigbinary.com/2016/02/29/rails-5-improves-redirect_to_back-with-redirect-back.html

app/controllers/meta_welcome_controller.rb
```diff
@@ -7,7 +7,7 @@ def index
     render_404 and return if current_site.nil?
 
     if current_site.configuration.home_page != "GobiertoCms"
-      redirect_to eval("#{current_site.configuration.home_page.underscore}_root_path")
+      redirect_to(current_site.root_path)
```
ferblape, line 10: 👍

ferblape (changes requested):

amiedes:
Done 💅

PR #1858, opened 2018-08-16
1783 edit plan nodes jsgrid

### LINKED ISSUE #1773: Edit a plan
Allow the editing of a plan content. We keep tab "Data" as the second one.

## Configuration tab

- [ ] At the end of the page we'll put a section with three input texts: 

```
h3. Definition of level names

Level 1
Level 2
Level 3
```

## Data tab

- First row (as in http://esplugues.gobify.net/admin/gobierto_calendars/events?collection_id=209) with: 
  - Import CSV - Export CSV  
- Main column (we won't put sidebar in this page, as all editing will be Ajax so no need to Update the whole plan):
  - A JS Grid table (we are already using it for Invoices in the front)
  - Columns:
    - Level 1 
    - Level 2 
    - Level 3 (these three will be selects, based on the values defined in the Conf tab, or the defaults)
    - Project
    - Progress
    - Status
    - Start Date
    - End Date
    - Custom [it should open a textarea]

Each field should have its corresponding field type (text, select, date...)

Something to discuss: The Status field is usually a set of fixed values, that may vary from plan to plan. But it would be good to be able to offer a select instead of an input, so we avoid errors. Ideas? 

## New page: Import CSV

The current page for Data would be converted to this new "Import CSV". Before the INPUT, we'd add: 

```
<p><strong>Warning</strong> If you upload a CSV, all actual content will be deleted, so any change you might have done through the Gobierto Web UI will be lost.</p>
```

entantoencuanto:
Closes #1773

Edit a Plan from admin, update plan nodes

## :eyes: Screenshots

https://www.dropbox.com/s/8j45vw3wtczq5fv/plan_jsgrid.mov?dl=0

No


ferblape:
A bit related with this issue, the exported file doesn't contain the '.csv' extension I'd expect. It could be confusing to users.

ferblape (changes requested):
Two small changes to close this:

- [x] A bit related with this issue, the exported file doesn't contain the '.csv' extension I'd expect. It could be confusing to users.
- [x] use the corresponding icons in the import / export options
- [x] when I add an invalid value in the options json field, no error is displayed and the user gets stuck

ferblape:
Now it's working well, I'm merging it.

PR #1892, opened 2018-09-12
Remove misleading rescue

amiedes:
This rescue is not useful because `ActionController::RoutingError` errors are raised before any controller gets instantiated: https://stackoverflow.com/a/37174557/4156152

We don't see the rollbars because they're being ignored in https://github.com/PopulateTools/gobierto/blob/master/config/initializers/rollbar.rb#L11.

You can test this in local - If in `master` you visit something like http://madrid.gobierto.test/a/b/c/d/e/f/g/h/i/j/k the exception is not being rescued.


ferblape (approved):

amiedes:
DO NOT MERGE

amiedes:
Fixed

PR #1913, opened 2018-09-25
1837 - Unify timezones

### LINKED ISSUE #1837: Adjust EventForm#starts_at timezone
(Related with #1811 first point)

Our local environment and the servers run in different timezone. When using `#to_time` these differences change the resulting time:

```
# local (our laptops are in Madrid time zone), so the new time is in Madrid time zone
 "2018-08-04 18:00".to_time
=> 2018-08-04 18:00:00 +0200

# server is in UTC, so the new time is in UTC time zone
"2018-08-04 18:00".to_time(:local)
=> 2018-08-04 18:00:00 +0000
```

We should ensure we are working always in UTC timezone.

This is causing #1811 first point.

amiedes:
Closes #1837 

## :v: What does this PR do?

Modifies the way the start end dates are passed to the datepickers. Now the value is set to the number of microseconds since epoch, since it's the only way it's seems we can pass the timezone to the datepickers.

## :mag: How should this be manually tested? (WIP)

Ensure the following forms now work correctly:

- [x] GobiertoCalendars -> create event
- [x] GobiertoCms -> create page
- [x] GobiertoPeople -> create person statement
- [x] GobiertoParticipation -> process
- [x] GobiertoParticipation -> process stage
- [x] GobiertoParticipation -> poll
- [x] GobiertoParticipation -> contribution container

No


ferblape (changes requested):
All forms work as expected except the new CMS page form:

screen shot 2018-09-27 at 16 43 52

:bat:

ferblape (approved):

PR #1920, opened 2018-10-01
Charters design

Crashillo:
## :v: What does this PR do?
Includes an sketch/mockup of the new screens 

Preview: [Home](http://madrid.gobify.net/sandbox/charters_home.html) - [Documento](http://madrid.gobify.net/sandbox/charter.html)

Docs: https://gobierto.readme.io/docs/charts#section-sparkline-context-data-options-

## :eyes: Screenshots
*Home*
imagen

*Documento*
imagen


app/javascript/visualizations/modules/sparkline.js
```diff
@@ -0,0 +1,132 @@
...
+    g.selectAll(".tick:first-of-type line").remove()
+    g.selectAll(".tick:last-of-type line").remove()
+    g.selectAll(".tick line")
+      .attr("y1", 0)
+      .attr("y2", -this.height)
+  }
+
+  _getColor() {
+    if (this.isPositive && this.trend === 'up' || !this.isPositive && this.trend === 'down') {
+      return '#AAC44B';
```
ferblape, line 92: Shouldn't we move the colors to the options of the sparkline to make it more generic?

Crashillo: Agree. That was what the old charts did. I just turned into a class

ferblape:
Question: is the widgets section width 100% on purpose? We don't usually do that in other Gobiert modules, the width is fixed to a maxium.

screen shot 2018-10-08 at 09 39 47

furilo:
@ferblape we are doing some tests

Crashillo:
Yeah, it is. Even though it's not definitive.

PR #1929, opened 2018-10-05
441 custom fields

entantoencuanto:
Closes https://github.com/PopulateTools/issues/issues/441

## :v: What does this PR do?

Implements a new feature based on custom user fields to define and manage custom attributes on different models of the application. The models must have this functionality enabled to generate custom fields for them

Go to admin as manager and visit http://gobify.net/admin/custom_fields/module_resources/gobierto_people-person. Currently this option is only available for `GobiertoPeople::Person` but can be used on other models by adding on the module configuration:

on `app/models/gobierto_people.rb`:
```ruby
  def self.classes_with_custom_fields
    [GobiertoPeople::Person]
  end
```

## :eyes: Screenshots
https://www.dropbox.com/s/3sjm4vqkgfpdxuo/custom_fields.mov?dl=0

No


ferblape:
When I create a field with options there are a couple of issues:

1. Javascript is not being evaluated (in the video it doesn't happen):

custom fields options

2. in the markup, between the create and cancel links there's a small underline

screen shot 2018-10-08 at 09 48 36

Are you planning to integrate this with the Person form?

app/models/gobierto_common/custom_field.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def localized_options(locale)
+      options.map do |id, translations|
+        [translations[locale.to_s], id]
+      end
+    end
+
+    private
+
+    def set_uid
```
ferblape, line 44: `self.uid ||= SecureRandom.uuid`

app/models/gobierto_common/custom_field.rb
```diff
@@ -0,0 +1,56 @@
...
+      return if uid.present?
+
+      self.uid = SecureRandom.uuid
+    end
+
+    def set_position
+      self.position = begin
+        self.class.where(site_id: site_id).maximum(:position).to_i + 1
```
ferblape: Shouldn't the position be scoped by `site_id` and `class_name` too?

app/models/gobierto_common/custom_field_value/localized_text.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module GobiertoCommon::CustomFieldValue
+  class LocalizedText < Base
+    def value
```
ferblape, line 5: I don't understand what does `super` in this method

app/controllers/gobierto_admin/gobierto_common/custom_fields/custom_fields_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+    module CustomFields
+      class CustomFieldsController < BaseController
+        before_action :check_permissions!
+        before_action :check_class, except: :create_option
+
+        def index
+          set_class_name
+
+          @class_human_name = @class_name.constantize.model_name.human(count: 2)
+          @custom_fields = ::GobiertoCommon::CustomField.where(class_name: @class_name)
```
ferblape: Shouldn't this be filtered by `current_site`? I can see custom fields from other sites in this page.

ferblape (changes requested):

app/models/gobierto_common/custom_field_value/localized_text.rb
entantoencuanto: Yes, its useless

app/models/gobierto_common/custom_field.rb
entantoencuanto: I used this because the migration sets a default value for uid as "", so the result of `self.uid ||= ...` is "". I'll add a migration to remove the default value for uid.

PR #1951, opened 2018-10-11
Don't assign vocabulary_id when it's not present

amiedes:
Avoid assigning `0` to `site.gobierto_people_settings. political_groups_vocabulary_id ` when no vocabulary is selected.

Linked to https://github.com/PopulateTools/gobierto-gencat-engine/pull/9

~This just makes sure than when we save the site we don't assing an ID to the vocabulary, and updates the tests so when no political groups exists, the welcome page still works. But as you said maybe we also want to enforce this vocabulary exists for all the sites that require it (all that don't have this engine configured?).~


ferblape:
@entantoencuanto please review this PR

ferblape:
And merge it if it's ok ;)

PR #1979, opened 2018-10-26
464 - Support for custom favicon

amiedes:
Closes [#464](https://github.com/PopulateTools/issues/issues/464).

## :v: What does this PR do?

Creates setting for custom favicon inside raw_configuration_variables (http://madrid.gobify.net/admin/sites/2/edit).

Check in http://madrid.gobify.net

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,3 +20,10 @@ Closes #N
 - [ ] new entry in `config/secrets.yml`?
 
 (Changes in these files might need to update the role in Ansible)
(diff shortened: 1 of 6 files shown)


furilo:
Have we added documentation for this? 
Is there documentation for similar settings?

amiedes:
Docs added in https://gobierto.readme.io/docs/administraci%C3%B3n-de-un-sitio#section-crear-y-personalizar-un-sitio. The left tabulation for the nested section looks a bit weird but I've been struggling to set it properly. It's caused because the example conf is tabbed one level to the right to align with the upper text.

PR #1984, opened 2018-10-29
Control errors when no plans available

### LINKED ISSUE #1937: Review GobiertoPlans empty state
When all the existing plans are drafted (or none exist), `/plans` breaks.

Related: https://rollbar.com/Populate/gobierto/items/1269/

ferblape:
Closes #1937

## :v: What does this PR do?

This PR refactors plans controller code to avoid a 500 error when no plans where available and to make it more robust.

Also adds a test to this scenario.

Go to a site with plans and unpublish all of them. You should get a 404 error and not a 500.

### DIFF
--- a/app/controllers/gobierto_plans/plan_types_controller.rb
+++ b/app/controllers/gobierto_plans/plan_types_controller.rb
@@ -5,16 +5,16 @@ module GobiertoPlans
     include ::PreviewTokenHelper
     include User::SessionHelper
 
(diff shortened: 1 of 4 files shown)


amiedes (changes requested):
Ok! Although not specified in the issue, we should also control that draft plans are not visible without the preview token (ej: http://madrid.gobify.net/planes/planes-de-4-niveles/2019).

PR #2010, opened 2018-11-12
473 avoid collections with archived containers

entantoencuanto:
Closes https://github.com/PopulateTools/issues/issues/473

## :v: What does this PR do?
* Exclude collections belonging to archived or not present resources in index of CMS, Calendars and Documents
* Redirects to index of corresponding CMS, Calendars or Documents trying to access to a collection belonging to a not available resource
* Refactors some methods like `pages_in_collections`, `attachments_in_collections`, etc. under collectionable concern `in_collections`, optimizing the query and using scopes
 
Visit admin and add documents, pages and events to a process. Review each index, and archive process .The documents, pages and events shouldn't appear and also the links to collections containing them shouldn't be available

No

No


ferblape:
Solves https://rollbar.com/Populate/gobierto/items/1324/?item_page=0&item_count=100&#instances

PR #2016, opened 2018-11-13
Parallel tests

ferblape:
## :v: What does this PR do?

This PR enables parallel tests in CircleCI with 3 workers.


ferblape:
Ping @amiedes @entantoencuanto 

Now tests run 2x faster. Just take care because the UI now has tabs (as many as workers) to check the errors.

PR #2032, opened 2018-11-20
Floating translations

Crashillo:
## :v: What does this PR do?
During scrolling, turn the locale switcher into a sticky element if there are enough I18n items.

## :eyes: Screenshots
Note that the first switcher does nothing, because there are no enough items


ferblape:
I have tested it in http://valencia.gobify.net/admin/citizens_charters/charters/6/edit and 👍

furilo:
Why change the position, and not just stick the component as it is?

Crashillo:
That was a mutual agreement with @entantoencuanto doing an initial proof of concept, otherwise the switcher overlapped the textareas, preventing you a confortable typing or reading what you had wrote down before

furilo:
How about something like

image

Crashillo:
I've updated code and screenshot

PR #2033, opened 2018-11-20
Small participation issues

ferblape:
Fixes many things I've found in Participations module

## :v: What does this PR do?

- Translation: "Su cuenta" > "Tu cuenta"
- In the welcome email, the link to the notifications was not linking to the notifications management page
- I have reused a lot of partials that were duplicated and removed a few i18n keys
- I have fixed the links to the issue that didn't include the issue slug, but the issue ID
- I have fixed two JS errors on contributions containers that avoided an user to use the tool
- I have added messages all around when sections where empty like "There aren't polls published yet" and similar

Test the changes described above.

No

No


furilo:
@entantoencuanto can you review this so we can merge?

entantoencuanto (changes requested):
All the changes are OK, but there are 2 remaining paths where the id is used instead of slug:

https://github.com/PopulateTools/gobierto/blob/498fe3de0dfbde01ca6804b4b08793f52b457bfa/app/views/gobierto_participation/scopes/show.html.erb#L9-L14

entantoencuanto (approved):

PR #2040, opened 2018-11-23
Bug yarn.lock keeping dependencies

Crashillo:
## :v: What does this PR do?
Test from new yarn-lock setup. Keep d3v4 dependencies


Crashillo:
Solved following Mojave steps from https://github.com/nodejs/node-gyp/issues/569

PR #2049, opened 2018-11-26
Add admin site activities for GobiertoCitizensCharters module

entantoencuanto:
## :v: What does this PR do?
* Creates an extension to track changes in models allowing to set the payload subject and group changes of a set of attributes under a common event. In this way the number of events and activities generated can be reduced. Changes in draft objects do not generate activities.
* Configures activities for creation, edition and publication of services and charters and activities for creation, edition, deletion and publication (in the case of commitments) of editions and commitments under the charter.
* Adds integration tests for admin management of services, charters and commitments and tests for subscriber

Go to admin and generate activities for module

## :eyes: Screenshots

site_activities_charters

No

No

### DIFF
--- a/app/controllers/gobierto_admin/gobierto_citizens_charters/api/editions_controller.rb
+++ b/app/controllers/gobierto_admin/gobierto_citizens_charters/api/editions_controller.rb
@@ -22,9 +22,9 @@ module GobiertoAdmin
 
         def destroy
           find_charter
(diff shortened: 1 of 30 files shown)


ferblape:
👍 I've been updating charters, adding new measures and new commitments and the activities and are created as I'd expect:

screen shot 2018-11-27 at 09 53 56

PR #2134, opened 2019-02-13
[WIP] Static ODS page

furilo:
Closes unexpected

## :v: What does this PR do?

Provides the front-end markup for a static ODS page for Budgets.


ferblape:
- Columns 6, 12 and 14 should be removed from the data table
- Once the code is ready we should run it on https://babeljs.io/repl to convert it to ES5 and make it compatible to all browsers

Issue #2152, opened 2019-03-11
Groups and permissions 


furilo:
- [x] Front-end
- [ ] Implementation
  - [ ] TBD
  - [ ] ...

ferblape:
This can be closed and continue in #2245

PR #2174, opened 2019-03-21
Groups and permissions vocabularies

entantoencuanto:
Related with #2152

## :v: What does this PR do?

* Changes index of vocabularies to include the table headers present in sandbox 
* Integrates index of terms of a vocabulary with edition and creation of new terms
* Reordering of elements is pending

The branch of this PR is built on top of https://github.com/PopulateTools/gobierto/tree/2152_groups_and_permissions_admin and merges https://github.com/PopulateTools/gobierto/tree/2152-groups-and-permissions. After merging another PR the branch should be rebuilt
### Before this PR
2152_voc_before

### After this PR
2152_voc_after

No

No


furilo:
When adding/editing a vocab term, can we indent with "--" the children? So instead a flat list:

image

We'd have something like: 

Parent
-- Children
---- Children
etc.

entantoencuanto:
I'll add the indentation

> When adding/editing a vocab term, can we indent with "--" the children? So instead a flat list:
> 
> image
> 
> We'd have something like:
> 
> Parent
> -- Children
> ---- Children
> etc.

Yes, I'll add it

PR #2180, opened 2019-03-25
Configure rollbar js

amiedes:
Closes [#623](https://github.com/PopulateTools/issues/issues/623)

## :v: What does this PR do?

Enables Rollbar for reporting JS errors.


config/initializers/rollbar.rb
```diff
@@ -4,10 +4,14 @@
 
 Rollbar.configure do |config|
   config.access_token = Rails.application.secrets.rollbar_access_token
-
   config.enabled = false if Rails.env.development? || Rails.env.test?
-
   config.exception_level_filters.merge!("ActionController::InvalidCrossOriginRequest" => "ignore",
                                         "ActionController::RoutingError" => "ignore",
                                         "ActionController::UnknownFormat" => "ignore")
+  config.js_enabled = true
```
ferblape: It should be only enabled in staging or production, don't you think?

amiedes: Right. I've fixed it.

ferblape:
Is the ansible role ready? I can't see any reference of the new api key configured in the server.

amiedes:
> Is the ansible role ready? I can't see any reference of the new api key configured in the server.

Yes, but it hadn't been executed on production yet, I just did it.

ferblape:
Ok, thanks. I'm going to merge this, Rollbar only delays 2ms the load of the page, it's working pretty fine, and we have found a few errors, so it's helpful too.

Issue #2182, opened 2019-03-25
Admin Plans / Moderation (projects list & new save widget)


furilo:
To do after #2181 

- To define: 
  - Confirm how to indicate default setting for items? On a per Plan basis, in Plan Conf: setting Moderation flow: Yes | No
  - Permissions - Permissions that should be available to assign to a group
    - Plan Admin: can create and admin plan (Conf tab)
    - Plan Moderator: can moderate projects
    - Plan Editor: can send and edit their plans

- In project list: 
  - [ ] Filter:
    - All
    - Not sent
    - Sent
    - In revision
    - Approved
    - Discarded
- In Project - Save Widget
  - [ ] Show moderation status
  - [ ] Allow to send for moderation
  - [ ] Control Publication action based on Moderation status

- For Plan Moderator: 
  - [ ] Widget moderation ([markup in project page](http://madrid.gobierto.test:3000/sandbox/admin_plans_plan_projects_project)) 
![image](https://user-images.githubusercontent.com/24740/54918704-41dadc00-4eff-11e9-95de-8c63ae168c2e.png)

entantoencuanto:
Pending tasks:
- [x] Role management in admin groups. Allow selection of manage, edit and moderate in modules
- [x] Add author to projects. A project can be created by an admin with edit permission
- Permissions of admins (#2209):
  - [ ] Allow projects tab only to admins with edit or manage roles.
  - [ ] Edit admins can create projects and delete their own projects.
  - [ ] Project edition form: Available for editors and moderators. Admins with only moderate role have disabled all edition fields except change moderation status. When they save the project only moderation status (and maybe publication) is updated.
- [ ] Remove data tab of plans (#2210)
- [ ] When a plan is imported from CSV their projects must be created as published.(#2211)
- [ ] If a moderator updates status of a project from approved to other status, ask if he wants to update publish status to unpublished.(#2208)
- [ ] Add filters to projects tab: Moderation status, edition (my projects, projects I can edit...) and author.(#2212)
- [ ] Display in Plan tab the plan vocabulary decorated with accumulated status and projects (#2213 )

PR #2191, opened 2019-03-26
2182 projects list

entantoencuanto:
Related to #2182

## :v: What does this PR do?

* Adds controller and view to list and filter projects (nodes) of a plan
* Defines a form object with all available filters and scopes in `GobiertoPlans::Node` model 
* Implements delete of projects action in list

Pending:
* Project edition form

Visit a plan and click on Projects tab

### Before this PR
∅

No

No


app/views/shared/decorators/gobierto_plans/category_term_decorator/_header.html.erb
```diff
@@ -0,0 +1,3 @@
+<% %w(Items Progreso).each do |element| %>
```
ferblape: Should this be translated?

app/controllers/gobierto_admin/gobierto_plans/projects_controller.rb
```diff
@@ -0,0 +1,220 @@
...
+        end
+      end
+
+      def destroy
+        find_plan
+        set_filters
+
+        @project = @plan.nodes.find params[:id]
+
+        raise_action_not_allowed unless current_admin_actions.include? :destroy
```
ferblape: You could move this line to the first line in the method to avoid unnecessary queries.

ferblape: Or to a a before_action

PR #2199, opened 2019-03-29
Create new save widget

### LINKED ISSUE #2181: Admin Plans / New Save Widget - without Moderation
Issues to implement new save widget. I leave for a second iteration all that has to do with moderation. 

http://madrid.gobierto.test:3000/sandbox/admin_plans_plan_projects_project

image

- [ ] Buttons save-preview
- [ ] "Editing version N"
- [ ] Show past versions -> ¿just load version in UI?
- [ ] Status
- [ ] Published version
- [ ] Publish button
- [ ] Once a version it's published: the main button should be "Unpublish" 

We leave for further iterations: 

- Minor change (doesn't save version)
- Change published version (fine grain control)

entantoencuanto:
Closes #2181

## :v: What does this PR do?
* Creates a save widget to manage from admin the visibility and versions and inform about moderation status of resources in their edition forms.
* The methods used by the widget relative to version are included in a decorator. Version info and select aren't shown if the resource doesn't have paper_trail enabled.
* The visibility status and version methods are encapsulated in other decorator, which is prepared to implement the moderation flow described in #2182. At the moment, the default flow is to allow any user to publish/unpublish a resource from the widget, since the moderation status is always approved.
* The new widget expects a resource and a form builder. Te resource will be used to initialize the different decorators.
* The preview button is not implemented yet, I'd like to have a little more definition about this. It's a link to front with the changes previously saved?

Replace, for example, the current widget from a form partial:

```ruby
<%= render partial: 'gobierto_admin/shared/save_widget', locals: { f: f, levels: @project_visibility_levels } %>
```

with:
```ruby
<%= render partial: "gobierto_admin/shared/admin_widget_save_v2", locals: { f: f, resource: @project_form.persisted? ? @project : @project_form.project } %>
```

This can be tested better combined with the changes introduced in the PR for the project of a plan

### Before this PR

Screen Shot 2019-03-29 at 17 56 16

### After this PR

Screen Shot 2019-03-29 at 17 56 54

No

No

### DIFF
--- a/app/decorators/gobierto_admin/moderated_resource_decorator.rb
+++ b/app/decorators/gobierto_admin/moderated_resource_decorator.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+module GobiertoAdmin
+  class ModeratedResourceDecorator < BaseDecorator
+    def initialize(resource)
+      @object = resource
+    end
+
+    def has_publication_status?
+      @has_publication_status ||= respond_to?(:visibility_level)
+    end
+
+    def publication_status
+      @publication_status ||= publicable? && visibility_level == unpublished_value ? :not_published : :approved
+    end
+
+    def published?
+      @published ||= publication_status == :approved
+    end
+
+    # This method will change with moderation concern
+    def publicable?
+      @publicable ||= true
+    end
+
+    # This method will change with moderation concern
+    def sent?
+      @sent ||= false
+    end
+
+    def step
+      @step ||= if publicable?
+                  published? ? :published : :publicable
+                elsif new_record?
+                  :new
+                elsif sent?
+                  :sent
+                else
+                  :pending
+                end
+    end
+
+    def locked?
+      @locked ||= [:new, :pending].include?(step)
+    end
+
+    def step_action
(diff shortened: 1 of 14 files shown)


ferblape:
There's something in the behaviour I'd change. Taking this project as an example: http://burjassot.gobify.net/admin/plans/2/projects/5792/edit

The v5 of the project is published. If I update the progress to 98% and I click "Save" I'd expect that:

- the version I'm editing now is v6
- but the published version **stays in v5**
- if I click Publish, then the published version will be v6

> The preview button is not implemented yet, I'd like to have a little more definition about this. It's a link to front with the changes previously saved?

With the flow I have described before the preview button is very easy to implement, because it will show the last version of the item (which can be different from the published version).

@furilo could you confirm?

furilo:
+1 to @ferblape comment

Preview: it's just the current URL with token we have for admins when the item is in draft.

furilo:
Another thing I'd add: asking for confirmation when clicking Publish/Unpublish

entantoencuanto:
The moderation model and concern has been added. Now the widget can work with an instance both including the moderation concern or not

PR #2239, opened 2019-04-18
2228 - Use updated forecast data

### LINKED ISSUE #2228: Revisit how updated forecast data is represented
In Gobierto sometimes the entities provide updated data of the forecast budgets. These data is very important because is the official amount, and it's closer to the executed amount. Internally the data is represented in an index with the same structure as the forecast data.

We need to update the module to introduce this information to make it more relevant:

- budgets home
  - cards: total amount and total amount per inhabitant should use the updated data if available
  - bubbles should be calculated with updated data if available
  - treemap and budgets explorer should use updated data
- budget line detail
  - cards should show the update budget line figure as the main one and the initial amount in small size
  - the rest of figured should use the updated amount too
- execution page
  - the main figures of income and expense should include also the initial amount
  - replace "real income" by "executed income". Same for expenses

amiedes:
Fixes #2228 

## What does this PR do

Updates all the places in the budgets module UI to use updated data if available.

Also, the method to generate the bubbles was updated in these commits:

- https://github.com/PopulateTools/gobierto_data/commit/30b3752b8c64fea5f48a807e38a08f63ea41ab16
- https://github.com/PopulateTools/gobierto_data/commit/5ee03dcdae25f2d82fd5391c4b77c03534345300

## How to review

The easiest way to review this is comparing our budgets production sites:

- http://santfeliu.gobify.net/presupuestos/resumen/2018
- http://esplugues.gobify.net/presupuestos/resumen/2018
- http://mataro.gobify.net/presupuestos/resumen/2019

with the staging ones:

- https://pressupost.santfeliu.cat/presupuestos/elaboracion
- https://portalobert.esplugues.cat/presupuestos/resumen/2018
- https://pressupost.mataro.cat/presupuestos/resumen/2019

Some things that may raise your attention when reviewing:

- The execution percentage hasn't changed, since it was already using the updated data.
- The income execution percentage for Esplugues (check the [mega bar](http://esplugues.gobify.net/presupuestos/resumen/2018)) is very low compared to production. This is not an issue with this PR, this data got corrupted some weeks ago in staging during an accidental import. If in dev you point to production ES it looks normal.


.circleci/config.yml
```diff
@@ -91,9 +91,6 @@ jobs:
       # Copy database config
       - run: cp config/database.yml.example config/database.yml
 
-      # Setup the environment
-      - run: cp .env.example .env
```
amiedes, line 95: I removed this because this file is not read from the Ruby docker image, so it was confusing.

test/integration/gobierto_budgets/elaboration_test.rb
```diff
@@ -46,14 +46,14 @@ def test_menu_subsections
   end
 
   def test_metric_boxes
-    with_current_site(placed_site) do
+    with(site: placed_site, js: true) do
```
amiedes, line 49: @ferblape I did this to avoid too much nesting on tests like:

```ruby
with_google_chrome do
  with_current_site(site) do
    with_user(user) do
      with_budget_factory(factory) do
        ...
```

.circleci/config.yml
ferblape: Ok!

ferblape:
@amiedes you forgot to update the highlighted budget lines widget, the rest works fine.

ferblape (changes requested):
Two things left:

- highlighted budget lines widget doesn't use the new values
- in the budget lines page the charts are empty because the API request is returning no results. See http://mataro.gobify.net/presupuestos/api/data/lines/budget_line/8121/2018/per_person/I/10/economic.json?comparison[]=8187&comparison[]=8089&comparison[]=8184&comparison[]=8101&comparison[]=8245&comparison[]=45134&comparison[]=46078

amiedes:
@ferblape changes adressed!

I've also [removed some code](https://github.com/PopulateTools/gobierto/pull/2239/commits/78caa3d50e0056cc15792f208dd0eccde3520ffe) related to ES queries which looks like was being used in the original comparator but not anymore here (ranking is not used in the gobierto featured BLs). Tests pass and I don't see anything broken because of this.

ferblape:
@amiedes I have seen other part which doesn't use updated data in the budget line page. Go to http://mataro.gobify.net/presupuestos/partidas/01/2018/functional/G

The values of the children and the economic budget lines are the initial and not the updated:

Screen Shot 2019-04-25 at 11 46 39

amiedes:
@ferblape this is fixed, but I've noticed that for some budget lines we have updated data for levels 1 and 2 but not for inner levels (I've checked this in ES), and it that case a fallback is made to the initial amount:

image

I'd leave it like this because they can solve it on their side by providing all the updated data, and I don't like the idea of hiding budget lines that don't have updated data.

PR #2262, opened 2019-04-27
2245 remove default actions for modules

### LINKED ISSUE #2245: Remove default actions for modules
For moment only plans module should have admin/edit/moderate actions. Remove the actions from the rest of modules:

Screen Shot 2019-04-23 at 12 29 03

entantoencuanto:
Closes #2245 

## :v: What does this PR do?

* Allows to set different custom actions for each site module. If nothing is configured, the default action is "manage".
* Also if the module only have an action the checkboxes with the available actions for the module is not shown

Visit an admin group edition form

### Before this PR
2245-before

### After this PR
2245-after

No

No


ferblape (approved):
Code looks good.

PR #2268, opened 2019-04-30
2260 change filter projects form

### LINKED ISSUE #2260: Change projects index filter form
- [x] Remove submit button and submit when a select or date interval is changed or a text is entered in search input and sent.
- [x] If a filter is active include a "Clear filters / Limpiar filtros" button redirecting to index without filter params.

entantoencuanto:
Closes #2260

## :v: What does this PR do?
* Hides submit filter projects button and autosubmits filter on changes.
* Displays a "clear filters" button redirecting to index without filter parameters.
* Fixes some links disabling turbolinks to load datepicker after redirections.
* Adds admin option to `with` context helper method for tests and updates tests

Some problems pending:

* After filter submit button vanish, part of date selector appears out of viewport:
Screen Shot 2019-04-30 at 22 51 56

* A bug found in other parts of the application: when a date range is selected, air-datepicker can't process the value and is rendered with errors:
Screen Shot 2019-04-30 at 22 52 18

Visit a plan projects tab and play with filters

### Before this PR
2260-before

### After this PR
2260-after

No

No


ferblape:
Maybe the NaN are because instead of providing a date you are providing a date range?

test/support/contexts.rb
```diff
@@ -20,23 +20,32 @@ def with_hidden_elements
...
   Capybara.current_driver = js_driver if params[:js]
+  sign_in_admin(admin) if admin
 
   if (site = params[:site])
     stub_current_site(site) { yield }
   else
     yield(params)
   end
 
+  sign_out_admin if admin
```
amiedes: I think we have to move

```ruby
sign_out_admin if admin
Capybara.reset_session! if params[:js]
```

to the `ensure` block to avoid keeping the admin logged in for the following tests if the current tests fails

entantoencuanto: Done!

entantoencuanto:
> Maybe the NaN are because instead of providing a date you are providing a date range?

I've added some changes to `addDatePickerBehaviors`. The error is produced for date ranges and forms with more than one datepicker. I've fixed behavior for ranges and added an option to allow empty date fields.

ferblape:
Why the author filter is not working in this plan? http://esplugues.gobify.net/admin/plans/14/projects Projects don't have an author yet?

entantoencuanto:
> Why the author filter is not working in this plan? http://esplugues.gobify.net/admin/plans/14/projects Projects don't have an author yet?

The authorship of projects imported from CSV is blank, the author of a project is assigned when an admin with edit permissions creates a project.

app/models/gobierto_plans/node.rb
```diff
@@ -14,7 +14,21 @@ class Node < ApplicationRecord
...
+        begin
+          Date.parse(date)
+        rescue ArgumentError
+          nil
+        end
+      end
+
+      dates = [nil, nil].zip(dates).map { |date| date.compact.first }
+
+      where("starts_at >= ? AND ends_at <= ?", *dates)
```
ferblape: I think I'd change this AND by an OR. This way, when I search projects between two dates I'll get all the projects active in those dates. If you use an AND you'll get all the projects that start or end inside that range, which I think is more difficult to understand from the users side.

PR #2288, opened 2019-05-08
JS Errors

### LINKED ISSUE #2264: Fix JS errors
There are three JS errors in Gobierto that cause recurrent Rollbars, maybe when the data takes too long to load, or with small resolutions:

- https://rollbar.com/Populate/gobierto/items/1567/
- https://rollbar.com/Populate/gobierto/items/1447/
- https://rollbar.com/Populate/gobierto/items/1450/

In the rollbar "occurences" tab you can check the OS and the browser. We should try to avoid them.

Crashillo:
Closes #2264 

## :v: What does this PR do?
Solves some rollbar errors:

- [x] https://rollbar.com/Populate/gobierto/items/1567/
- [ ] https://rollbar.com/Populate/gobierto/items/1447/ (See comment)
- [x] https://rollbar.com/Populate/gobierto/items/1450/ <sup>1</sup>

<sup>1</sup> Caused by 1567

### DIFF
--- a/app/javascript/lib/visualizations/modules/card_simple.js
+++ b/app/javascript/lib/visualizations/modules/card_simple.js
@@ -33,7 +33,7 @@ export class SimpleCard extends Card {
 
     // If the data is 0
     if (divCard.find("div.indicator_widget.padded").find("div.widget_body").find("span.widget_figure").text() == 0)
-      divCard.remove()
+      return divCard.remove()
 
     // Append source
     this.div.selectAll('.widget_src')


Crashillo:
The 2nd rollbar item throws errors due to missing dependencies caused by mixed contents. When we download the gencat template, there are some urls over `http`, instead of `https`. Once we're in production, the site is `https` so these scripts are blocked by the browser, causing some resources unavailables. cc/ @ferblape

PR #2301, opened 2019-05-13
661 fix missing plan configuration exceptions

entantoencuanto:
Closes PopulateTools/issues#661

## :v: What does this PR do?

* Avoid some errors produced when the configuration of a plan is blank.
* Defines a default level names configuration for levels not present in configuration plan attribute. The names are taken from the I18n translations
* Uses the count of published projects in the plan summary

Edit a plan from admin, delete configuration and save.

### Before this PR
661-before

### After this PR
661-after

No

No

### DIFF
--- a/app/models/gobierto_plans/plan.rb
+++ b/app/models/gobierto_plans/plan.rb
@@ -48,23 +48,6 @@ module GobiertoPlans
       nodes.average(:progress).to_f
     end
 
-    def to_s
-      text = ""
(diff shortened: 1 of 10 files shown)


app/models/gobierto_plans/plan.rb
```diff
@@ -48,23 +48,6 @@ def global_progress
       nodes.average(:progress).to_f
     end
 
-    def to_s
```
ferblape, line 51: 👏

ferblape (approved):

PR #2308, opened 2019-05-16
651 indicators management

entantoencuanto:
related to PopulateTools/issues#651

## :v: What does this PR do?

* Defines a new `data_grid` custom field type.
* Adds association of plans nodes with indicators using custom fields records of type `data_grid`
* Uses slickgrid to edit indicators content on projects edit form.

Pending:

* For the moment, the Indicators custom field of type `data_grid` has to be created explicitly at `GobiertoPlans::Node` level, but the idea is to have custom fields definition at Plan instance level for the depending nodes.
* The current project edition form only allows to edit the indicators that already exist, there is no possibility to create new or delete existing ones. A project with an indicator exists at http://talavan.gobify.net/admin/plans/17/projects/8690/edit

All these pending tasks will be included in a new issue improving the custom fields feature.

Visit http://talavan.gobify.net/admin/plans/17/projects/8690/edit

### Before this PR

🚫 

### After this PR
651-after

No

No


ferblape:
Are we using a setting to force the numbers of the indicators to be integers? I can't set a float value

ferblape (approved):

PR #2327, opened 2019-05-27
2314 separate projects content from plan tree

### LINKED ISSUE #2314: Plans / Front / JSON: separate structure from projects content
In order to reduce the weight and processing time of the JSON used in the front, separate in different calls: 

1. Main call which loads the structure and project list (same call as we have in the admin, so potential perfomance improvements applies to both. Analyze how projects in draft/not published affects this call)
2. Individual calls to load the content of each project

entantoencuanto:
Closes #2314

## :v: What does this PR do?

* Optimizes plan tree generation in front using decorators.
* Allows the generation of plan tree not including projects
* Adds API endpoints providing data of projects of a category in the same format the complete tree has.
* Stores a list of opened tables of projects with an unique id for each, instead of reseting the list when active node is changed.

Visit front page of a plan and open projects tables

### Before this PR

2314-before

### After this PR

2314-after

No

No


ferblape:
Very nice, and the projects of a category load really fast and they get cached to avoide repeating the request 👍 

Screen Shot 2019-05-29 at 16 04 18

PR #2333, opened 2019-05-30
2320 custom fields at instance level

### LINKED ISSUE #2320: Custom fields at instance level
Currently the custom fields are defined for all instances of classes by site. The idea is to:

- Add another configuration in modules defining classes which admits definition of custom fields in their instances
- Editing an element which admits custom fields, the form shows custom fields defined for the site and model and custom fields provided by the instance the element belongs to. To add flexibility, in this case the instance should be provided by the controller as a "context instance" (for example, a project can belongs to many plans, but the project form is nested under a plan, so the context can be provided by the controller)

Steps:

- Add polymorphic relation between custom_fields and instance (current global custom fields will not have instance defined)
- Create a controller to edit, at level instance, the custom fields. The controller and views are close to current global but:
  - Must check if the instance belongs to a class allowed to define custom fields
  - Must allow the creation of custom fields only for the classes configured for the module, not for other modules
- In the custom field form use a block for the global configuration and other one provided by a context instance
- Adapt the form to generate and accept custom fields

entantoencuanto:
Closes #2320

## :v: What does this PR do?

* Adds a polymorphic association of custom fields with instance to allow creation of custom fields at instance level.
* Creates a configuration in modules to define which models allows custom fields at instance level.
* Changes controllers, forms and views to manage global custom fields and instance level custom fields.
* Configures `GobiertoPlans::Plan` to allow custom fields at instance level and includes custom field controls in admin projects edition and creation forms.
* Adds some integration tests.
* Adds a dummy control for data grid custom field type, pending of a future plugin definition.

From admin go to an existing plan configuration and click on "Custom fields" link. Define some custom fields and edit or create a project of the plan

### Before this PR

🚫 

### After this PR

2320-after

No

No

### DIFF
--- a/app/controllers/gobierto_admin/gobierto_common/custom_fields/custom_fields_controller.rb
+++ b/app/controllers/gobierto_admin/gobierto_common/custom_fields/custom_fields_controller.rb
@@ -10,13 +10,26 @@ module GobiertoAdmin
         def index
           set_class_name
 
+          @custom_field_form = form_class.new(
+            resource_name: params[:module_resource_name],
+            instance_class_name: params[:instance_type],
+            instance_id: params[:instance_id]
+          )
+          @decorated_instance = ::GobiertoAdmin::BaseResourceDecorator.new(@custom_field_form.instance) if @custom_field_form.instance
           @class_human_name = @class_name.constantize.model_name.human(count: 2)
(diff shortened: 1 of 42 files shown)


entantoencuanto:
Moved custom field management at instance level to custom fields section.

2320-after-2

The model of the instances must have a `belongs_to` association with site.

PR #2334, opened 2019-05-30
Add fragment cache to plan JSON

### LINKED ISSUE #660: IBM Notes attendees list
Import attendees from IBM Notes events.

ferblape:
Closes #660

## :v: What does this PR do?

This PR adds a fragment cache on the JSON generated with the tree of categories. It speeds-up the load of the plans in the front page.

It uses the attribute `cache_key` available in any object, so the cache can be expired calling `touch`.

1. You can visit any plans page and except if it's the first visit it should load fast
2. You should update any project attribute, move a project from a category to another, create new categories, etc and the cache should get updated
3. You should check the import/export expires the cache too


ferblape:
Turbolinks issues solved. The spinner takes a bit to appear, because the inner div doesn't have a fixed height, but I don't think it's a problem.

PR #2354, opened 2019-06-04
2323 vocabulary tags custom field

### LINKED ISSUE #2323: Custom Field: Vocabulary - Tags
Following #2321, an admin will be able to add a custom field which is based on a vocabulary, selecting type "Tags"

- When showing this custom field in the instance, the control should appear as tagging, which allows item creation (https://select2.org/tagging)
- When adding a new item, it should be created as a term for the vocabulary, in the language the user is using as UI locale

Analyze first how different would be from the previous option (Multiple select)

entantoencuanto:
Closes #2323

## :v: What does this PR do?

* Adds a tags custom field type which allow the creation of terms of tags vocabulary on creation

### Pending

* There is a known stale issue (https://github.com/select2/select2/issues/4743) that causes problems with the creation of tags when they have accents (The first character works fine with accents, but the following accents fails, observed not only in Firefox, also in Chrome and MacOS, https://github.com/select2/select2/issues/4743#issuecomment-443444285).

I'll try to change the behaviour with the `createTag` callback of select2.

### Before this PR

🚫 

### After this PR

2323-after

No

No


ferblape:
Tags with accents don't work very well, I haven't been able to create tags with accents even with your solution. I have been investigating but we need to go deeper.

ferblape:
> Tags with accents don't work very well, I haven't been able to create tags with accents even with your solution. I have been investigating but we need to go deeper.

But I'm mergin this to avoid branch conflicts.

Issue #2346, opened 2019-06-03
Plans / Plugin: Human Resources


furilo:
This plugin, using #2343 will provide a way for users to assign human resources to a project. This might be an iteration from #2344 as data structure and UI is very similar: An admin must:

- be able to add N resources to a project in the form of a slickgrid data table
- the plugin will define the structure of the available columns: 
  - resource type (autosuggest from vocabulary)
  - Cost
  - Start Date
  - End Date
- the first column will allow to select a **resource type** from a vocabulary.
  - pending:
    - decide how to configure the vocabulary for a given custom field-plugin
    - define an API so an external system can retrieve the Terms list
- the front-end component will write to the plugin hidden field after each edit of the user, a JSON representation of the data

ferblape:
For start and end dates, there's a special editor of type calendar integrated in Slickgrid that uses https://flatpickr.js.org/getting-started/.

Here's an example: https://6pac.github.io/SlickGrid/examples/example-autocomplete-editor.html

Issue #2344, opened 2019-06-03
Plans / Plugin: Indicators


furilo:
This plugin, using https://github.com/PopulateTools/gobierto/issues/2343 will provide a way for users to create Indicator Groups. An admin must: 

- be able to add an indicator group (a table)
- after adding the group, the table will appear and user can edit its contents
- the first column will allow to select an indicator from a vocabulary. 
  - **pending**: 
    - decide how to configure the vocabulary for a given custom field-plugin
    - define an API so an external system can retrieve the Terms list
- the next rows will be date fields
- the front-end component will write to the plugin hidden field after each edit of the user, a json representation of the data

amiedes:
Pending tasks (will add more):

- [x] Bug: vocabulary id is not loaded when other custom field is present, and API request returns 404
- [x] Plugin type sometimes is not visible when editing the custom field
- [x] Not all plugins will use a vocabulary, so update the custom fields form so that the vocabulary selector appears only when relevant
- [x] Translate text "Add" to add a column
- [x] Delete rows button should only appear when a row is selected
- [x] Translate delete rows button text
- [x] Refactor JS code to avoid duplicated code
- [x] Add form field title
- [x] Bug: if we assign the plugin type and the vocabulary in the same request, there's a bug that prevents the vocabulary from being saved (then 404 happens)
- [x] https://rollbar.com/instance/uuid?uuid=0ba83dc3-8a29-477e-a614-bf8831eb5267
- [x] After adding some rows, if form validations fails, information is lost when re-rendering

Pending design adjustments (i'll ping when the rest of the behavior is working properly):

- [ ] Delete rows button
- [ ] Select2

Pending to resolve in this or next issue:

- [ ]  ~rename plugin: plugin structure one plugin per engine~ #2346 
- [ ]  ~decide where to extract the common code~ #2376

ferblape:
![hidefield](https://user-images.githubusercontent.com/17616/59695092-ef514900-91e9-11e9-8e6f-389e9043f5a6.gif)

amiedes:
>  After adding some rows, if form validations fails, information is lost when re-rendering

I haven't been able to reproduce this again and it's working ok now, so I'm marking it as checked.

ferblape:
This can be reviewed in staging. Things to polish in the UI:

- design of the delete rows button, only visibile when a row is selected
- height and width of the table (I think with CSS it'll fit better)
- styles of the `<label>` of the custom field plugin
[...]

PR #2360, opened 2019-06-07
Custom field datagrid plugin

### LINKED ISSUE #2343: Plans / Custom field: Plugin
An admin must be able to add a new custom field: "Plugin". This will be the basis for the plugins we'll be creating, and we should define the way it works along the way. 

Initially, the plugin will: 

- expose the value of the field in the view (input)
- invoke the front-end code for the plugin. This code will optionally provide the UI so the user can edit the value.
- the front-end code will write the output to a hidden field, so it's saved the same way any other normal/custom field in the instance.

entantoencuanto:
Closes #2343 #2344 #2346 

## :v: What does this PR do?

* Adds a new `plugin` custom field type
* The plugin accepts a configuration on type. The available plugin types are defined in the app at `Rails.application.config.custom_field_plugins` and can be accessed by engines in vendor to add new types.
* Implements `indicators_plugin` and `human_resources_plugin`

Create some plugin custom fields on models setting available types.

### Before this PR

2343-before

### After this PR

custom_field_plugin

No, but there is a new `Rails.application.config.custom_field_plugins` array with existing plugins which can be changed by an engine railtie

Yes, describe how to add new plugins based on https://github.com/PopulateTools/gobierto/commit/cb0ada6428894c23f1ee0319291e21b008ddc142 or by adding an engine updating `Rails.application.config.custom_field_plugins` which includes the javascript required to manage the custom field payload.


ferblape:
I have updated the PR with a different approach:

- I have removed the plugins from app/javascript
- I have created the data grid plugin in vendor/

Plugins logic and JS works fine, we are ready to create new plugins from here.

amiedes:
Human resources plugin and indicators plugin can be reviewed in staging in http://talavan.gobify.net/admin/plans/17/projects/8690/edit

PR #2383, opened 2019-06-22
2347 progress custom field

### LINKED ISSUE #2347: Plans / Plugin: Progress
This plugin will calculate the progress of a project based on: 

- advancement of its Human Resources:
  - the plugin will read the resources assigned to the project, start-end dates and cost, and calculate a percentage and money spent:
    - if project has 1 resource that has start-end dated 1 jan 19-31 may 19, the resource costs 1.000, and we are on 31 mar 19 (ie. in the middle of the period), the progress is 50% and cost is 500
    - etc. 
- advancement of its Budget Execution 
  - the plugin will read the budget execution of the budget lines associated to the project and calculate a percentage and money spent 
- with both values final values will be calculated: 
  - % progress: arithmetic average
  - money: sum

Other issues: 

- progress should be calculated daily, so this should be server code that can be invoked automatically and goes project by project calculating values and updating progress accordingly
- progress should also be calculated when Budget or Human Resources associations are changed (added/edited/deleted)
  - maybe we can offer a hook so those plugins can inform this one to run, server-side? (Only drawback is that user won't see in real time the progress change, just when saving and reloading)
- as in https://github.com/PopulateTools/gobierto/issues/2344 - available resources should be read from a Vocabulary

entantoencuanto:
Closes #2347

## :v: What does this PR do?

* Adds a new type of plugin named progress
* The value of the custom field is calculated from other plugin custom fields configured in the progress plugin providing progress functions. The total progress is the average of all the values received from the configured custom fields.
* This plugin manages a value which can't be edited directly by admins in the projects form. Instead, the value is generated by a task provided by the plugin or a callback invoked when a custom field referenced from the progress plugin configuration is saved.

* Visit the custom fields configuration of a plan. Add one or more human resources plugin (human resources plugin implements a progress function).
* Add a progress plugin and add the ids or uids of human resources in the corresponding places:
```json
{
  "custom_field_ids": [
    ...
  ],
  "custom_field_uids": [
    ...
  ]
}
```

* Visit a project of the plan, fill human resources with some data and save the project. The progress will show the data calculated from previous data.

### Before this PR

🚫 

### After this PR

2383-after

#### Update

Some changes has been included:
* The callback used to update the progress payload also updates the progress of the item the progress belongs to if it has a progress attribute.
* The javascript of the plugin hides the plugin view provided by custom fields and replaces the progress select with a disabled input text containing the corresponding value.

2347-after

Yes! It changes the cron file and Ansible template should be updated.
It also changes the deploy script.

- The documentation about the configuration of the plugin is present at its README file


ferblape:
I guess the rake task is pending, let's add it to staging cron too, to valid it works.

entantoencuanto:
Added functions to calculate progress in budgets plugin in the PR for budgets plugin https://github.com/PopulateTools/gobierto/pull/2385/commits/d8535d7050b52e21ec95c99944ea9935a81e3d2c

ferblape:
@entantoencuanto could you check the failing tests?

entantoencuanto:
> @entantoencuanto could you check the failing tests?

Yes, they are related with the callbacks of plugins

amiedes:
Hey @entantoencuanto, is it possible that you also have to add the argument to be able to retrieve the progress / cost of human resources for a specific version (published one) [here](https://github.com/PopulateTools/gobierto/blob/e797e62493d923a90a1a7c3df17cec138fdc1541/vendor/gobierto_engines/custom-fields-data-grid-plugin/app/models/gobierto_common/custom_field_functions/human_resource.rb#L5-L21)?

vendor/gobierto_engines/custom-fields-progress-plugin/lib/tasks/calculate_progress.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :gobierto_admin do
+  desc "Calculate progress"
+  task calculate_progress: :environment do
```
ferblape: Could you add another namespace with the plugin name at least? `gobierto_admin:calculate_progress` is very generic

ferblape (changes requested):

ferblape:
You forgot to mention the changes in Ansible and in the deploy script.

vendor/gobierto_engines/custom-fields-progress-plugin/lib/tasks/calculate_progress.rake
entantoencuanto: The namespaced version will be: `rake custom_field_plugins:progress:calculate_progress`

PR #2385, opened 2019-06-24
2345 - Budgets plugin form custom fields

### LINKED ISSUE #2345: Plans / Plugin: Budgets
This plugin, using https://github.com/PopulateTools/gobierto/issues/2343 will provide away to associate budget lines with projects from Planification. An admin must: 

- be able to associate a year and a budget line of that year, assigning a weight for that budget line (in percentage)
- be able to remove associations
- the front-end app should check that Budgets Module is activated and has data, or fail silently otherwise 

Related with this task, we may need to create some API points so an external system can: 

- check available years and budget lines
- get the value of budget and execution for year+budget line combination

amiedes:
Closes #2345 

## :v: What does this PR do?

Implements the budgets plugin for a custom field.

- A user can add/remove associated budget lines to a project
- The corresponding amount for the selected budget line / year is refreshed when needed
- The final amount is automatically calculated when the user selects the weight

You can check the behavior in http://talavan.gobify.net/admin/plans/15/projects/8687/edit

## :eyes: Screenshots

plugin_presupuestos_3

No

No


amiedes:
I've created a [follow-up issue](https://github.com/PopulateTools/gobierto/issues/2386) for the things we've discussed. Maybe we can merge this in master so the next issues can be branched from master.

PR #2399, opened 2019-06-28
2382 project page front 2

Crashillo:
## :v: What does this PR do?
Add frontend support for gobierto_plans plugins and custom_fields

## :eyes: Screenshots
imagen


app/views/gobierto_plans/plan_types/_planification-content.html.erb
```diff
@@ -111,49 +111,74 @@
...
+            </div>
+            <div class="row">
+              <div class="mandatory-title"><%= t("gobierto_plans.plan_types.show.status") %></div>
+              <div class="mandatory-desc">{{ activeNode.attributes.status | translate }}</div>
+            </div>
           </div>
         </div>
+
+        <!-- Rails must add the markup if plugin is available -->
+        <% availablePlugins = ['plugin1', 'plugin2'] %>
```
Crashillo: cc/ @ferblape @amiedes @entantoencuanto 

First off, ruby gets all available plugins from configuration. 
1. Add the markup as following:  `<div id="WHATEVER_ELEMENT_ID"></div>`
2. Call the controller function `add` to insert the plugin as: `window.GobiertoPlans.tablerow_controller.add("FETCHED KEY", "WHATEVER_ELEMENT_ID")` <sup>1</sup>
3. Start the listeners as: `window.GobiertoPlans.tablerow_controller.show()`. That could be inserted before the plugins markup, it doesn't matter.

<sup>1</sup> Fetched key is the name of the children key of the `plugins_data` object. [Check this](https://github.com/PopulateTools/issues/issues/689#issuecomment-505869087)

app/javascript/gobierto_plans/modules/plan_types_controller.js
```diff
@@ -81,6 +81,24 @@ window.GobiertoPlans.PlanTypesController = (function() {
                   Vue.set(this.model, "children", json)
                   this.$emit("toggle");
                   this.isOpen = !this.isOpen;
+
+
+                  json.forEach(element => {
```
Crashillo: cc/ @ferblape @amiedes @entantoencuanto 

This snippet triggers an event called `gobierto-plans-mount-plugin` to inform the `tablerow_controller` that some div element has to to be mounted.
If there's no plugins, nothing happens. If so, is the plugin (tablerow_controller) who is listening such event and doing the stuff

Crashillo: To avoid unnecessary calls, I got rid of `forEach`, moving this snippet to the component `table-view` and the root component as well.
You may check the file: https://github.com/PopulateTools/gobierto/pull/2399/files?file-filters%5B%5D=.js#diff-d67a57234a740f54a76cd1b5269ef1f7

app/javascript/gobierto_plans/plugins/indicators.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <div>
+    <div class="tablerow__title">{{ title | translate }}</div>
```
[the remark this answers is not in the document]
Crashillo: We need to include https://eslint.vuejs.org/

Crashillo:
Now, there are three steps to make the plugins work:

1. HTML, `plugin` value must match with `plugins_data` key, and the vue component name
https://github.com/PopulateTools/gobierto/blob/30ff133169ee7bd6e105c5c1a1429d6d929c6d4d/app/views/gobierto_plans/plan_types/_planification-content.html.erb#L170

2. After the `fetch` API brings data, if there's `plugins_data`, it sends an event, which this function is listening to. As the component has not been mounted yet, we must wait for the nextTick
https://github.com/PopulateTools/gobierto/blob/30ff133169ee7bd6e105c5c1a1429d6d929c6d4d/app/javascript/gobierto_plans/modules/plan_types_controller.js#L321-L323

3. This is how should be initializated a plugin after a vue app is running. Any special handling must be done here (f.e. add html directives)
https://github.com/PopulateTools/gobierto/blob/30ff133169ee7bd6e105c5c1a1429d6d929c6d4d/app/javascript/gobierto_plans/modules/plan_types_controller.js#L363-L376

4. Note the plugin path above. We assume all the plugins are located in a `plugins` folder, root module level (in this case, just in `gobierto_plans`)

PR #2403, opened 2019-06-28
2387 - Improve horiz/vert content disposition for SlickGrid

### LINKED ISSUE #2387: Improve dimensions of Slickgrid-related plugins
- [ ] Avoid the empty space on the right:

image

Avoid too much empty "vertical" space when there are few rows in the table. The objective should be:

- [ ] We never show a vertical scrolling bar
- [ ] When the table is almost emtpy there's not too much empty space. Right now it's fixed via plugin JS to 500px weight. Test with 200.

amiedes:
Closes #2387 

## :v: What does this PR do?

Improves horizontal and vertical content disposition for Slickgrid tables.

...

### Before this PR

image

### After this PR

When the table is empty:

image

If you add new rows, the height is readjusted:

add_rows

For the indicators table, which has variable width, once the columns don't fit in the page anymore a horizontal scrollbar appears:

horizontal_scrolling

### DIFF
--- a/vendor/gobierto_engines/custom-fields-data-grid-plugin/app/javascripts/custom_fields_data_grid_plugin/modules/budgets_plugin.js
+++ b/vendor/gobierto_engines/custom-fields-data-grid-plugin/app/javascripts/custom_fields_data_grid_plugin/modules/budgets_plugin.js
@@ -1,7 +1,7 @@
 import { Grid, Editors, Plugins } from 'slickgrid-es6';
 import { Select2Formatter, Select2Editor } from './data_grid_plugin_select2';
 import CheckboxDeleteRowPlugin from './checkbox_delete_row_plugin';
-import { applyPluginStyles } from './common_slickgrid_behavior';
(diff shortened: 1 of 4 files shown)


amiedes:
There's a styling issue when expanding the selec2 from the last row:

image

@Crashillo can you fix this? Feel free to add commits to this PR.

You can see it in action in http://talavan.gobify.net/admin/plans/15/projects/8687/edit, or in local in happens also in http://madrid.gobierto.test/admin/plans/28856268/projects/252926840/edit

Crashillo:
@amiedes it's not straightforward. select2 container is calculated via javascript, based on how high the selector is. Perhaps, force the selector to fit the cell height. The coolest way would be handling the absolute elements, but there's a lot of wrapped elements each other quite hard to deal with.

Please, make an issue to take into account in future iterations

ferblape:
Ok, let's create other issue to fix the styles. @amiedes / @entantoencuanto  feel free to merge this PR when you can, code looks good.

amiedes:
Created follow-up issue for the styles #2425

Issue #2408, opened 2019-07-01
GobiertoPlans / Add functions to human resources


entantoencuanto:
Add 2 functions In human resources replacing cost
  * planned_cost: The sum of all the costs of the human resources
  * executed_cost: The current cost function. The sum of the costs of each human resource up to current date

ferblape:
If the dates of all the human resources are in the future the plugin returns a Not a number error.

amiedes:
Closed via https://github.com/PopulateTools/gobierto/pull/2420

Issue #2407, opened 2019-07-01
Slick grid calendar closes when changing the month


ferblape:
![datepicker](https://user-images.githubusercontent.com/17616/60432785-6bdd2200-9c03-11e9-8dcb-9a69d00932a9.gif)

amiedes:
Closed via https://github.com/PopulateTools/gobierto/pull/2420

Issue #2386, opened 2019-06-25
Plans / Budgets plugin - UX improvements


amiedes:
Right now it's not clear for a user which cells are editable and which are not, so:

- [x] Remove assigned import column
- [x] Move full import column to the rightmost position
- [x] Give all cells the same white color, and the full import (non-editable) a darker one
- [x] Display years from most recent to lest recent.
- [x] Select current year by default.
- [ ] ~Don't allow a user select a budget line unless he/she has already selected a year. If it's too complicated, an alternative is to select the last year by default.~

amiedes:
Closed via https://github.com/PopulateTools/gobierto/pull/2420

PR #689, opened 2017-05-04
Indicators rewrite

martgnz:
### What does this PR do?
This PR is a rewrite of the indicators module, creating five different card types (simple, table, bar chart, sparkline and comparison) and isolating the JavaScript for each one. As a summary, the module has gotten much more modular and easy to extend.

- The HTML has been rewritten with partials and the module is fully translated.
- The graphics are full width now, with several design and interaction improvements.
- There are several new cards with more datasets.

PR #2420, opened 2019-07-03
2408 - Update projects API & implement front

### LINKED ISSUE #2406: Include custom fields in project API
After https://github.com/PopulateTools/gobierto/pull/2393 we need to modify the core API to include custom fields information. We should:

- Exclude plugin type (since it has a custom format)
- Keep the order of the custom fields
- Include the custom field type, so the front can decide if it is displayed full-width or not.

Also:

- Create (or modify) a method to retrive the custom field records associated to a project. If there are any records defined at instance level, only this ones should be returned. In the other case, return the ones defined globally. This method will be used from the admin UI (index/show) and in the API.

amiedes:
Closes #2406, [#689](https://github.com/PopulateTools/issues/issues/689), #2424, #2386, [#696](https://github.com/PopulateTools/issues/issues/696), #2421, #2417, #2408, #2407 

## :v: What does this PR do?

Implements API changes to include custom field records for projects, and the frontend code.

TODO

## :eyes: Screenshots

TODO

No

No


test/integration/gobierto_admin/gobierto_plans/projects/project_versions_test.rb
```diff
@@ -24,6 +24,10 @@ def site
           @site ||= sites(:madrid)
         end
 
+        def default_test_context
+          { site: site, js: true, admin: admin, window_size: :xl }
```
amiedes, line 28: @ferblape @entantoencuanto if you happen to have a tests which fail due to some page contents being invisible due to window size, you can resize it like so.

entantoencuanto (approved):
🎉

PR #2438, opened 2019-07-08
Prevent initialize indicator grid with an empty object

entantoencuanto:
Closes PopulateTools/issues#704

## :v: What does this PR do?

* Fixes wrong indicator plugin display when the payload of the record contains an empty object. 

Visit a project with a plugin of type human resources and create some values. Go to the admin of custom fields and change the type to indicators of the previous fields. When editing the project the custom field should be empty and with correct height

### Before this PR

Screen Shot 2019-07-08 at 17 19 56

### After this PR

Screen Shot 2019-07-08 at 17 18 53

No

No


ferblape:
👍

PR #2452, opened 2019-07-11
2380 project permissions groups

### LINKED ISSUE #2380: Plans / Allow other admins as editors of a project
If I'm a moderator or an editor owner of a project, I should be able to grant view+edit access to a project to other editors. This UI should live inside a project. 

- [X] UI 
  - Code in #2396 - https://github.com/PopulateTools/gobierto/tree/2380-allow-other-admins
  - http://madrid.gobierto.test:3000/sandbox/admin_plans_plan_projects_project.html
  - New "Permissions" button in the top right of the item
    image
  - This opens a modal: 
    image
      - I've left a placeholder for a select2. This select2 will show existing admins in the system 
        - To discuss: only non-priviledged admins, as they already have access? Or all admins, so they receive an invitation to the item?
- [ ] Implementation
  - In the "Permissions" button - add the number of users with permission

entantoencuanto:
Closes #2380

## :v: What does this PR do?

* Adds a group_type enum to admin_groups, with two types:
  * `normal`: The groups which can be managed in backoffice as before
  * `system`: Groups created automatically by application, not visible to admins.
* Adds a polymorphic association to admin groups with resources
* Adds timestamps to admins admin groups join table records to store the time when an admin was added to a group
* Adds concerns to models and forms to create system admin groups associated with resources
Add a permissions system to manage the list of editors of a project:
  * When a project is created or updated an admin group with the project as resource is generated. The group has an edit permission on the project and the author of the project is added as member
  * If a group of permissions exists on a project a modal window is available to manage their members
  * When an admin is added or removed from a permissions list a site activity is created.

Create a project and check that the list includes the author in permissions list. Add and delete some other admins

### Before this PR

🚫 

### After this PR

2380-after

No

No


ferblape (approved):
Ok to me!

PR #2455, opened 2019-07-12
Plugins API custom categories

### LINKED ISSUE #2429: Speedup tests > Bulk all budgets seeds at once
Modify the budgets factories so when running `BudgetsSeeder.seed!` we can create all documents in a single network request.

ferblape:
Integrates data from [#issues/709](https://github.com/PopulateTools/issues/issues/709)
Closes #2429

## :v: What does this PR do?

This PR changes budgets plugin to use custom categories of type expense.

Also:

- adds a .nvmrc file
- fixes the random test related with Algolia and JSON
- removes the synk check after each yarn request

Go to a plan with the plugin enabled and check the categories displayed in the budgets plugin are the custom ones.

### After this PR

Screen Shot 2019-07-12 at 09 56 34


ferblape:
IMO user experience is not very good because it takes a lot of time to Select2 to build the whole list. 

We should probably replace the whole static list by an Ajax call that returns a small list of results (and also only the budget lines with data the selected year). What do you think?

test/factories/budgets_factory.rb
```diff
@@ -93,9 +93,7 @@ def teardown
   private
 
   def bulk(params = {})
-    result = client.bulk(params)
-    sleep 1 # wait for search index to be ready
-    result
+    client.bulk(params.merge(refresh: true))
```
amiedes, line 96: When I originally did this I tried this option but for some reason it was not working, so I used the sleep. Tests are passing so good to me.

ferblape:
PR updated to return only budget lines with data that specific year.

If the user does select the year after the budget line, if it has data the amount is loaded, otherwise "No disponible" message is added to the cell and she's able to search again.

entantoencuanto:
> IMO user experience is not very good because it takes a lot of time to Select2 to build the whole list.
> 
> We should probably replace the whole static list by an Ajax call that returns a small list of results (and also only the budget lines with data the selected year). What do you think?

I agree, it would be a good improvement. I'll add a issue for this.

Issue #2355, opened 2019-06-05
Preserve order of rows in editions jsgrid table


entantoencuanto:
When a row is edited, if the page is reloaded, the edited row appears at the end of the table

PR #2487, opened 2019-07-22
Use the same order criteria for editions in front as in admin

entantoencuanto:
Related with #2355

## :v: What does this PR do?

Uses the same order criteria for editions in front as in admin

Update editions in admin. The editions should be displayed with the same order in admin jsgrid and front

No

No


ferblape:
👍

PR #2523, opened 2019-08-21
Magnific popup animations

Crashillo:
## :v: What does this PR do?
Add some CSS available transitions for magnificPopups

Currently, the default value is `mfp-move-horizontal`


ferblape:
Deployed to staging

PR #2544, opened 2019-09-05
2537 regular admins site activities

entantoencuanto:
Closes #2537

## :v: What does this PR do?

* Avoid not manager admins to access site activities and redirects them from admin root path to the first module with permissions or admin settings if no modules are available.

## :mag: How should this be manually tested?

Sign in as regular admin

### Before this PR

2537-before

### After this PR

2537-after

## :shipit: Does this PR changes any configuration file?

No
(Changes in these files might need to update the role in Ansible)

## :book: Does this PR require updating the documentation?

No

PR #2548, opened 2019-09-09
Update supported node version

ferblape:
Follows #2544 2543

## :v: What does this PR do?

Updates node version to v12.10

Remember we have a `.nvmrc` and we should run Gobierto always in that version of node.


ferblape:
BTW, that version of node is already installed in staging and production, so if tests pass, it's safe to deploy.

PR #2608, opened 2019-10-11
Fix two rollbars

ferblape:
Fixes two Rollbars:

- https://rollbar.com/Populate/gobierto/items/2507/?item_page=0&item_count=100&#traceback
- https://rollbar.com/Populate/gobierto/items/2450/

## :v: What does this PR do?

1. When the main host was loaded (i.e gobify.net) the middleware that redirected to the main module was raising an error

2. When there's no execution data the width of a box was returning an error.


ferblape:
FYI @entantoencuanto

PR #2609, opened 2019-10-11
Allow definition of session_store key using an environment variable

entantoencuanto:
## :v: What does this PR do?

Allows configure a key for session_store configuration using an environment variable.

Define an environment variable `GOBIERTO_SESSION_KEY` to override the default `_gobierto_session` value for session store key.

This is useful when staging and production environments share the same domain but use different machines and databases. For example, if we have `production.gobierto.test` and `staging.gobierto.test` the `tld_length` argument can't be used because a `tld_length` of value 3 would generate errors since the expected domain names should have the form: `something.production.gobierto.test`

Visit staging and production environments deployed with this domain name problem and log in as admin in both environments. The session in one application shouldn't affect the other one.

- [ ] new entry in `config/application.yml`? NO
- [ ] new entry in `config/secrets.yml`? NO

No

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,7 @@ SECRET_KEY_BASE=
 RACK_ENV=development
 RAILS_ENV=development
 TLD_LENGTH=2
+GOBIERTO_SESSION_KEY=_gobierto_session
 HOST=gobierto.test
 BASE_HOST=gobierto.test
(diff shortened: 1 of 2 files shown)


ferblape:
This is OK to me, you can merge.

PR #2639, opened 2019-10-30
Set initialization vars AFTER footer

Crashillo:
## :v: What does this PR do?
**This PR is just an HYPOTHESIS** to test the behaviour of the following rollbar errors:
- https://rollbar.com/Populate/gobierto/items/1460/
- https://rollbar.com/Populate/gobierto/items/1459/
- https://rollbar.com/Populate/gobierto/items/1527/
- Perhaps this also: https://rollbar.com/Populate/gobierto/items/2705/

Looking the rollbar error times, they are triggered at startup. That's quite odd and difficult to replay, as the missing variables are defined in the `window` scope, so this hypothesis tries to delay when these variables are called, moving the script beneath the footer.

### DIFF
--- a/app/views/layouts/_gobierto_footer.html.erb
+++ b/app/views/layouts/_gobierto_footer.html.erb
@@ -2,3 +2,35 @@
 
 <!-- generic content hook -->
 <%= yield(:gobierto_footer_content) %>
+
+<!-- site js variables -->
+<script type="text/javascript">
+  I18n.defaultLocale = "<%= I18n.default_locale %>";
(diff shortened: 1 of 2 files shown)


ferblape:
OK, it looks like doesn't break anything. Worth to try.

ferblape:
OK, I'm going to merge this. You can mark the Rollbars as Resolved and let's see if they happen again.

amiedes:
Won't moving the JS out of the head cause issues with Turbolinks?

ferblape:
We tested it in staging, and because it's just variable initialization is not causing an issue.

PR #2643, opened 2019-10-30
Sourcemaps in rollbars

Crashillo:
Enables sourcemaps in rollbar reports

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -16,6 +16,15 @@ Rollbar.configure do |config|
   config.js_options = {
     accessToken: Rails.application.secrets.rollbar_post_client_item,
     captureUncaught: true,
-    payload: { environment: Rails.env }
+    payload: { 
+      environment: Rails.env,
+      client: {
+        javascript: {
+          source_map_enabled: true,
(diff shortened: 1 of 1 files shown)


ferblape:
Sourcemaps are not visible yet in staging: https://rollbar.com/Populate/gobierto/items/2808/

Crashillo:
Hmmmm, it's odd. Look at this rollbar https://rollbar.com/Populate/gobierto/items/2809/. It does have the sourcemaps, even it's the same site. 😕 🤔

amiedes:
Are we missing this?

> You'll need to generate source maps, if you aren't already. This can be done using tools like the Closure Compiler or UglifyJS 2.

Source: https://docs.rollbar.com/docs/source-maps#section-source-map-requirements

Crashillo:
Nope, we actually have sourcemaps, webpack takes care of it. You might check that out, open devtools>sources or debugger and you'll find an item at the left sidebar called Webpack (or similar). There you'll find the original javascript code

Crashillo:
I'd let this go on, meanwhile we're watching if the next rollbar errors come full-informed

PR #2656, opened 2019-11-04
Prevent clearing webpack cache in tests

### LINKED ISSUE #2654: Make webpack detect unchanged assets during tests
When launching integration tests webpack is unable to detect which assets haven't changed, so it's compiling all of the all the time.

This causes timeouts during tests, so they fail.

amiedes:
Closes #2654 

## Reason

`Rails.cache.clear` (present several times in tests) removes cache stored by Webpack, making impossible to detect if assets have changed or not.

This introduces a wrapper around `ActiveSupport::Cache::MemoryStore` to avoid clearing webpack cache.


lib/custom_cache_file_store.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def delete(*args)
+    super
+    @cache_keys.delete(args[0])
+  end
+
+  def keys
+    @cache_keys.keys
+  end
```
ferblape: Why can't you overwrite only `#clear` and left the rest of methods in the super class?

amiedes: Because the superclass does not provide a method to get the list of the existing keys, that's why I'm recording them in `@cache_keys`

ferblape: OK, now I get it.

ferblape:
Ok to me. Please @entantoencuanto review

ferblape:
If we replace FileStore by [MemoryStore](https://edgeguides.rubyonrails.org/caching_with_rails.html#activesupport-cache-memorystore) in test environment are webpack files removed too? I'd like to avoid to add a new class to the code.

entantoencuanto (approved):
Cool!

amiedes:
> If we replace FileStore by MemoryStore in test environment are webpack files removed too? I'd like to avoid to add a new class to the code.

Looks like it works also. If tests pass I'll merge.

PR #2664, opened 2019-11-07
798 gobierto data

entantoencuanto:
Closes PopulateTools/issues#798

#### Implements a new module: GobiertoData

The module configuration accepts a `db_config` setting with an `ActiveRecord::Base` connection configuration and has a `GobiertoData::Connection` model with an `execute_query` method able to send sql to the database configured in the module for the site
Adds an API endpoint at `api/v1/data` accepting requests with an `sql` param to perform queries

#### Adds an initializer to create new modules generating the initial files required in a module

The initializer can be called using:
```
rails generate gobierto_module GobiertoData
```

The branch of this PR is included in https://github.com/PopulateTools/gobierto/pull/2679 branch

- [x] new entry in `config/application.yml`?

- [ ] new site template?
- [x] new module/submodule settings?

### DIFF
--- a/app/models/gobierto_data/connection.rb
+++ b/app/models/gobierto_data/connection.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require_dependency "gobierto_data"
+
+module GobiertoData
+  class Connection < ActiveRecord::Base
+    self.abstract_class = true
+
+    class << self
+      def execute_query(site, query)
+        base_connection_config = connection_config
+        return null_query unless (db_config = site&.gobierto_data_settings&.db_config&.dig(Rails.env)).present?
+
+        establish_connection(db_config)
+        connection.execute(query)
+      rescue ActiveRecord::StatementInvalid => e
+        failed_query(e.message)
+      ensure
+        establish_connection(base_connection_config)
+      end
+
+      private
+
+      def null_query
+        []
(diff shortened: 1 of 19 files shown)


app/models/gobierto_data/connection.rb
```diff
@@ -0,0 +1,33 @@
...
+require_dependency "gobierto_data"
+
+module GobiertoData
+  class Connection < ActiveRecord::Base
+    self.abstract_class = true
+
+    class << self
+      def execute_query(site, query)
+        base_connection_config = connection_config
+        return null_query unless (db_config = site&.gobierto_data_settings&.db_config&.dig(Rails.env)).present?
```
ferblape: Why Rails.env here?

entantoencuanto: Yes, maybe this is unnecessary. I used the configuration hash generated by `config/database.yml`, with the option to have different configurations for each environment, but since each configuration on each site may be different, the environment can be useless. I'll change it

ferblape:
PR looks OK but I'd add some tests to the controller and the model. We can't forget about tests

ferblape:
If I'm not wrong we can close it and merge #2679, right?

entantoencuanto:
Yes, it's ready to merge and tests has been added

PR #2663, opened 2019-11-07
Update unemployment_rate.js

Crashillo:
Closes #2662 
Fix https://rollbar.com/Populate/gobierto/items/2776

## :v: What does this PR do?
Avoid error on `null` container. Also refactor d3 imports for that chart.

PR #2668, opened 2019-11-08
Unemployment charts

Crashillo:
## :v: What does this PR do?
Continues from https://github.com/PopulateTools/gobierto/pull/2663. Handle style error on null containers and destructuring d3.

### Discussion
These both charts plus this one https://github.com/PopulateTools/gobierto/pull/2663 are 95% similars. Worth it refactor? cc/ @ferblape


ferblape (approved):

ferblape:
Not sure, we might be chaning the observatory at some point by our visualizer.

Crashillo:
Tss, I've seen a new related rollbar. The solution is the same, but I'm afraid it's often repeated in the legacy charts. I'll prepare a PR for all affected charts (`d3.select(this.container).style(...)`), also refactoring.

PR #2755, opened 2020-01-13
Fix production queue adapter

amiedes:
Should fix https://github.com/PopulateTools/populate-ansible/issues/95 (ensure check is working before closing the issue).

At systems levels everything is configured OK, I made a quick test changing the adapter in the server and restarting and jobs were correctly processed by sidekiq:

Captura de pantalla 2020-01-13 a las 10 26 13

And the queues started showing the jobs:

image


ferblape:
Thanks!

PR #2756, opened 2020-01-13
2750 users auth token

### LINKED ISSUE #2750: Gobierto Data / Auth
* [ ] Create a token model for users allowing them to have one or more tokens
* [ ] A user has a token by default. The user must not have access to this token
* [ ] Other tokens can be created, updated or deleted for a user
* [ ] For the moment add a migration to define a token by user, and a callback to generate a user token after creation
* [ ] The token functionality is as defined for admins with jwt

entantoencuanto:
Closes #2750

## :v: What does this PR do?

* Adds a `User::ApiToken` model allowing users to have at least one token
* When a user is created a primary token is created automatically
* Adds a concern to be used in API controllers allowing to set current_user from requests looking for a token both in headers and query params.
* Protects CUD actions of queries and visualizations controllers with token authentication:
  * Authenticated users can create resources owned by them (`user_id` is assigned with current_user)
  * Authenticated users can update/delete only resources owned by them.

No

Added a page of authentication to readme.io and created an edit suggestion to gobierto data API documentation covering changes in queries and visualizations controllers.

### DIFF
--- a/app/controllers/concerns/user/api_authentication_helper.rb
+++ b/app/controllers/concerns/user/api_authentication_helper.rb
@@ -3,12 +3,6 @@
 module User::ApiAuthenticationHelper
   extend ActiveSupport::Concern
 
-  included do
-    if respond_to?(:helper_method)
-      helper_method :current_user, :user_authenticated?
-    end
-  end
-
   private
 
   def current_user
--- a/app/controllers/concerns/user/api_authentication_helper.rb
+++ b/app/controllers/concerns/user/api_authentication_helper.rb
@@ -28,6 +28,6 @@ module User::ApiAuthenticationHelper
   end
 
   def token
-    @token = request.headers["token"] || params["token"]
+    @token = request.headers["Authorization"] || params["token"]
   end
 end
--- a/app/controllers/concerns/user/api_authentication_helper.rb
+++ b/app/controllers/concerns/user/api_authentication_helper.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module User::ApiAuthenticationHelper
+  extend ActiveSupport::Concern
+
+  included do
+    if respond_to?(:helper_method)
+      helper_method :current_user, :user_authenticated?
+    end
+  end
+
+  private
+
+  def current_user
+    @current_user ||= find_current_user
+  end
+
+  def user_authenticated?
+    current_user.present?
+  end
+
+  def authenticate_user!
+    raise_unauthorized unless user_authenticated?
+  end
+
+  def find_current_user
+    return unless token.present?
+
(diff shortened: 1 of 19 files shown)


app/models/user/api_token.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class User::ApiToken < ApplicationRecord
+  belongs_to :user
+  has_secure_token
+  validates :user, presence: true
+  validates :name, presence: true, unless: :primary?
+  validates :name, uniqueness: { scope: :user }
```
ferblape: user and site?

db/migrate/20200110121522_create_user_api_tokens.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateUserApiTokens < ActiveRecord::Migration[5.2]
+  def change
+    create_table :user_api_tokens do |t|
+      t.references :user
+      t.string :name
+      t.string :token
```
ferblape, line 8: should we add a unique key in the database?

entantoencuanto: You are right

entantoencuanto: Fixed!

app/controllers/concerns/user/api_authentication_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    current_site.users.confirmed.joins(:api_tokens).find_by(user_api_tokens: { token: token })
+  end
+
+  def raise_unauthorized
+    render(json: { message: "Unauthorized" }, status: :unauthorized, adapter: :json_api) && return
+  end
+
+  def token
+    @token = request.headers["token"] || params["token"]
```
ferblape: Why not using the header "Authorization" which is the standard one?

app/controllers/concerns/user/api_authentication_helper.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module User::ApiAuthenticationHelper
+  extend ActiveSupport::Concern
+
+  included do
+    if respond_to?(:helper_method)
```
ferblape: Where is necessary this `respond_to?`

entantoencuanto: Hmm, not really

ferblape:
Ok, to me, you can deploy it.

Are you running a script to loop over all users and create a token?

entantoencuanto:
> Are you running a script to loop over all users and create a token?

Yes, in the migration ` User.all.each(&:primary_api_token!)` generates a primary token for each user

PR #2764, opened 2020-01-16
2696 gobierto data module settings

### LINKED ISSUE #2696: Gobierto Data / Front available toggle
To support cases where we store data in Gobierto Data but we don't want to provide the module to the public, we should allow an admin to toggle between "Frontend visible". If not visible, it shouldn't be able to browse any public screen.

entantoencuanto:
Closes #2696

## :v: What does this PR do?

* Adds a module setting to Gobierto Data allowing to hide its frontend part. When disabled:
  * Links to module in all main navigation menus are hidden.
  * Frontend routes of module are disabled and the user is redirected to root when trying to access them.
  * If user notifications were defined for the module, it won't be available in the notification user settings.
* Splits JSON configuration of database in two, using different keys:
  * `read_db_config`: Connection configuration to be used by datasets and queries to get data
  * `write_db_config`: Connection configuration to load data into datasets tables. The datasets method `load_data_from_file` makes use of this connection
* If the database configuration field is filled only with a connection configuration and no key is provided, the settings form interpret the configuration as `read_db_config`

As admin visit settings form of Gobierto Data module.
* Enable/disable front-end visible option
* Change configuration of module connection setting connections under `read_db_config` and `write_db_config`.

No

Yes, documentation added in readme.io for module

### DIFF
--- a/app/controllers/concerns/gobierto_common/module_helper.rb
+++ b/app/controllers/concerns/gobierto_common/module_helper.rb
@@ -8,6 +8,10 @@ module GobiertoCommon
       raise_module_not_enabled(redirect) unless site.configuration.modules.include?(module_namespace.to_s)
     end
 
+    def module_frontend_enabled!(site, module_namespace, redirect = true)
+      raise_module_not_enabled(redirect) unless site.configuration.modules_with_frontend_enabled.include?(module_namespace.to_s)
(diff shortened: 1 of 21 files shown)


ferblape (approved):
Ok to deploy!

PR #2782, opened 2020-01-21
BlockHeader vue-component refactor as library

Crashillo:
## :v: What does this PR do?
In order to create the vue componnts library, this is a sample of what changes are required for each of the components we gonna externalize.

- Move CSS classes to the *container*
- Import the new component from the library
- Add the *import*/*export* entry to the module index
- Move translations to a common place, if there are.

Due to the amount of files to review, the next PRs will be topic branches from this on.

### Caveats
To make the things easier, I've left the styles where they were `/app/assets/stylesheets`; but that implies that there are no default styles for the components, so you'll need to declare them in your module (currently, in `module-investments`):

Pros: The whole component will look like as you set
Cons: Every single time you use the component you'll have to declare the styles

Open for discussion, reviewers.

### Pending
Documentation. It will be done all-in-once, in future PRs.


ferblape:
i18n! https://app.circleci.com/jobs/github/PopulateTools/gobierto/16194

app/javascript/lib/vue-components/modules/BlockHeader.vue
```diff
@@ -32,16 +31,16 @@ export default {
     return {
       labelAll: "",
       labelNone: ""
-    }
+    };
   },
   created() {
-    this.labelAll = I18n.t("gobierto_investments.projects.all")
-    this.labelNone = I18n.t("gobierto_investments.projects.none")
+    this.labelAll = I18n.t("gobierto_common.vue_components.block_header.all");
```
ferblape, line 37: What about renaming `vue_components` to `js_components`?

Crashillo: You mean in every place, right? Both translations and folder?

ferblape: I mean the key in the translation, but you are right the folder is also vue-component, so leave it.

jorgeatgu (approved):
It's OK for me.

ferblape (approved):
It's ok to me too

entantoencuanto (approved):

PR #2783, opened 2020-01-21
Calendar vue-component refactor as library

Crashillo:
## :v: What does this PR do?
Calendar component refactor

Check [previous PR](https://github.com/PopulateTools/gobierto/pull/2782) as reference.

This is a topic branch so merge it AFTER that. Blocked 'til then

### DIFF
--- a/app/assets/stylesheets/module-investments.scss
+++ b/app/assets/stylesheets/module-investments.scss
@@ -383,7 +383,7 @@
       }
     }
 
-    &--calendar-input {
+    &--calendar-button input {
       position: absolute;
       top: 0;
       left: 0;
--- a/app/javascript/gobierto_investments/webapp/containers/home/Home.vue
+++ b/app/javascript/gobierto_investments/webapp/containers/home/Home.vue
@@ -35,6 +35,7 @@
           <!-- Filter type: calendar -->
           <template v-if="filter.type === 'daterange'">
             <Calendar
+              class="investments-home-aside--calendar-button"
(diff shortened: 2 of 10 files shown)


app/assets/stylesheets/module-investments.scss
```diff
@@ -383,7 +383,7 @@
       }
     }
 
-    &--calendar-input {
+    &--calendar-button input {
```
jorgeatgu, line 386: Maybe it would be nice to change this class to `--calendar-button-input` so too, and we avoid giving styles directly to tags.

And then add it here
https://github.com/PopulateTools/gobierto/pull/2783/files#diff-6e2de985444ba546487177ef4c1b19a7R10

Crashillo: Agree, but not in this context for now. Note the CSS file, it belongs to `module-investments`, that means the module is just using the vue-component Calendar (hence the module *shouldn't even know* there's an input tag inside) so you cannot add such class to the *input* element, otherwise you'd name the class (which belongs to the module) inside the component and that's definitely wrong.

You'll ask then, *so, why you have a class inside the component?*. Yep, `datapicker-here` is a requirement to make it work the calendar, the `air-datepicker` dependency 🤷‍♀️

I wrote down [this](https://github.com/PopulateTools/issues/issues/896#issuecomment-577299692) because we need some default styles, absolutely. I'll face it in the following PRs.

ferblape (approved):

entantoencuanto (approved):

PR #2791, opened 2020-01-23
Update datasets columns attribute in meta to include data types

### LINKED ISSUE #2769: Add column types to datasets/meta request
Datasets meta endpoint should include the column types in the meta request. Instead of returning an array with the column names, return an object where column names are the keys and the types are the values.

Remember to update the documentation.

entantoencuanto:
Closes #2769

## :v: What does this PR do?

* Replaces the columns attribute array of column names with an object with column names as keys and data types as values

Visit meta endpoint of a dataset

### Before this PR

Screen Shot 2020-01-23 at 11 28 58

### After this PR

Screen Shot 2020-01-23 at 11 27 37

No

API documentation updated

### DIFF
--- a/app/serializers/gobierto_data/dataset_meta_serializer.rb
+++ b/app/serializers/gobierto_data/dataset_meta_serializer.rb
@@ -12,7 +12,11 @@ module GobiertoData
     end
 
     attribute :columns do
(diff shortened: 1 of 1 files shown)


ferblape:
If tests are green go!

Issue #2801, opened 2020-01-24
Inject component styles only from JS, not sprockets


Crashillo:
I'm thinking about some components need a default styles to render properly, e.g. RangeBars or HorizontalCarousel, so they have to be isolated in some way, perhaps, in the VUE file inside the `style` tag. But don't `scoped` as they have to be able to be modified.

_Originally posted by @Crashillo in https://github.com/PopulateTools/issues/issues/896#issuecomment-577299692_

**UPDATE**

Let's face this in two main groups: *modules* and *components*. Modules are easier to make it work, and that's where we will start with:

- [x] Investsments. https://github.com/PopulateTools/gobierto/pull/2806
- [ ] ~Charters~
- [x] Data. https://github.com/PopulateTools/gobierto/pull/2915
- [x] Observatory https://github.com/PopulateTools/gobierto/issues/2801
- [ ] Participation https://github.com/PopulateTools/gobierto/issues/2801
- [ ] Consultations 
- [ ] ~Citizen charters~
- [x] Budgets https://github.com/PopulateTools/gobierto/issues/2801
- [ ] Activity

Use https://github.com/PopulateTools/gobierto/pull/2806 as template reference.
After that, we have to do the same stuff but foreach component inside `/assets/stylesheets`

PR #2806, opened 2020-01-27
Take out specific module CSS from the main bundle (investments)

Crashillo:
## :v: What does this PR do?
Removes any reference of a specific module CSS from the main CSS bundle. That will shrink the bundle size, bringing only those specific CSS files the module requires.

In this PR, it only includes `gobierto-investments` files.

Related to #2801 

Results of XHR requests, when the module css is isolated:
imagen

## Caveats
**IMPORTANT**: This procedure will eventually remove the module specifics CSS, but they're keep working because the main bundle still has the remaining styles.
For the time being, we're gonna removing module by module their styles, and finally we'll add for each one, what are the rest of the styles (dependencies) they'll need.

Please look at the code comments below, what are the considerations for each file. cc/ @ferblape @jorgeatgu 

Yes, about how to add CSS from now on.


app/assets/stylesheets/application.scss.erb
```diff
@@ -107,7 +107,7 @@
 @import "module-admin";
 @import "module-citizens-charters";
 @import "module-citizens-charters-custom";
-@import "module-investments";
+
```
Crashillo, line 110: First off, we remove the reference to the module file from `application.scss`

app/assets/stylesheets/mixins.scss
```diff
@@ -1,3 +1,33 @@
+/* Responsive mixin
```
Crashillo, line 1: We need to include the mixins, from legacy file `populities`, that we're using. Quite possible, this file won't suffer future modifications, once all the mixins are included.

app/assets/stylesheets/module-investments.scss
```diff
@@ -1,3 +1,6 @@
+@import "css-conf";
+@import "mixins";
```
Crashillo, line 2: Important point here. We must import, at the top of the module CSS, all the required CSS for build the module. 
Note there are only two files imported here, that's because we are importing yet the main bundle, who contains all remaining styles.

The second step of this migration is about include here EVERY SINGLE FILE we need to show the module properly. That is, for instance, we're using some CSS classes belong to the `comp-tables.scss`, then we will include here such file also.

An exception would be those CSS files related with javascript, i.e. vue-components or default styles from javascript features, like read-more, horizontal-slider... who will be imported together the JS, but don't in this point.

app/javascript/gobierto_investments/index.js
```diff
@@ -1,3 +1,5 @@
+import "../../assets/stylesheets/module-investments.scss"
```
Crashillo, line 1: Here we set the reference to the module CSS; in that way, webpack will add this file to the proper chunk.

jorgeatgu:
This way is much better, and we avoid load unused CSS in every module.

ferblape (approved):
Ok to me, I don't have a strong opinion

ferblape:
Wait, review the errors in CircleCI: https://app.circleci.com/jobs/github/PopulateTools/gobierto/16439

Crashillo:
I'm merging `master` again, let's wait for the tests again 🤯

ferblape:
Is this merged in staging?

Crashillo:
In staging

ferblape (approved):

PR #2821, opened 2020-01-28
Datasets load validations

entantoencuanto:
Related to PopulateTools/issues#932
Closes PopulateTools/issues#933

## :v: What does this PR do?

* Adds some validations to datasets `table_name` attribute to prevent errors in database transaction
* Avoids SSL checks for data sources using HTTPS protocol
* Fixes errors when schema in JSON body include uppercase letters
* Improves detection of columns defined by user schema

Use the requests of the issues (For PopulateTools/issues#932 correct the encoding of the file and use UTF-8)

No

Hmm, not really


ferblape (approved):
I miss a test on the schema thing 🤓

PR #2828, opened 2020-01-30
940 show users they are using a saved query

jorgeatgu:
Closes #N
https://github.com/PopulateTools/issues/issues/940

## :v: What does this PR do?
When user load query, we show name query, privacy status and edit button

### Before this PR
before-load-query 2020-01-30 11_26_45
### After this PR
saved-query 2020-01-30 11_25_08

### DIFF
--- a/app/javascript/gobierto_data/webapp/components/sets/SQLEditorHeader.vue
+++ b/app/javascript/gobierto_data/webapp/components/sets/SQLEditorHeader.vue
@@ -259,9 +259,12 @@ export default {
       this.runQuery()
     },
     queryParams(queryParams) {
+      this.saveQueryState = true;
+      this.showBtnCancel = true;
+      this.showBtnSave = true;
       this.disabledRecents = false;
       this.disabledSave = false;
-      this.disabledRunQuery = false;
+      this.$root.$emit('saveQueryState', true);
 
       this.labelQueryName = queryParams[0]
       this.privacyStatus = queryParams[1]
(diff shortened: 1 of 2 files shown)


app/javascript/gobierto_data/webapp/components/sets/SQLEditorHeader.vue
```diff
@@ -62,10 +62,10 @@
         <input
           ref="inputText"
           :class="disableInputName ? 'disable-input-text' : ' '"
-          :placeholder="labelQueryName"
+          v-model="labelQueryName"
           type="text"
           class="gobierto-data-sql-editor-container-save-text"
-          @keyup="nameQuery = $event.target.value"
+          @keyup="labelQueryName = $event.target.value; onSave()"
```
Crashillo: Multiple actions on DOM = method

Crashillo (changes requested):

ferblape:
Save a query doesn't work, the API returns 422

Screen Shot 2020-02-04 at 11 01 16

And the error is that the dataset is blank:

```json
{"errors":[{"source":{"pointer":"/data/attributes/dataset"},"detail":"no pot estar en blanc"}]}
```

jorgeatgu:
> Save a query doesn't work, the API returns 422
> 
> Screen Shot 2020-02-04 at 11 01 16
> 
> And the error is that the dataset is blank:
> 
> ```json
> {"errors":[{"source":{"pointer":"/data/attributes/dataset"},"detail":"no pot estar en blanc"}]}
> ```

Fixed in staging 
post-query 2020-02-04 14_43_34

entantoencuanto (approved):

PR #2830, opened 2020-01-30
enable recent queries | make modal more bigger

### LINKED ISSUE #2807: Gobierto Data / Make bigger the Recent-Queries modal
- [x] Button should be available by default. Right now if you go into a dataset, it is disabled, and becomes active if you focus on the editor. It should be active by default
- [x] This modal should be bigger, there is no reason to make it so small 

image

jorgeatgu:
Closes #2807

## :v: What does this PR do?
Enable recent queries by default.
Make more bigger modal.

### Before this PR

before-modal-recents 2020-01-30 12_27_37

### After this PR

after-modal-recents 2020-01-30 12_27_17

### DIFF
--- a/app/assets/stylesheets/module-data.scss
+++ b/app/assets/stylesheets/module-data.scss
@@ -433,8 +433,8 @@
     background-color: #fff;
     z-index: 10;
     top: 135%;
-    width: max-content;
-    min-width: 50rem;
+    width: 50rem;
+    max-width: 50rem;
 
     &.active {
       display: block;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/module-data.scss
```diff
@@ -433,11 +433,21 @@
     background-color: #fff;
     z-index: 10;
     top: 135%;
+    width: max-content;
+    min-width: 50rem;
```
Crashillo: Have you test it in mobile/tablet? Too width to be absolute. Or missing media query

Crashillo (changes requested):

entantoencuanto (approved):

PR #2842, opened 2020-02-06
944 editing a query duplicates it

jorgeatgu:
Closes PopulateTools/issues#944

## :v: What does this PR do?

Load a saved query, edit the query and save it. Now we do not duplicate the query, we edit the one that was already saved.

Staging


entantoencuanto (changes requested):
Opening a query of other user:
* [x] Hide edit icon.
* [x] If query is changed, treat it as a unsaved new query.

entantoencuanto:
It seems that the last changes have recovered the initial error. As authenticated user if I edit my own query and save changes a new query is created instead of updating the initial query

save_query_issue

Other unexpected behaviour (maybe it should be in another issue): If I access to a dataset from permalink, select queries tab and click on a query authored by me the toolbar above the sql editor box doesn't seem to reflect that I've accessed my query:
Screen Shot 2020-02-11 at 22 49 17

If I click on the same query from Datos tab and Consultas dropdown the result is as expected:
Screen Shot 2020-02-11 at 22 53 36

jorgeatgu:
You are right, fixed in Staging.

About this point 'Other unexpected behaviour...' I think it's related to this https://github.com/PopulateTools/issues/issues/948

entantoencuanto (approved):

Crashillo (commented):
The file `SQLEditorHeader.vue` is massive. Is there any way to make it simpler? Break it in minor components... keep that in mind in future PRs

Crashillo (approved):

PR #2843, opened 2020-02-06
936 the order of recent queries is not correct

### LINKED ISSUE #936: Admin report for polls results implementation
Depends on #934. Working on https://github.com/PopulateTools/gobierto/pull/1005

jorgeatgu:
Closes  PopulateTools/issues#936

## :v: What does this PR do?

Reorder the recent queries executed by user, also filter recent queries by dataset.

Staging
### After this PR

reorder-filter-queries 2020-02-06 19_00_42


app/javascript/gobierto_data/webapp/components/sets/RecentQueries.vue
```diff
@@ -18,25 +21,36 @@ import axios from 'axios'
...
+    tableName: {
+      type: String,
+      required: true
     }
   },
   data() {
     return {
-      items: []
+      items: [],
+      filterItemsByDataset: [],
```
Crashillo: You're setting all variables you need inside `data`, but you don't render them. Remove them from here.

It's ok you deal with `this.` to scope the variables, but as a good practice, you should get used to scope with `this.` only those vars you need throughout multiple functions. Otherwise, you should use `let` and `const`

I saw this behaviour multiple times, in different PRs. Do not abuse of `this.`, but declaring vars inside `data()` is definitely wrong.

Crashillo (changes requested):

entantoencuanto (approved):

app/javascript/gobierto_data/webapp/components/sets/SQLEditor.vue
```diff
@@ -79,18 +80,26 @@ export default {
       url: '',
       endPoint: '',
       recentQueries: [],
-      newRecentQuery: null
+      newRecentQuery: null,
+      orderRecentQueries: [],
```
Crashillo: You're doing the same over and over.
Do not declare variables that you won't use in the template, inside the `data()` function.
Review all the `vue` files in this PR, and remove from data all variable you don't render.

Crashillo (changes requested):

entantoencuanto (approved):

PR #2846, opened 2020-02-07
902 sql editor autocomplete

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/902

## :v: What does this PR do?
Active autocomplete for SQL and columns from datasets

Staging

### After this PR

autocomplete 2020-02-07 09_29_05


app/javascript/gobierto_data/lib/sql-theme.css
```diff
@@ -0,0 +1,21 @@
+.sql {
```
Crashillo, line 8: Too simple name

app/javascript/gobierto_data/lib/sql-theme.css
```diff
@@ -0,0 +1,21 @@
...
+  color: #0000FF;
+}
+
+.sql::after,
+.table::after {
```
Crashillo, line 13: Too simple name

app/javascript/gobierto_data/lib/sql-theme.css
```diff
@@ -0,0 +1,21 @@
...
+  float: right;
+  color: #F0964D;
+}
+
+.table {
+  color: #8A2BE2;
+}
+
+.table::after {
+  content: 'tabla';
```
Crashillo: Spanish?

app/javascript/gobierto_data/lib/sql-theme.css
```diff
@@ -0,0 +1,21 @@
...
+  color: #0000FF;
+}
+
+.sql::after,
+.table::after {
+  content: 'sql';
+  padding-left: 1.5rem;
+  font-style: italic;
+  float: right;
+  color: #F0964D;
```
Crashillo: An straight color set here? With no variable?

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
```diff
@@ -1,61 +1,62 @@
 <template>
   <div>
     <div class="codemirror">
-      <codemirror
+      <textarea
         ref="myCm"
```
Crashillo: "myCm", myCodemirror? use a proper name. What are you gonna type inside the textarea? A query? Then use names like `ref="queryEditor"`

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
```diff
@@ -1,61 +1,62 @@
 <template>
   <div>
     <div class="codemirror">
```
Crashillo: Is this a required class for codemirror?

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
```diff
@@ -75,7 +79,7 @@ export default {
         mode: 'text/x-sql',
         hintOptions: {
           completeSingle: false,
-          tables: {}
+          hint: this.hint
```
Crashillo: What is `this.hint`? Looking the code is null. Do not initialize `data` variables with another `data` variable value. Perhaps a `prop`, but we should use simple types

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
```diff
@@ -108,32 +112,44 @@ export default {
...
+      editor.showHint()
+      this.$root.$emit('activeSave', false);
+      if (this.saveQueryState === true) {
+        this.$root.$emit('updateActiveSave', true, false);
       }
     })
-    */
 
-    this.cmOption.hintOptions.tables = tables;
+    this.editor.on('focus', editor => {
```
Crashillo, line 138: Have you tried to apped this event to the textarea with `@focus`, `@change`?

Crashillo (changes requested):

app/javascript/gobierto_data/lib/sql-theme.css
jorgeatgu: This class is the codemirror. I only use it to styling autocomplete

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
jorgeatgu: `This.hint` activate autocomplete for SQL and Columns https://github.com/PopulateTools/gobierto/blob/902_sql_editor_autocomplete/app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue#L215

Crashillo: I see. That's due to the same I tell you [here](https://github.com/PopulateTools/gobierto/pull/2843#discussion_r378160857)

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
jorgeatgu: Yes, and it doesn't work :\

app/javascript/gobierto_data/webapp/components/sets/SQLEditorCode.vue
Crashillo: Even more, do not nest objects in `data()`. Every single time you do so, Vue creates observer methods for each property. In a nested object, you create observers for `hintOptions`, and then for `completeSingle`, `tables`, and so on.

That's a bad practice, who penalizes performance

entantoencuanto (dismissed):

entantoencuanto (changes requested):
If I send a query which returns empty data an exception is produced, the statistics of the query (duration and 0 rows count) are not shown and the table of the last query remains in the page
empty_response

jorgeatgu:
> If I send a query which returns empty data an exception is produced, the statistics of the query (duration and 0 rows count) are not shown and the table of the last query remains in the page
> empty_response

This error belongs to this [branch](https://github.com/PopulateTools/gobierto/tree/908_experimental_branch_dont_merge_perspective), and it has happened when changing the display of the table 

I have openened a new issue https://github.com/PopulateTools/issues/issues/965

entantoencuanto (approved):

Issue #2774, opened 2020-01-20
Queries endpoint should be able to respond a query of 1M rows


ferblape:
The queries endpoint should be able to respond to user queries that fetch a considerable amount of rows.

To define scope, I'd try to reach the 1M row milestone (and plan a different issue for 10M). I'd tackle this in an iterative way:

- objective 1: 10,000 rows
- objective 2: 100,000 rows
- objective 3: 500,000 rows
- objective 4: 1,000,000 rows

Looks like using streaming could be a good way to implement this, and modern browsers and JS libraries support it.

It's important to understand once we reach the limit, what is causing that limit and what steps we need to follow to reach the next mileston of 10M.

PR #2877, opened 2020-02-21
857 large downloads stream

entantoencuanto:
Related to #2774

## :v: What does this PR do?

* Tests different alternatives to download large query results, using streaming:
  * With an enumerator
  * Using `ActionController::Live`
* The streaming option is only available for json and csv formats

In addition to the normal endpoint there are two new endpoints that return the data in streaming:
For example:
* Normal request: `/api/v1/data/data.{csv|json}?sql=select%20*%20from%20renfe_ave%20limit%202000`
* Stream with enumerator request: `/api/v1/data/stream_data.{csv|json}?sql=select%20*%20from%20renfe_ave%20limit%202000`
* Stream using `ActionController::Live`:  `/api/v1/data/live_stream_data.{csv|json}?sql=select%20*%20from%20renfe_ave%20limit%202000`

No

No

### DIFF
--- a/app/controllers/gobierto_data/api/v1/base_controller.rb
+++ b/app/controllers/gobierto_data/api/v1/base_controller.rb
@@ -160,24 +160,24 @@ module GobiertoData
           row_num = 0
           Enumerator.new do |lines|
             lines << "{\"data\":["
-            execute_stream_query(sql) do |execution|
+            with_query(sql) do |execution|
               time = Benchmark.measure do
-                execution.stream_each do |row|
+                execution.each do |row|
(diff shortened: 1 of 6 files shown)


entantoencuanto:
Unfortunately the option of using `PG::Result#stream_each` seems to fail when another request is sent to the database because `set_single_row_mode` is disabled. I've chosen to use the normal query and iterate on the result using `each` or `each_row` (these functions don't support batches, to improve the response time we could explore a way to implement it)

With the use of `each` and `each_row` without streaming the response is slower (about 10%-20%)

PR #2881, opened 2020-02-25
Fix people query for Gencat searchbox

amiedes:
## What does this PR do?

Fixes a bug that made the people searchbox in the home page only return people that was linked to the site via events. If the person was linked only via invitations, trips etc. it did not appear in the results.

No

No


amiedes:
@entantoencuanto I'm leaving this to you. As a summary:

- The query that was generating the results for the searchbox was only looking by events
- A few weeks ago I fixed a query for the department page that took into account all models: events, trips, invitations, etc.
- I've extracted that query from `models/department.rb` and moved it into `QueryWithEvents#filter_people`, so we can use from any context.
- I've written tests fore most cases, but there are a few pending ones which I've marked with "skip"
- There's still a failing test, you can see it in CircleCI

ferblape:
@entantoencuanto did you incorporate the changes of this PR in yours?

entantoencuanto:
The related issue is fixed in my PR but i've not included the code of this. Aniway it would be useful is to reuse the tests that are already done

PR #2889, opened 2020-02-26
2885 plans csv import

### LINKED ISSUE #2885: Plans / CSV Import
## Initial import

An admin will be able to import a CSV to an empty plan and have all data imported. 

* The plan will have to be already created; destination native vocabularies (categories, status) and custom fields must be already created and empty. 
* For the time being we ignore plugin custom fields 

Tasks: 

* [x] Add external_id native field to plan model
* [x] **A user will be able to download an empty CSV that will serve as a template**: Once that a plan has the native vocabularies and custom fields configured, the user can download an empty CSV with header names they will use as a template to import data. 
  - The external_id field will be included in the example
  - If the external_id is not null, it will be used
  - If the external_id is null, we'll create a secuential ID specific to this plan (as in Github Issues of a repo). This will allow us to to import the same plan in different contexts and maintain the external_id
  
## Update

* We base the process in the external_id
* If no external_id is provided we consider the project as a new one, and a secuential external_id is generated for this project
* We ignore deleted projects - if user wants to delete a project they should do it in the UI
* We ignore the position attribute - the one in the system will be kept
* When a project is matched, all their fields will be updated
  * Categories: we only take into account the last level (previous levels are ignored)
* The import works in two phases: 
  * Validation: We validate we can update/create all projects in the CSV: If there is a missmatch in some vocabulary field (we receive a term value that doesn't exist), or any other validation error, we provide feedback to the user and nothing is changed
  * Actual import

Tasks: 

- [x] A user will be able to export the current plan in CSV
- [x] Import preparation (validations)
- [x] Feedback to the user
- [x] Actual update

### LINKED ISSUE #2923: Plans / Export plain MD as is
When exporting a plan with a custom field with markdown, export (and reimport) the markdown as is, without converting to HTML.

entantoencuanto:
Closes #2885
Closes #2923

## :v: What does this PR do?

Improves CSV import/export process in plans:
* Allows to import projects from a CSV including custom fields of any type except plugin.
* If the plan is empty, the categories vocabulary is emptied and the terms are generated from the CSV. Also new terms are added to vocabularies associated with custom fields if required.
* If the plan has projects the CSV is used to update it and errors are raised both for categories and vocabularies related with custom fields if not found in the load process.
* If the CSV contains an `external_id` column its values are used to update the projects with the same id. If the `external_id` can't be found or is not present a new project is added and the `external_id` is generated automatically when not present.
* Allows to export plans to CSV files which can be reimported with the same result.
* For empty plans the exported CSV contains columns for three levels of categories (more levels can be added or deleted), the native attributes and the custom fields of the project.
* Fixes #2923 exporting custom fields using MD as is

* Create a CSV and import.
* Export a plan to CSV and reimport it.

No

Pending


furilo:
I'm not seeing MD in the export, but converted HTML: 

image

I should be seeing the MD, which is what I see in the textarea when editing the project in Gobierto.

entantoencuanto:
Sorry, the latest branch changes with exports in MD format were not yet deployed in staging. I think it is working properly now.

furilo:
Mmm... I'm seeing that we have a bad behaviour when saving the contents of a textarea/MD custom field: if you write MD and save it, it gets converted to HTML: 

image

This doesn't happen in a normal page, and should not happen here.

entantoencuanto:
Fixed!

PR #2898, opened 2020-03-05
Allow embedding via iframe all public module routes

ferblape:
## :v: What does this PR do?

This PR prepares Gobierto to be embedable in an iframe and provides the JS to load it in a widget.

### Embedding Gobierto

How? If a parameter "embed" is provided in the URL:

- expects a configuration parameter in the site named`allowed_iframe_origin`, which sets the headers to allow iframe embeding
- sends a message to the parent iframe with the current url and page title
- adds automatically a parameter "embed" to all the urls generated in that request

### JS Widget

A JS widget is provided in `public/embed.js` and can be used adding these two lines to a page:

```
<script type="text/javascript" src="https://xxxxx.gobify.net/embed.js"></script>
<div gobierto-embed="https://xxxx.gobify.net/agendas" base-path="https://remote.host/folder/"></div>
```

Produces an iframe that is communicating with Gobierto via [postMessages](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to update the parent page URL. The parameters are:

- `gobierto-embed`: the Gobierto URL to embed
- `base-path`: the full URL where Gobierto will be embedded (necessary to calculate URLs)

The widget:

- updates URLs while navigating
- uses history API, so forward and backward buttons work

### Catch all URL

The site that will embed Gobierto needs to create a wildcard route to catch all the descendant routes. Example:

I'm embedding Madrid Gobierto agendas (https://madrid.gobierto.es) in my site https://mysite.com/madrid/agendas and I visit an event from Person 1 then in my site the URL will be updated to https://mysite.com/madrid/agendas/person-1/events

Then, the wildcard to catch all the routes in my site should be `https://mysite.com/madrid/agendas/*`

## How can it be tested

You can play with it in  https://gobierto.es/demo-embed/

## Documentation

Documentation has been updated https://gobierto.readme.io/docs/administraci%C3%B3n-de-un-sitio#integrar-gobierto-en-otra-web


ferblape:
@furilo @entantoencuanto  please review and provide feedback. Next step in my side should be update the documentation

furilo:
Issues I see (we can do this at a later stage)

- [ ] we should show the menu/submenu in this layout. We have to extract it from the header
- [ ] height control needs improvement: now I see the scrollbar of the iframe, and I should never see it

ferblape:
Branch updated with latest commits in master

ferblape:
@Crashillo @entantoencuanto sorry but we need to merge this  for a customer, but feel free to add here in the comments any suggestion in the code, as I'm openning another issue related to this to fix how the iframe calculates its height.

app/javascript/lib/commons/index.js
```diff
@@ -30,4 +31,9 @@ document.addEventListener("DOMContentLoaded", () => {
   if (!disableTurbolinks) {
     Turbolinks.start()
   }
-})
\ No newline at end of file
+})
+
+$(document).on("turbolinks:load", () => {
```
Crashillo, line 36: Hum, but this is something I already let you know: those pages whose turbolinks are disabled, they won't never reach this code.
How do you deal with that?

public/embed.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
+
+// From https://www.abeautifulsite.net/parsing-urls-in-javascript
+function parseURL(url) {
```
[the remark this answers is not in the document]
Crashillo: Why hound.bot tells you this if it's actually being used?

public/embed.js
```diff
@@ -0,0 +1,144 @@
+"use strict";
```
Crashillo, line 1: Why is in the public folder? Is webpack fetching it in some way? Why don't just move them to lib/commons?

public/embed.js
```diff
@@ -0,0 +1,144 @@
...
+      searchObject = {},
+      queries, split, i;
+  // Let the browser do the work
+  parser.href = url;
+  // Convert query string to object
+  queries = parser.search.replace(/^\?/, '').split('&');
+  for (i = 0; i < queries.length; i++) {
+    split = queries[i].split('=');
+    searchObject[split[0]] = split[1];
+  }
```
Crashillo, line 23: URLSearchParams API is made for that

public/embed.js
```diff
@@ -0,0 +1,144 @@
...
+    port: parser.port,
+    pathname: parser.pathname,
+    search: parser.search,
+    searchObject: searchObject,
+    hash: parser.hash
+  }
+}
+
+// Removes embed parameter from an URL
+function toURLWithoutEmbedParameter(url) {
```
Crashillo, line 37: Same for the URLSearchParams

public/embed.js
```diff
@@ -0,0 +1,144 @@
...
+    if (parameters.length) {
+      queryString = "?" + parameters.join("&");
+    }
+  }
+
+  return parsedURL.pathname + queryString + parsedURL.hash;
+}
+
+// Adds embed parameter to an URL
+function addEmbedParameter(url) {
```
Crashillo, line 56: Hello there. I'm URLSearchParams API feeling alone

public/embed.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+}
+
+
+function postMessageHandler(event) {
+  if(event.data.type !== 'gobiertoIframeUrl') {
+    return;
+  }
+
+  messages++;
```
Crashillo, line 74: Nice. global variables. Back to the '90s

Crashillo (commented):
Perhaps you should open new issues about this 😇

PR #2940, opened 2020-03-26
2886 automatic publication

### LINKED ISSUE #2886: Plans / Option to control automatic publication
Situation: 

* Now when you save a project, a new version is created but is unpublished. If you are in draft mode of the site or the plan, and want to do several edits (but you don't care about versions at this stage, and you want all your edits automatically published) this behaviour forces you to first save, then publish, which is not good

Possible solutions: 

* Have an option in the plan configuration such as "New edits publish that version directly"
* Change the save widget so you can save+publish in the same step. We would include
  * Checkbox: "Publish directly"
  * Checkbox: "Minor edit - don't save a version"

Thoughts @entantoencuanto ?

entantoencuanto:
Closes #2886

## :v: What does this PR do?

Adds features:
* Automatic publication of last version
* Minor changes without version upgrade (only available for last version)
* Fixes some previous bugs

No

Pending


furilo:
About the literal: "Publicar automáticamente la última versión"

Maybe it's best: "Publica automáticamente esta versión"

If you are editing a version which is not the last, and you get shown "Publicar automáticamente la última versión" it may be confusing. "Publica automáticamente esta versión" solves the possible doubts. 

OK for the rest.

ferblape (approved):

PR #2943, opened 2020-03-27
2931 terms external

### LINKED ISSUE #2931: Add an external_id to vocabulary terms
As concluded in #2920, we'll add an external_id to vocabulary terms. 

It should be visible in the UI.

entantoencuanto:
Closes #2931

## :v: What does this PR do?

* Adds a configuration to `HasExternalId` concern to set the scope to calculate automatically the `external_id` attribute if not present
* Migrates the `terms` table adding an `external_id` attribute and uses the `HasExternalId` scoped to the terms of the vocabulary
* Adds management of `external_id` attribute from admin of terms

Visit admin section of terms of a vocabulary, add terms and edit existing ones.

No

No

### DIFF
--- a/app/models/concerns/gobierto_common/has_external_id.rb
+++ b/app/models/concerns/gobierto_common/has_external_id.rb
@@ -8,15 +8,27 @@ module GobiertoCommon
       before_validation :set_external_id
     end
 
+    module ClassMethods
+      def external_id_scope(method)
+        @external_id_method = method
+      end
+
+      def external_id_method
+        @external_id_method
+      end
+    end
+
     def set_external_id
       return if external_id.present?
 
-      self.external_id = new_external_id
+      self.external_id = self.class.external_id_method.present? ? scoped_new_external_id(send(self.class.external_id_method)) : new_external_id
     end
 
     def new_external_id
+      table_name = self.class.table_name
+
       base_id = id || self.class.maximum(:id) + 1
-      if (related_external_ids = self.class.where("external_id ~* ?", "#{base_id}-\\d+$")).exists?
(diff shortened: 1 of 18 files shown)


furilo:
OK for me.

app/models/concerns/gobierto_common/has_external_id.rb
```diff
@@ -8,15 +8,27 @@ module HasExternalId
...
+      def external_id_method
+        @external_id_method
+      end
+    end
+
     def set_external_id
       return if external_id.present?
 
-      self.external_id = new_external_id
+      self.external_id = self.class.external_id_method.present? ? scoped_new_external_id(send(self.class.external_id_method)) : new_external_id
```
ferblape, line 24: Why is necessary to set automatically the external_id if it's not present? Take into account that the external id can be a string, so the autoincrement you are proposing below won't work in all scenarios.

ferblape (changes requested):

app/models/concerns/gobierto_common/has_external_id.rb
entantoencuanto: We made the decision to set an id automatically if not present in plans projects to allow updates. This behavior was extracted to the concern the terms are using also. This method detects ids formed by digits, get the maximum and sets a new id with the maximum increased and converted to string, but it can coexist with arbitrary strings set by the user manually. I'll add an option to the concern to disable automatic id generation. Well, actually is easier to remove the concern from the model :)

ferblape: ok then!

PR #2983, opened 2020-04-17
2976 transactional datasets creation

### LINKED ISSUE #2976: Data / Transactional dataset creation
The creation of a dataset and its associated table should be transactional, and if the creation of the table fails, the dataset should not be created.

It's very easy to reproduce if you do a valid call to the dataset creation api but you provide an invalid CSV or an invalid schema.

entantoencuanto:
Closes #2976

## :v: What does this PR do?

* Wraps datasets creation in FormObject used by the API in a transaction to roll back the creation of a dataset when the load of csv fails
* Adds some tests

Upload a dataset with a wrong CSV or schema

No

No

### DIFF
--- a/app/forms/gobierto_data/dataset_form.rb
+++ b/app/forms/gobierto_data/dataset_form.rb
@@ -81,6 +81,9 @@ module GobiertoData
     rescue CSV::MalformedCSVError
       errors.add(data_file.present? ? :data_file : :data_path, "CSV file malformed or with wrong encoding (expected UTF-8)")
       false
+    rescue PG::Error => e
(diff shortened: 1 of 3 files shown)


ferblape:
It works OK and the datataset is not created, but the API still returns a 500 error. Could we make it more user friendly (and we don't want to receive a Rollbar each time a user uploads a wrong CSV).

What I'm uploading is a CSV with more columns in the data than in the headers, and the error is from postgres:

```PG::BadCopyFileFormat (ERROR:  extra data after last expected column
CONTEXT:  COPY municipios_wrong_raw, line 3: "1002,u</aaa>1001,1,Amurrio,amurrio,Araba/Álava,País Vasco,-3.0013109,43.0518654"
):```

PR #2987, opened 2020-04-20
2904 endopoint for data retrieving

### LINKED ISSUE #2904: Data / Use CSV endopoint for data retrieving
- Use only CSV when retrieving data in the editor
- Default query should be SELECT * FROM table LIMIT 50
- Adapt Perspective to use CSV instead of JSON
- There are two fields that we lose: response time and metadata:
  - As reponse time use the response time of the ajax request in Javascript
  - Ignore the metadata, we should already have it from the general /metadata request, rigth?
- Right now if you have browsed in several datasets and do a query, the API gets called several times. Review that only one call is made

jorgeatgu:
Closes #2904

## :v: What does this PR do?

- Adapt Perspective to use CSV instead of JSON
- Default query SELECT * FROM table LIMIT 50
- Calculate response time with Javascript
- Remove the number of rows from the editor.
- Show private public icon on every visualization.
- Include button to remove every Visualization.
- Avoid multiple requests in datasets.
- When we switch between datasets the query must remain in each of the editor
- Disabled keyboard shortcuts while try to saving a query or visualization.

Staging
DevTools open with network panel.

Steps:

Multiple requests:
1. Go to any dataset, run a query, you should see only a request to API.
2. Goes to another dataset, run a query and check again network panel, you should see a unique request.

Disable shortcuts:
1. Try to saving a query with letters c, r(shortcuts to display modal queries), shortcuts dont be fired and the request should be unique.

Visualizations tab:
1. Go to editor and save two or more visualizations(at least one of them private)
2. Go to Visualization tab, you should see icons for public and private visualizations, also another button to delete any visualization.


app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -126,13 +125,12 @@ export default {
       recentQueries: [],
       resourcesList: [],
       currentQuery: null,
-      items: [],
+      items: '',
```
ferblape, line 128: Why is items now by default a string? Sounds like this variable should be an empty array by default

ferblape (approved):
Works fine

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: Because now we receive from the API a string with all the CSV information instead of a JSON array.

app/javascript/gobierto_data/lib/factories/data.js
```diff
@@ -2,7 +2,7 @@ import { baseUrl } from "../commons.js";
 import { getToken } from "../helpers";
 import axios from "axios";
 
-const endPoint = `${baseUrl}/data`;
+const endPoint = `${baseUrl}/data.csv`;
```
Crashillo, line 5: If the endpoint returns a CSV, shouldn't the `"Content-type": "application/json"` (line 8) set as well accordingly? I ask, not sure. cc/ @ferblape @entantoencuanto

app/javascript/gobierto_data/webapp/components/Dataset.vue
Crashillo: Hummmm, it's confusing. Even though we'll get an CSV... who is transforming into an array? Are you passing to all children components a string? Wouldn't be better parse the endpoint response in this component, and set `items` as an array for all its children?

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -221,25 +219,36 @@ export default {
...
-    );
+  activated() {
+    this.$root.$on("deleteSavedQuery", this.deleteSavedQuery);
+    // change the current query, triggering a new SQL execution
+    this.$root.$on("setCurrentQuery", this.setCurrentQuery);
+    // execute the current query
+    this.$root.$off("runCurrentQuery");
+    this.$root.$on("runCurrentQuery", this.runCurrentQuery);
+    // save the query in database
+    this.$root.$off("storeCurrentQuery");
```
Crashillo, line 241: Why are there 2 `$off` events here?

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -377,28 +386,20 @@ export default {
...
-      // wrap the result in an small number of records
-      let query = "";
-      if (this.currentQuery.includes("LIMIT")) {
-        query = this.currentQuery;
-      } else {
-        query = `SELECT * FROM (${this.currentQuery}) AS data_limited_results LIMIT 100 OFFSET 0`;
-      }
+      let query = this.currentQuery;
 
       const params = { sql: query };
```
Crashillo: Since we don't need to do nothing to the `this.currentQuery`, just left `const params = { sql: this.currentQuery };` Save time and code!

app/javascript/gobierto_data/webapp/components/Dataset.vue
ferblape: right, is a CSV, thanks!

app/javascript/gobierto_data/webapp/components/commons/RecentQueries.vue
```diff
@@ -38,7 +38,10 @@ export default {
   created() {
     document.addEventListener("keyup", this.onKeyUpNextRecentItem)
   },
-  beforeDestroy() {
+  activated() {
+    document.addEventListener("keyup", this.onKeyUpNextRecentItem)
+  },
+  deactivated() {
```
Crashillo: This is what I say about the `keep-alive`. Why `RecentQueries` should be keep alive? it's only present at DataTab, if I remember

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <perspective-viewer ref="perspective-viewer" />
+  <perspective-viewer v-if="items" ref="perspective-viewer" />
```
Crashillo, line 2: Since there's the unique element here, perhaps better don't mount the component Visualizations if there's no items. I mean, set this `v-if` in its parent.

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -136,6 +136,7 @@ export default {
     onKeyUpTextHandler(event) {
       const { value } = event.target
       this.labelValue = value
+      this.$emit('disable-key-events')
```
Crashillo: Are you trying an `stopPropagation`? (1/2)

app/javascript/gobierto_data/webapp/components/sets/data/SQLEditorHeader.vue
```diff
@@ -56,6 +56,7 @@
       :placeholder="labelQueryName"
       :value="queryName"
       @save="onSaveEventHandler"
+      @disable-key-events="removeKeyboardListener"
```
Crashillo: Yes, you're trying an `stopPropagation`! 🥳  (2/2)

Crashillo (changes requested):

app/javascript/gobierto_data/lib/factories/data.js
entantoencuanto: The content-type returned for csv is `text/plain`. I seem to remember that we decided to use `text/plain` to avoid default browser behaviour of downloading a file instead of showing the content, but if required we can change the content type to `text/csv`

app/javascript/gobierto_data/webapp/components/sets/data/SQLEditorHeader.vue
jorgeatgu: Fixed, I tried to use `stopPropagation` but doesn't work with `keyup`, with `keyup` the event is already fired.

https://github.com/PopulateTools/gobierto/blob/3813890173cd667de1714e76c04de0efe69c0dbe/app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue#L12

I've changed it for `keydown`

Crashillo: I had to set as well the events for the EditorCode/Header as `keydown` because this:
https://github.com/PopulateTools/gobierto/blob/3813890173cd667de1714e76c04de0efe69c0dbe/app/javascript/gobierto_data/webapp/components/sets/data/SQLEditorCode.vue#L107-L108

To make the `stopPropagation` work, both events should be the same, otherwise nothing will happen. That is, both elements must listen `keydown` or `keyup` or whatever. Also, the keyCodes are different depending on the event!! 🤦‍♀️ 🤦‍♂️ 🤷‍♀️ 🤷‍♂️
Keep care of that!

app/javascript/gobierto_data/webapp/components/Dataset.vue
Crashillo: What component is parsing the CSV response?

app/javascript/gobierto_data/webapp/components/commons/RecentQueries.vue
```diff
@@ -36,21 +36,23 @@ export default {
     }
   },
   created() {
-    document.addEventListener("keyup", this.onKeyUpNextRecentItem)
+    document.addEventListener("keydown", this.onKeyUpNextRecentItem)
   },
   beforeDestroy() {
-    document.removeEventListener("keyup", this.onKeyUpNextRecentItem)
+    document.removeEventListener("keydown", this.onKeyUpNextRecentItem)
```
Crashillo: Since you've renamed the event to `keydown`, update the function names accordingly

app/javascript/gobierto_data/webapp/components/commons/RecentQueries.vue
```diff
@@ -36,21 +36,23 @@ export default {
...
   },
   methods: {
     onKeyUpNextRecentItem(event) {
-      const { keyCode } = event
 
+      const { keyCode } = event
       // TODO: evitar el comportamiento del browser
```
Crashillo: Remove the TODO if solved

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -9,7 +9,7 @@
         :disabled="!isSavingPromptVisible"
         type="text"
         class="gobierto-data-sql-editor-container-save-text"
-        @keyup.stop="onKeyUpTextHandler"
+        @keydown.stop="onKeyUpTextHandler"
```
Crashillo: Update function names

app/javascript/gobierto_data/webapp/components/sets/data/SQLEditorHeader.vue
Crashillo: Have you deprecated the `disable-key-events`? If so, remove dead-code

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: `Dataset.vue` call API and get response, props-down to `DataTab.vue` > `SQLEditorResults.vue `> `Visualizations.vue`

ferblape:
@jorgeatgu there's a conflict with master

@Crashillo only your +1 is pending to merge

jorgeatgu:
> @jorgeatgu there's a conflict with master
> 
> @Crashillo only your +1 is pending to merge

Done!

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -9,7 +9,7 @@
         :disabled="!isSavingPromptVisible"
         type="text"
         class="gobierto-data-sql-editor-container-save-text"
-        @keyup.stop="onKeyUpTextHandler"
+        @keydDownn.stop="onKeyUpTextHandler"
```
Crashillo: Typo

```suggestion
        @keydown.stop="onKeyDownTextHandler"
```

Crashillo (changes requested):

Crashillo (approved):

PR #3011, opened 2020-04-25
Ruby 2.7.1

ferblape:
## :v: What does this PR do?

Upgrades Ruby version to 2.7.1

It's been released to staging. We'll play around for a few days and then release to production.


app/queries/gobierto_admin/admin_resources_query.rb
```diff
@@ -63,7 +63,7 @@ def groups_admins_join_table
     end
 
     def moderations_table
-      @moderations_table ::GobiertoAdmin::Moderation.arel_table
+      @moderations_table ||= ::GobiertoAdmin::Moderation.arel_table
```
ferblape, line 66: Careful @entantoencuanto

entantoencuanto: ups!

PR #3014, opened 2020-04-28
3003 publication preference

### LINKED ISSUE #3003: Plans / When importing a CSV take into account the publication preference
This is activated in https://diba-plan.gobify.net/admin/plans/41/edit

image

but projects don't get automatically published: 

image

If the preference is active, imported projects should automatically be published.

entantoencuanto:
Closes #3003

## :v: What does this PR do?

* Makes importation from CSV to take into account the Publish new versions of projects automatically plan preference. If not checked, the result of importing a plan from a CSV depending of the setting is:
  * When not checked the visibility status of each project remains unchanged and also the published version for published projects. New projects are created in draft status and for the existing projects a new version is added if there are changes.
  * When checked a new version is added for the existing projects if there are changes and the last version is published. The new projects are created with published visibility level.

Import a plan using the setting

No

Pending


ferblape (approved):

ferblape:
Ping @entantoencuanto

PR #3018, opened 2020-04-30
Fixes auto-suggest on editor

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/991

## :v: What does this PR do?

The columns of the datasets are displayed while typing in the editor.

Staging

### Before this PR
before_showautosuggest 2020-04-30 18_10_23

### After this PR
after_showautosuggest 2020-04-30 18_09_41


ferblape:
After you select an option in the autocomplete could you add a space? 

I'm not sure about this because it's true that in your code editor when you autcomplete the cursor stays in the end of the completed word, and it's not a problem to me there, but in the query editor it's causing me a lot of trouble. Thoughts?

app/javascript/gobierto_data/webapp/components/sets/data/SQLEditorCode.vue
```diff
@@ -107,32 +110,53 @@ export default {
...
+      if (!isEditorActive && e.keyCode > 64 && e.keyCode < 91) {
+        editor.showHint({ completeSingle: false })
+      }
+
+      this.$root.$emit('enableSavedButton')
+    },
+    onChange(editor) {
+      this.mergeTables();
+      const value = editor.getValue();
+      this.$root.$emit("setCurrentQuery", value);
```
Crashillo, line 136: You're updating the query only `onchange`event. Is that what we want? I say so, because as far as I know, `onchange` is triggered only when you *blur* the editor, so ctrl+enter event does not run the query properly

jorgeatgu: `change` event for CodeMirror is different to `onchange`

> "change" (doc: CodeMirror.Doc, changeObj: object)
    Fired whenever a change occurs to the document. changeObj has a similar type as the object passed to the editor's "change" event.

[documentation](https://codemirror.net/doc/manual.html#event_doc_change)

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -366,9 +366,6 @@ export default {
         this.isQueryModified = true;
       }
 
-      // set the new query, trimming it to remove potentially harmful voids
-      this.currentQuery = sql.trim();
```
jorgeatgu, line 368: I've had to remove this method. `trim()` removes whitespace, and this causes the query to updated and resend to the editor but without whitespaces, which causes strange and annoying cursor movement in the editor.

jorgeatgu:
> After you select an option in the autocomplete could you add a space?
> 
> I'm not sure about this because it's true that in your code editor when you autcomplete the cursor stays in the end of the completed word, and it's not a problem to me there, but in the query editor it's causing me a lot of trouble. Thoughts?

Not sure about this. I've tried other editors like CodePen, JSBin, and they don't add a space after autocomplete. I made a few changes, so when you can, play with the editor again.

ferblape:
Ok, let's keep that behaviour by default, thanks

BTW, there's something broken in staging, no matter what query I run, the request to the server is always executing the default query. See the screenshot:

Screenshot 2020-05-07 at 09 27 41

Could you confirm?

jorgeatgu:
> Ok, let's keep that behaviour by default, thanks
> 
> BTW, there's something broken in staging, no matter what query I run, the request to the server is always executing the default query. See the screenshot:
> 
> Screenshot 2020-05-07 at 09 27 41
> 
> Could you confirm?

My fault. Fixed.

ferblape (approved):

Crashillo (approved):

PR #3025, opened 2020-05-05
2930 datasets performance response

### LINKED ISSUE #2930: Data / Improve first load time
If you enter https://getafe.gobify.net/datos for the first time, you have to wait several seconds with a blank screen. It seems https://getafe.gobify.net/api/v1/data/datasets is taking +5s to load: 

image

If you load it a few times in a row, it seems something has been cached and the response takes 400-500ms (which is not super fast, but kind of ok). 

Take a look why the first cold request takes 5s. I'd start looking at queries done and their time.

entantoencuanto:
Closes #2930

## :v: What does this PR do?

Make some changes to improve data response. The controller by default gets the result of the required queries using directly the connection pool.

The times generating csv from a remote server are better and improve as the number of csv rows is greater. For the query in https://github.com/PopulateTools/issues/issues/1008#issuecomment-621899116 these are the improvements 

| Limit | % of time saved for each csv generation (remote server) |
| ----- | ----- |
| 10 | 14.3% |
| 100 | 7.4% |
| 1000 | 17.5% |
| 10000 | 40.4% |

However for small queries the improvement is poor.

No

No

### DIFF
--- a/app/controllers/gobierto_data/api/v1/datasets_controller.rb
+++ b/app/controllers/gobierto_data/api/v1/datasets_controller.rb
@@ -165,7 +165,7 @@ module GobiertoData
 
         def cached_item_json
           Rails.cache.fetch("#{@item.cache_key}/show.json") do
-            query_result = execute_query @item.rails_model.all
+            query_result = execute_query @item.rails_model.all, include_stats: true
             {
               data: query_result.delete(:result),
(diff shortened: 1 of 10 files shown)


ferblape:
Changes are good, but I'll keep the issue open until we are sure we can't dig more on the performance.

PR #3027, opened 2020-05-05
Integrate precommit hook to normalize i18n

### LINKED ISSUE #2836: Add Husky setup
Install https://github.com/typicode/husky to add hooks to the PRs. That will avoid bad commits/pushes. Useful ones are:

- [x] Run `i18n-tasks`
- [ ] Run linters

ferblape:
Closes #2836

## :v: What does this PR do?

Adds a pre-commit hook that runs `i18n-tasks normalize`

Husky is sooo simple and I think it can be very useful, to avoid errors like https://app.circleci.com/pipelines/github/PopulateTools/gobierto/584/workflows/f3ab774d-772d-490c-b845-0333217e1a5b/jobs/19159

Just run `yarn install` and see how each time you commit the hook is executed 🤩

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "babel-eslint": "^10.0.3",
     "eslint": "^6.4.0",
     "eslint-plugin-vue": "^5.2.3",
+    "husky": "^4.2.5",
     "sass-lint": "^1.13.1",
     "webpack-bundle-analyzer": "^3.3.2",
     "webpack-cli": "^3.3.1",
@@ -76,6 +77,12 @@
   "scripts": {
     "jslint": "eslint $(git diff --name-status master | grep -E '.js$|.vue$' | cut -f 2)",
     "jslint:fix": "yarn run jslint --fix",
-    "csslint": "sass-lint -v $(git diff --name-status master | grep -E '.css$|.scss$' | cut -f 2)"
+    "csslint": "sass-lint -v $(git diff --name-status master | grep -E '.css$|.scss$' | cut -f 2)",
+    "pre_commit_i18n_check": "script/check_i18n.sh"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run pre_commit_i18n_check"
+    }
   }
 }
--- a/script/check_i18n.sh
+++ b/script/check_i18n.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+changes=`git status | grep -E 'config/locales/' | wc -l`
+if [ $changes -gt 0 ]; then
+  echo "I18n changes detected. Running normalization...."
+  i18n-tasks normalize && i18n-tasks health && git add config/locales/
+fi
--- a/yarn.lock
+++ b/yarn.lock
@@ -648,6 +648,13 @@
   dependencies:
     regenerator-runtime "^0.13.2"
 
+"@babel/runtime@^7.9.2":
+  version "7.9.6"
+  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.9.6.tgz#a9102eb5cadedf3f31d08a9ecf294af7827ea29f"
+  integrity sha512-64AF1xY3OAkFHqOb9s4jpgk1Mm5vDZ4L3acHvAml+53nO1XbXLuDodsVpO4OIUsmemlUHMxNdYMNJmsvOwLrvQ==
+  dependencies:
+    regenerator-runtime "^0.13.4"
+
 "@babel/template@^7.1.0", "@babel/template@^7.4.4", "@babel/template@^7.6.0":
   version "7.6.0"
   resolved "https://registry.yarnpkg.com/@babel/template/-/template-7.6.0.tgz#7f0159c7f5012230dad64cca42ec9bdb5c9536e6"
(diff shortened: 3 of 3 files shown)


entantoencuanto:
It sounds veery good!

Crashillo:
Why not one step beyond and `i18n-tasks normalize && i18n-tasks health`? (and what else)

package.json
```diff
@@ -77,5 +78,10 @@
     "jslint": "eslint $(git diff --name-status master | grep -E '.js$|.vue$' | cut -f 2)",
     "jslint:fix": "yarn run jslint --fix",
     "csslint": "sass-lint -v $(git diff --name-status master | grep -E '.css$|.scss$' | cut -f 2)"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "i18n-tasks normalize"
```
Crashillo: And also `"pre-push": "npm run jslint:fix && npm run csslint"` ?

entantoencuanto:
Hmm, something seems to not work as expected. I've make a test and created a unused translation which requires normalization. I added the changes and commited. The commit was created with my changes but the result of normalization was not included in it remaining as not staged changes.

package.json
ferblape: great, I'll investigate it

eltercero:
Same as @entantoencuanto says. I changed some translation keys order and tried to commit. The commit succeeded (even if it wasn't normalized) and the `i18n-taks normalize` task created a new unstaged file with the normalized order. 

I guess you can always amend the last commit, but it's easy that it may go unnoticed.

package.json
jorgeatgu: I prefer use pre-commit instead of pre-push.

Pre-push:
- Add files with some mistake like bad indentation or vue/order-components
- git commit -m 'foo'
- git push
❌ husky...

Pre-commit:
- Add files with some mistake like bad indentation or vue/order-components
- git commit -m 'foo'
❌ husky...

I use lint-staged in my projects. [I've seen that people also use it with husky.](https://codeburst.io/continuous-integration-lint-staged-husky-pre-commit-hook-test-setup-47f8172924fc)

ferblape: I also prefer per-commit just to avoid doing two commits: one with the wrong files and the next one with the fixed files.

ferblape:
@entantoencuanto @eltercero please review when you can again, and let me know if you feel it a bit slower

@Crashillo @jorgeatgu I prefer to do a first release with just i18n things which I understand. For a second release we can add your linters, because I've run them in my laptop and got hunders of errors. I prefer we fix those errors and the linters show only what's wrong.

eltercero:
@ferblape it now works great 👍 the only drawback I can see is that it takes around 1 minute to commit. I don't know if Husky supports such a thing, but it'd be nice if it'd only run if there's any staged file under `config/locales`. Otherwise, maybe we can create our custom script that would check that condition and then run `i18n...` only if the condition is met. Husky may call it from the pre commit hook.

Crashillo:
> because I've run them in my laptop and got hunders of errors

Hundreds, huh? I wonder why :trollface:

ferblape:
@eltercero done, I'm checking first if there are changes in `config/locales/`

eltercero:
> @eltercero done, I'm checking first if there are changes in `config/locales/`

It does not seem to work, it runs the i18n health report even if the commit change is in a Rails controller or a vue file. Besides, it's very very slow this time. It takes a couple of minutes just to start  (maybe it's my computer). 

When I run `git diff --name-status master` from `husky` branch this is what I get:

```
M       app/assets/stylesheets/comp-block-header.scss
M       app/assets/stylesheets/module-dashboards.scss
M       app/controllers/application_controller.rb
M       app/javascript/gobierto_dashboards/modules/contracts_controller.js
M       app/javascript/gobierto_dashboards/webapp/containers/contract/ContractsIndex.vue
M       app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
M       app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
M       app/javascript/gobierto_dashboards/webapp/containers/summary/Summary.vue
M       app/javascript/gobierto_dashboards/webapp/lib/config.js
M       app/javascript/lib/visualizations/modules/amount_distribution_bars.js
M       app/javascript/lib/visualizations/modules/group_pct_distribution_bars.js
M       app/javascript/lib/vue-components/modules/BlockHeader.vue
M       config/locales/gobierto_dashboards/views/ca.yml
M       config/locales/gobierto_dashboards/views/en.yml
M       config/locales/gobierto_dashboards/views/es.yml
M       package.json
A       script/check_i18n.sh
M       test/integration/gobierto_dashboards/dashboards_contracts_test.rb
M       yarn.lock
```

But the changes in my branch are in a .rb and in a .vue file.

ferblape:
The grep was wrong, now it's ready for review!

eltercero (approved):
Much better now! If there are no locale changes, it takes a few seconds but it's bearable. 

I did a few tests with changes in locales:

- If only the order is incorrect, it reorder the keys and does the commit without complaining (which is nice)

- If there are missing keys, it shows the error and abort the commit.

So, I think it works well now 👍

ferblape:
Thanks, let's go with this for the moment. Happy to get improvements!

PR #3029, opened 2020-05-06
Includes some improvements when working with queries and visualizations

### LINKED ISSUE #2973: Data / Save query-vizz interaction details
I suggest we adjust several details when working with queries-vizzs:

1. if I load a query that is not mine, or I'm logged out, we should not allow any interaction. When pressing "Save" redirect to login (or show login modal if we can)
2. if I load a query that is not mine and I'm logged in:
  - by default (= have not modified the query yet): just show the query title (not the icon, not the button)
  - if I modify the query: 
    - We show a button: I propose that instead of "Edit" we show "Fork" (with a tooltip: "Click to create your own version of this query"). 
    - If you click on Fork: the input activates and you get focused on it, with the text in the input selected (so you can start typing right away to substitute it). The action button would be back to "Save", with the control of public/private (as when you save a query for the first time)
3. The query-title-input (active or not) should be wider to avoid situations like this (you have space but the title is hidden due to the textarea being narrow): 
image

Contribute your ideas.

jorgeatgu:
Closes #2973 

## :v: What does this PR do?

Includes a fork button when we loaded a query that is not ours.
Increases the width of the input if the name of the query is too large.
Showed string "consulta modificada" when modified the query on editor.

Staging

### After this PR


Crashillo:
Solve the conflicts first. Are you currently working on this? If so, set it as draft in the meantime, it's faster to detect it.

jorgeatgu:
> Solve the conflicts first. Are you currently working on this? If so, set it as draft in the meantime, it's faster to detect it.

Conflicts resolved. You can review

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -592,6 +606,40 @@ export default {
     },
     isSavingPromptVisibleHandler(value) {
       this.isSavingPromptVisible = value
+    },
+    QueryIsNotMine() {
```
Crashillo: camelCase 🐫

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -139,32 +177,54 @@ export default {
...
+      });
+    }
+  },
   methods: {
     onClickSaveHandler() {
+
+      if (getUserId() === '') {
+        location.href = '/user/sessions/new?open_modal=true';
+        return false;
+      }
```
Crashillo: Back to the old times. Why is this here? In the saving dialog? Didn't we move it to the Dataset (the parent) component?

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -200,13 +260,51 @@ export default {
...
+      const inputValueLength = inputValueSplit.length
+
+      if (inputValueLength > 25 && inputValueLength < 50) {
+        const newWidth = inputValueLength * 7.5
+        this.$nextTick(() => this.$refs.inputText.style.width = `${newWidth}px`);
+      } else if (inputValueLength >= 50) {
+        this.$nextTick(() => this.$refs.inputText.style.width = '369.5px');
+      } else {
+        this.$nextTick(() => this.$refs.inputText.style.width = '200px');
+      }
```
Crashillo, line 295: Whaaaaaaaaaat?

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
jorgeatgu: I have an explanation for this, check if the user is logged in or not should be the first thing when the user click the button. If we move this to the dataset, we're emitting an event and doing checks to finally send the user to the login, better to send it first, right? And we need this to check if the user is logged in when trying to save a visualization.

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
jorgeatgu: haaaaaappeeens?

Crashillo: The absolute numbers there, in the style width. They're yelling: why!?

Crashillo (approved):

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
Crashillo: My idea is about sending a prop (I don't know if it already exists), something like `isUserLogged`, a boolean, so when this component is rendered, already it knows the status, and display/hide stuff consequently. No querying at all, just a flag

PR #3035, opened 2020-05-06
2996 dashboards module configuration

### LINKED ISSUE #2996: Implement dashboards module configuration
We need admins to be able to activate a dashboard, and provide the URLs of the datasets of the different resources (contracts, tenders in our case). These URLs will be read by the frontend application.

Gobierto provides a generic model to store modules configuration in JSON format. See other modules, such as investments or data to see how it's implemented.

The format of the configuration should be:

```
{
  dashboards: {
    contracts: {
      enabled: true,
      data_urls: {
        contracts: "http://example.com/contracts.csv",
        tenders: "http://example.com/tenders.csv"
      }
    }
  }
}
```

The module should implement a validation over this configuration to prevent admins to store a wrong configuration (as it's done in data module, for example)

eltercero:
Closes #2996 

## :v: What does this PR do?

This PR adds the ability to edit the configuration of the Dashboards module. It uses the GobiertoModuleSettings and it only stores a single attributes, a json with the all the dashboards. 

By default the field is empty. I couldn't find a way to add documentation for the field to suggest the proper json structure (which is for now as a method in `GobiertoDashboards.default_dashboards_configuration_settings`). 

After enabling the module, if you access the configuration through the sidebar link you'll be able to see the settings page. 

Some valid values for the field:

```json 
{
  "dashboards": {
    "contracts": {
      "enabled": false,
      "data_urls": {
        "endpoint": ""
      }
    },
    "tenders": {
      "enabled": false,
      "data_urls": {
        "endpoint": ""
      }
    }
  }
}
```

```json 
{
  "dashboards": {
    "contracts": {
      "enabled": false,
      "data_urls": {
        "endpoint": ""
      }
    }
  }
}
```

```json 
{
  "dashboards": {}
}
```

The value should **not** be valid if any of these happen:

- main 'dashboards' key is missing
- any of the inside keys is not one of the available dashboard names (for now they are just `contracts` or `tenders`)
- field is completely empty
- it is not a valid JSON

### After this PR

Captura de pantalla 2020-05-06 a las 16 44 35

- [ ] new site template?


ferblape (approved):

ferblape:
Just a clarification, there'll be only a dashboard for contracts and tenders, they are not sepparated dashboards, but as we agreed on the call, this can be fixed in the next issue.

app/views/gobierto_admin/gobierto_dashboards/configuration/settings/edit.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+<div class="tabs">
+  <ul>
+    <li class="active">
+      <%= link_to t("gobierto_admin.gobierto_dashboards.configuration.settings.title"),
+            edit_admin_dashboards_configuration_settings_path %>
+    </li>
+  </ul>
+</div>
+<% Rails.logger.debug @settings_form.dashboards_config %>
```
entantoencuanto: Is this going to be used in production?

ferblape: Not yet ;) But that's a good catch.

eltercero: yes it is. I'll fix it. Thanks!

entantoencuanto (approved):

PR #3048, opened 2020-05-08
2941 worker perspective

### LINKED ISSUE #2941: Data / Research if we can show various Perspective visualizations without a worker for each
I guess workers are used for the streaming part; since we are not using that, check if we can get rid of them to allow having various visualizations in the same page (for listings or dashboards) 

(although maybe workers are also used to do the calculations needed to paint visualizations)

jorgeatgu:
Closes #2941 

## :v: What does this PR do?

We now only use one `worker` for all Perspective Vizz.

Staging(if you've saved vizz)

### Before this PR

before_perspective

### After this PR

after_perspective


ferblape (changes requested):
When the query doesn't return results there's a js error. Steps to reproduce:

1. Go to https://demo-datos.gobify.net/datos/datadista-ccaa-covid19-confirmados-pcr-long/editor
2. Run a query that doesn't return results. I.e `SELECT * FROM datadista_ccaa_covid19_confirmados_pcr_long where fecha='2020-06-30'`

Crashillo (approved):

jorgeatgu:
@ferblape Solved, but whats happens when the result is empty, clear the perspective table? not clear? I've some doubts

@Crashillo I've made a few changes, so review again ;P

ferblape:
Yes, you should clear the table, otherwise the user might think the query returns the same results, which is confusing.

jorgeatgu:
> Yes, you should clear the table, otherwise the user might think the query returns the same results, which is confusing.

Done! If the result is empty, we reset/cleaned the table.

ferblape:
Cool, thanks. Please review the conflicts now I've merged the other branch and let's review one last time before the merge 💪

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -414,7 +408,8 @@ export default {
       const {
         data: { data: items },
       } = response;
-      this.publicQueries = items;
+
+      this.publicQueries = !items ? [] : items
```
Crashillo: You may initialize in the previous line

```suggestion
const { data: { data: items = [] } } = response;
this.publicQueries = items
```

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -42,22 +46,36 @@ export default {
...
+      const data = items.trim()
+      //Only the columns of the query.
+      const arrayColumnsQueryString = this.arrayColumnsQuery.toString()
+
+      if (arrayColumnsQueryString !== data) {
+        this.initPerspective(this.items)
+      } else {
+        this.viewer.clear()
+        // Well, it's a bit tricky, but reset the table with .clear() only responds when trigger an event, if not trigger an event .clear() isn't fired
+        window.dispatchEvent(new Event('resize'))
```
Crashillo, line 71: I see, but why is required to _resize_? What for?

Crashillo (approved):

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
jorgeatgu: I tried it with some events like scroll, click, doublick, focus and doesn't works, but Perspective reacts/listened to the Resize event

Crashillo: Nop, my question is about why do we need an event? Only to make it work `this.viewer.clear()`? So you're saying that clause do nothing until you trigger an event? That's weird, so I asked.

PR #3053, opened 2020-05-11
Tenders and contracts lists pages for Dashboards module

eltercero:
Closes #2997

## :v: What does this PR do?

Integrates the markup done is #3028 with the remote csv data defined in Gobierto Dashboard settings. This PR enables several new pages for Contracts and Tenders dashboard in the main site:

- Summary
- Contracts index 
- Contract details
- Tenders index
- Tender details

Some parts, like the filters and the summary, are empty on purpose since this is still a work in progress.

Navigation should work between main parts (Summary ,Tenders and Contracts) and within the details page of every contract/tender. 

## :mag: How should this be manually tested?

It's available in staging: https://burjassot.gobify.net/dashboards/contratos/resumen

## :eyes: Screenshots

Captura de pantalla 2020-05-11 a las 12 09 32

Captura de pantalla 2020-05-11 a las 12 09 21

Captura de pantalla 2020-05-11 a las 12 09 42

## :shipit: Does this PR changes any configuration file?

- [ ] new environment variable in `.env.example`?
- [ ] new entry in `config/application.yml`?
- [ ] new entry in `config/secrets.yml`?

(Changes in these files might need to update the role in Ansible)

## :book: Does this PR require updating the documentation?

- [ ] new site configuration variable?
- [ ] new site template?
- [ ] new module/submodule settings?
- [ ] significant changes in some feature?

PR #3057, opened 2020-05-12
Summary main stats page for Dashboards

### LINKED ISSUE #3044: Contracts / Summary page main stats
Screenshot 2020-05-08 at 15 57 14

Integrate the main stats of the summary page. These stats (as in invoices) could be calculated in plain Javascript, taking into account the filters. For the moment we don't have active filters, but in the future filters will update these stats.

eltercero:
Closes #3044

## :v: What does this PR do?

It populates the summary page with the calculated data from the csv. 

This PR includes the updated code from #3053, so it may seem a lot of changes until that PR is merged.

It can be checked at: https://burjassot.gobify.net/dashboards/contratos/resumen

## :eyes: Screenshots

Captura de pantalla 2020-05-12 a las 17 11 42

- [ ] new site template?


app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -1,3 +1,215 @@
...
+    const sumContracts = d3.sum(amountsArray);
+    const meanContracts = d3.mean(amountsArray);
+    const medianContracts = d3.median(amountsArray);
+    const meanSavings = d3.mean(savingsArray);
+
+    // Calculations headlines
+    const lessThan1000Total = contractsData.filter((contract) => parseFloat(contract.final_amount) < 1000).length;
+    const lessThan1000Pct = lessThan1000Total/numberContracts;
+
+    const largerContractAmount = d3.max(contractsData, function(contract) { return parseFloat(contract.final_amount)} );
```
ferblape: `d3.max(contractsData, (contract) => parseFloat(contract.final_amount))`

app/javascript/gobierto_dashboards/webapp/lib/utils.js
```diff
@@ -0,0 +1,10 @@
+export function formatCurrency(amount) {
```
ferblape: No need to define a new function here, there's a filter for currencies in `app/javascript/lib/shared/modules/vue-filters.js`. Filters can be used in the HTML of a component like these {{ amount | money }}

app/javascript/gobierto_dashboards/webapp/containers/tender/TendersShow.vue
```diff
@@ -0,0 +1,66 @@
...
+    this.tender = this.tendersData.find((tender) => {
+      return tender.id === itemId
+    });
+
+    this.initFormattedTender();
+  },
+  methods: {
+    initFormattedTender(){
+      this.formattedTender = Object.assign({}, this.tender);
+      this.formattedTender.initial_amount = formatCurrency(this.formattedTender.initial_amount);
```
ferblape: See my comment about the money filter.

app/javascript/gobierto_dashboards/webapp/containers/tender/TendersShow.vue
```diff
@@ -0,0 +1,66 @@
...
+import { formatCurrency } from "../../lib/utils.js";
+
+export default {
+  name: 'TendersShow',
+  data() {
+    return {
+      tendersData: this.$root.$data.tendersData,
+      tender: null,
+      formattedTender: null,
+      labelTenderAmount: I18n.t('gobierto_dashboards.dashboards.contracts.tender_amount'),
```
ferblape, line 44: There's a filter also for translations, called `translate`, in `app/javascript/lib/shared/modules/vue-filters.js`

Instead of declaring data variables for translations I'd prefer to use it in the template like this:

`{{ 'gobierto_dashboards.dashboards.contracts.tender_amount' | translate }}`

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -1,3 +1,215 @@
...
+    document.getElementById("mean-contracts").innerText = meanContracts.toLocaleString(I18n.locale, {
+      style: 'currency',
+      currency: 'EUR'
+    });
+    document.getElementById("median-contracts").innerText = medianContracts.toLocaleString(I18n.locale, {
+      style: 'currency',
+      currency: 'EUR'
+    });
 
+    document.getElementById("mean-savings").innerText = meanSavings.toLocaleString(I18n.locale, {
```
ferblape, line 189: to format this amount can we use also the money filter as plain function

ferblape (commented):
I've added some details we can discuss later

app/javascript/gobierto_dashboards/webapp/containers/tender/TendersShow.vue
eltercero: That does not seems to work, though. I think `translate` is supposed to receive an Object and not a string?

ferblape: Hmmm, correct. I want to create at some point a filter to translate strings in the template and avoid so many variables in data (it happens too in Gobierto Data). We'll addresss this in the future.

Crashillo:
I'll review this once #3053 be merged

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -77,6 +80,10 @@ export class ContractsController {
           router,
           data: Object.assign(options, remoteCsvData),
         }).$mount(entryPoint);
+
+        EventBus.$on('summary_ready', () => {
+          _self._renderSummary(remoteCsvData);
+        });
```
Crashillo: No self/this.

```suggestion
        EventBus.$on('summary_ready', () => this._renderSummary(remoteCsvData) )
```

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -115,4 +122,76 @@ export class ContractsController {
...
   }
+
+  _renderSummary(remoteCsvData){
+    this._renderTendersMetricsBox(remoteCsvData.tendersData);
+    this._renderContractsMetricsBox(remoteCsvData.contractsData);
+  }
+
+  _renderTendersMetricsBox(tendersData){
+    // Calculations
+    const amountsArray = tendersData.map((tender) => parseFloat(tender.initial_amount) );
```
Crashillo: The point of destructuring if quite useful in this sort of situations:
```suggestion
    const amountsArray = tendersData.map(({ initial_amount = 0 }) => parseFloat(initial_amount) );
```

Unless you know (because you set it up before) that a variable can't come `undefined`, this approach is better to handle data that we don't know (i.e. comes from DB). If `tender` comes null or undefined, `parseFloat` won't never fail

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -115,4 +122,76 @@ export class ContractsController {
...
+
+    const numberTenders = tendersData.length;
+    const sumTenders = d3.sum(amountsArray);
+    const meanTenders = d3.mean(amountsArray);
+    const medianTenders = d3.median(amountsArray);
```
Crashillo, line 140: `import { sum, mean, median }` from 'd3-array'

imagen

app/javascript/gobierto_dashboards/webapp/containers/contract/ContractsShow.vue
```diff
@@ -1,18 +1,18 @@
 <template>
   <div>
-    <h1>{{ formattedContract.title }}</h1>
+    <h1>{{ contract.title }}</h1>
```
Crashillo: I notice you don't use `contract` object in any way. What if you declare each variable (title, description, ...) in the data() and only work with its properties?

app/javascript/gobierto_dashboards/webapp/containers/summary/Summary.vue
```diff
@@ -1,13 +1,113 @@
...
+    this.labelTenders = I18n.t('gobierto_dashboards.dashboards.contracts.summary.tenders'),
+    this.labelTendersFor = I18n.t('gobierto_dashboards.dashboards.contracts.summary.tenders_for'),
+    this.labelContracts = I18n.t('gobierto_dashboards.dashboards.contracts.summary.contracts'),
+    this.labelContractsFor = I18n.t('gobierto_dashboards.dashboards.contracts.summary.contracts_for'),
+    this.labelMeanAmount = I18n.t('gobierto_dashboards.dashboards.contracts.summary.mean_amount'),
+    this.labelMedianAmount = I18n.t('gobierto_dashboards.dashboards.contracts.summary.median_amount'),
+    this.labelMeanSavings = I18n.t('gobierto_dashboards.dashboards.contracts.summary.mean_savings'),
+    this.labelLessThan1000 = I18n.t('gobierto_dashboards.dashboards.contracts.summary.label_less_than_1000'),
+    this.labelLargerContractAmount = I18n.t('gobierto_dashboards.dashboards.contracts.summary.label_larger_contract_amount'),
+    this.labelHalfSpendingsContracts = I18n.t('gobierto_dashboards.dashboards.contracts.summary.label_half_spendings_contracts')
```
Crashillo: To data()

app/javascript/gobierto_dashboards/webapp/containers/tender/TendersShow.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <div>
-    <h1>{{ formattedTender.title }}</h1>
+    <h1>{{ tender.title }}</h1>
```
Crashillo, line 3: Some concept as Contracts. Is actually so different this component from the contracts one that we need another component?

config/locales/gobierto_dashboards/views/ca.yml
```diff
@@ -6,18 +6,34 @@ ca:
...
-        summary: Resum
+        summary:
+          contracts: Contractes adjudicats
+          contracts_for: contractes per import de
+          label_half_spendings_contracts: El <strong> <span id = "half-spendings-contracts-pct">
+            </ span> </ strong> de contractes concentren el 50% de tota la despesa
+          label_larger_contract_amount: El major contracte suposa un <strong> <span
+            id = "larger-contract-amount-pct"> </ span> </ strong> de tota la despesa
+            en contractes
+          label_less_than_1000: El <strong><span id="less-than-1000-pct"></span></strong>
```
Crashillo, line 29: HTML in a translation block?

Crashillo (changes requested):

app/javascript/gobierto_dashboards/modules/contracts_controller.js
eltercero: This is great, thanks!

app/javascript/gobierto_dashboards/webapp/containers/contract/ContractsShow.vue
eltercero: Will do!

eltercero: @Crashillo now I have this:

```javascript 
    this.title = contract.title
    this.description = contract.description
    this.assignee = contract.assignee
    this.document_number = contract.document_number
    this.final_amount = contract.final_amount
    this.initial_amount = contract.initial_amount
    this.status = contract.status
    this.process_type = contract.process_type
```

Is there a way to do it with destructuring?

app/javascript/gobierto_dashboards/webapp/containers/summary/Summary.vue
eltercero: Sorry, I forgot about this one!

app/javascript/gobierto_dashboards/webapp/containers/tender/TendersShow.vue
eltercero: Tenders page may not be in the final versión. If it finally is, I'll merge the logic.

config/locales/gobierto_dashboards/views/ca.yml
eltercero: Yeah, I thought about this one for a while. I've just realized that in Budget there's a suffix "_html" in the key for this kind of translations.  Is there a better way to approach this problem?

app/javascript/gobierto_dashboards/webapp/containers/contract/ContractsShow.vue
Crashillo: Two steps:
```
    const {
      title,
      description,
      assignee,
      document_number,
      final_amount,
      status,
      process_type
    } = contract

this.title = title
....
```

One step (perhaps more confusing)
```
    ({
      title: this.title,
      description: this.description,
      assignee: this.assignee,
      document_number: this.document_number,
      final_amount: this.final_amount,
      initial_amount: this.initial_amount
      status: this.status,
      process_type: this.process_type
    } = contract)
```

You might want to initialize them also.

eltercero: Thanks!

config/locales/gobierto_dashboards/views/ca.yml
Crashillo: I've seen the file. This way is quite a lot ofuscated, as you're adding an `id` to the block which you're gonna update that value later with a `document.getElementById`, is that right?
This is barely legible.
I would split the translations into smaller parts, to leave here just plain text

eltercero: So, spliting in smaller parts would make these translation keys. There are quotation marks because otherwise it'd be easy to miss the required whitespaces (although it can also be fixed through js).

```yaml 
label_half_spendings_contracts_1: "El "
label_half_spendings_contracts_2: " de contratos concentran el 50% de todo el gasto"
```

Later in vue I'd need to do this (triming just in case). 

```javascript
let arrayHalfSpendingsContracts = [
  I18n.t('gobierto_dashboards.dashboards.contracts.summary.label_half_spendings_contracts_1'),
  '<strong><span id="half-spendings-contracts-pct"></span></strong>',
  I18n.t('gobierto_dashboards.dashboards.contracts.summary.label_half_spendings_contracts_2')
]

this.labelHalfSpendingsContracts = arrayHalfSpendingsContracts.map((str) => str.trim()).join(' ')
```

What do you think? All in all, I'm not entirely sure this adds more clarity.

eltercero:
@ferblape @Crashillo I'm merging this PR if that's ok, there are two other PRs (#3063, #3075) that are depending on this one.

ferblape:
Ok, to me yes

Crashillo (approved):

config/locales/gobierto_dashboards/views/ca.yml
Crashillo: No, the goal is avoid messing with the logic/stuff. HTML in the template tag, javascript in the script tag, translation in the yml, and so on.
So this array is also mixing unalike stuff

PR #3059, opened 2020-05-13
3004 table custom fields from csv

### LINKED ISSUE #3004: Plans / Import custom field table CSV
- An admin will have a new option in the Import screen to import a CSV that will import data to a table plugin field custom. 
  - Add a heading "Plan content" just beneath the current INPUT FILE 
  - Add a new heading below: "Custom field (table plugin)" + input + submit buttton (submit buttons are independent, we need one for each import type)
- The syntax of the import CSV is the following: `external_id | custom_field | col_1 | col_2 | col_3 | col_4 | col_N`
  - The file may have a Node.title as the first column: it must be ignored
  - The system will search for a custom field with the slug defined in the header of the second column
  - If found, as a transaction:
    - values of that table for the corresponding external_id will be deleted (for those cases in which there is more current content than the content being imported)
    - col_1, col_2, etc. will be added in the same order

We have an example here https://docs.google.com/spreadsheets/d/1Fx1dP9IkpwXGD3TkDGnQWC_8W8mWLrSwCO9pAnySooE/edit#gid=461642504. We should import it in https://diba-plan.gobify.net/admin/plans/41/import_csv to check it works (custom field has to be created and header name updated)

entantoencuanto:
Closes #3004

## :v: What does this PR do?

Allows admins to upload custom field from a CSV according to description of the related issue

The CSV of the example in the issue has been uploaded after creating a custom field of type table with `campo-indicadores` uid and configuration:

```json
{
  "columns": [
    {
      "id": "name",
      "type": "text",
      "name_translations": {
        "ca": "Nom",
        "en": "Name",
        "es": "Nombre"
      }
    },
    {
      "id": "value",
      "type": "text",
      "name_translations": {
        "ca": "Valor",
        "en": "Value",
        "es": "Valor"
      }
    },
    {
      "id": "description",
      "type": "text",
      "name_translations": {
        "ca": "Descripció",
        "en": "Description",
        "es": "Descripción"
      }
    },
    {
      "id": "other",
      "type": "text",
      "name_translations": {
        "ca": "Otro texto",
        "en": "Other text",
        "es": "Otro texto"
      }
    }
  ]
}
```

The value field can be configured as type float, but some values as `0,5` would be converted to 0 and `10%` to 10

### Before this PR

🚫 

### After this PR

3004_after

No

Pending

### DIFF
--- a/app/controllers/gobierto_admin/gobierto_plans/plans_controller.rb
+++ b/app/controllers/gobierto_admin/gobierto_plans/plans_controller.rb
@@ -122,6 +122,7 @@ module GobiertoAdmin
       def import_csv
         @plan = ::GobiertoPlans::PlanDecorator.new(find_plan)
         @plan_data_form = PlanDataForm.new(plan: @plan)
+        @plan_table_custom_fields_form = PlanTableCustomFieldsForm.new(plan: @plan)
       end
 
       def export_csv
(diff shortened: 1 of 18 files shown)


ferblape:
> The value field can be configured as type float, but some values as 0,5 would be converted to 0 and 10% to 10

Why is this an issue? Why not `.tr(',', '').to_f`?

ferblape (changes requested):
I think you should deal properly with floats

entantoencuanto:
Changes added to parse better the strings:

* If a string contains more than a "," or at least a ".", the commas are removed and the value is transformed using `to_f`
* If the string contains only a "," is replaced by a "." and transformed using `to_f`

These changes affect also to the way a custom field of type numeric parses the values received

ferblape:
In https://diba-plan.gobify.net/admin/plans/41/import_csv I'm importing this CSV:

```
Node.Title,external_id,custom_field,objective,value_reached
Wadus,133,campo-indicadores,100,"10,3"
```

- the external ID is valid
- but the rest of the CSV I think it's incomplete. I don't see the project has been updated, and there's no message informing the users

Screenshot 2020-05-22 at 16 10 23

entantoencuanto:
A validation message should appear there. The CSV is expected to contain `col_1 | col_2 | col_3 ... col_N` headers for the table contents. In this way multiple tables can be loaded with the same CSV. Other names are ignored. The form ignores the `objective` and `value_reached` columns and do nothing

ferblape:
Do you mean in your comment that you are going to review it?

entantoencuanto:
Yes, i'll add an alert if no valid columns are present

entantoencuanto:
Fixed. Currently there was a validation error, but the error message were not being showed.

Also I've fixed the validation condition and added some translations.

I think now it's ready to merge

ferblape:
Normalize locales and merge

Screenshot 2020-05-28 at 07 14 41

PR #3065, opened 2020-05-15
Optimize fetch last activity date query

ferblape:
Today arrived a notifiication from AppSignal with this query, which is totally wrong, because doesn't use a `LIMIT`:

```
MESSAGE
PG::QueryCanceled: ERROR:  canceling statement due to statement timeout
: SELECT "activities"."created_at" FROM "activities" WHERE "activities"."site_id" = $1 AND (action ~* 'gobierto_budgets.budgets_updated') ORDER BY "activities"."created_at" ASC
```

Activity table has 280k records. For some sites, this query returns more than 1000 records, when we just need one.

The table also lacks of two basic indexes: created_at and action.

## :v: What does this PR do?

- Fixes all the wrong queries I've found in budgets, plans, indicators.

- Adds the missing index. You can see the sort key index and the filter index:

```
                                                   QUERY PLAN
----------------------------------------------------------------------------------------------------------------
 Limit  (cost=1.28..1.28 rows=1 width=8) (actual time=0.082..0.082 rows=1 loops=1)
   ->  Sort  (cost=1.28..1.28 rows=1 width=8) (actual time=0.081..0.081 rows=1 loops=1)
         Sort Key: created_at DESC
         Sort Method: top-N heapsort  Memory: 25kB
         ->  Seq Scan on activities  (cost=0.00..1.27 rows=1 width=8) (actual time=0.033..0.035 rows=5 loops=1)
               Filter: ((site_id = 1009469985) AND ((action)::text = 'gobierto_budgets.budgets_updated'::text))
               Rows Removed by Filter: 13
 Planning Time: 9.518 ms
 Execution Time: 0.141 ms
```

- Updated dates in indicators, budgets and plans shouldn't change

### DIFF
--- a/app/controllers/gobierto_admin/gobierto_calendars/calendar_configuration_controller.rb
+++ b/app/controllers/gobierto_admin/gobierto_calendars/calendar_configuration_controller.rb
@@ -71,8 +71,10 @@ module GobiertoAdmin
       def set_last_sync
         @last_sync = if collection_container = @calendar_configuration_form.try(:collection_container)
                        current_site.activities.where(action: 'admin_gobierto_calendars.calendars_synchronized', subject: collection_container)
-                         .order(created_at: :asc)
-                         .last.try(:created_at)
+                         .order(created_at: :desc)
+                         .limit(1)
+                         .pluck(:created_at)
+                         .first
                      end
       end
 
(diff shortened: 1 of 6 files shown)


ferblape:
Link to app signal https://appsignal.com/populate/sites/5ebb88520264445dc7e68b5c/exceptions/incidents/113?timestamp=2020-05-15T04%3A00%3A52Z

eltercero (approved):
Todo correcto.

PR #3081, opened 2020-05-19
Hide custom field records when not sent by the API

### LINKED ISSUE #3080: Avoid js errors in plans with not existing custom field records
In the template of the plan an error is produced trying to get the length of a null value. Let's hide the value instead

entantoencuanto:
Closes #3080
Closes PopulateTools/issues#1032
Closes PopulateTools/issues#1033

## :v: What does this PR do?

* Avoids errors trying to display custom field contents not provided by the API. When a custom field is not provided by the API the field is hidden in the front
* Allows to remove the value of a numeric custom field record in admin and set the custom field value to nil
* Adds 2 new configuration params on plans:
  * `fields_to_not_show_in_front`: specify a list of field UIDs which we should not expose in the API (the client wants to store the info but not show it on the front). Default: `[]`
  * `show_empty_fields`: be able to decide if fields with no value should be shown. Default: `false`

Visit a project without records for a custom field of type text (this can be achieved by creating a new custom field of type paragraph and not entering to edit the projects from admin) 

No

Sent suggestion to document configuration params

### DIFF
--- a/app/controllers/gobierto_plans/plan_types_controller.rb
+++ b/app/controllers/gobierto_plans/plan_types_controller.rb
@@ -58,6 +58,7 @@ module GobiertoPlans
       @sdgs = SdgDecorator.new(find_plan)
       @sdg = @sdgs.sdg_term(params[:sdg_slug])
       @projects = @sdgs.projects_by_sdg(@sdg)
+      @projects_term = @plan.level_key(2, @plan.levels + 1)
     end
 
     private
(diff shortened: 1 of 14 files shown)


furilo:
Let's also use the "projects" literal configured in this view

https://diba-plan.gobify.net/planes/pla-de-mandat/2019/ods/4

image

entantoencuanto:
Ups, you're right!

entantoencuanto:
Changes made and deployed to staging

ferblape:
Doing some testing in https://diba-plan.gobify.net/admin/plans/41/edit and 👍

PR #3084, opened 2020-05-20
Enable 5 runners for CircleCI

ferblape:
## :v: What does this PR do?

Sets 5 runners in CircleCI instead of four. Let's see the improvement


ferblape:
Not a good idea, the rest of builds are blocked waiting... It'll make the runner slow

PR #3090, opened 2020-05-21
Dashboards Contracts sidebar filters

### LINKED ISSUE #3047: Contracts / Sidebar filters
Screenshot 2020-05-08 at 16 12 00

Sidebar filters should reuse the same components to build Investments and Data modules. 

They should behave in the same way: don't show the filter if there's no data, unchecking an option should show again all the filters, etc

When a filter is activated, it should notify dc charts about the change. I thought in implemting it via events, but I'm open to suggestions.

eltercero:
Closes #3047

## :v: What does this PR do?

It adds the sidebar filters to Dashboards Contracts.

@Crashillo I think this was the first time that Dropdown and Blockheader are being used together, and there were a couple of bugs. The first is a matter of styling, I added a new rule `widht: 100%` in `app/assets/stylesheets/comp-block-header.scss`. Otherwise the "All" link and the blockheader title would stay together (even with the flex and justify-between rules). The other bug was that when clicking "All" would also work as a toggle and collapse. The fix, which @jorgeatgu figured out, was adding the event modifier `stop` (as in `@click.stop="selectAll"`). Now it works.

What's still pending to do is using the Caret component in Blockheader and make it rotate accordingly.

### Some implementation details

I'd like to share some details so you can better understand the reasons behind some parts of the code. 

The most important thing you need to know is that all the filtering in the sidebars is actually made by the dc charts, so there's no filtering logic in the vue components.

To make this work, the dc charts have to be initialized no matter the page you are or start at. So, for instance, if you enter `/dashboards/contratos/contratos` the dc charts (which are not in that component) have to be created because you could filter from the sidebar there. 

Another issue that arised is that, since they could be created without the component being shown, when going to Summary the charts didn't have the proper size (the size is calculated when creating the chart using the size of the target element). So there's an event called `moved_to_summary` that will be listened just once to redraw the charts when going to Summary from others . 

Last, there's a sidebar filter for dates. To make this work, there's a hidden dc chart for dates that does all the filtering with the other charts.

Head to https://getafe.gobify.net/dashboards/contratos/resumen and play with the sidebar filters. Case scenarios:

- Clicking on any sidebar filter will update the current filters. It should also update the chart bars if the filter has a char version.
- When filtering with the chart bar, the sidebar filter should update if it applies.
- If there's a version of the filter as a chart, the chart filter should be updated too.
- If you're using the sidebar filter from any index (contracts or tenders), listing should be updated

## :eyes: Screenshots

Captura de pantalla 2020-05-21 a las 10 51 40

No

No

### DIFF
--- a/app/javascript/gobierto_dashboards/modules/contracts_controller.js
+++ b/app/javascript/gobierto_dashboards/modules/contracts_controller.js
@@ -6,7 +6,6 @@ import { scaleThreshold } from 'd3-scale';
 
 const d3 = { scaleThreshold, sum, mean, median, max }
 
-import * as dc from 'dc'
 import crossfilter from 'crossfilter2'
 
 import { getRemoteData } from '../webapp/lib/utils'
--- a/app/javascript/gobierto_dashboards/modules/contracts_controller.js
+++ b/app/javascript/gobierto_dashboards/modules/contracts_controller.js
@@ -327,7 +327,7 @@ export class ContractsController {
 
   _formalizedContractsData(contractsData){
     return contractsData.filter(({status}) =>
-      status == 'Formalizado' || status == 'Adjudicado'
+      status === 'Formalizado' || status === 'Adjudicado'
     )
   }
 
--- a/app/javascript/gobierto_dashboards/modules/contracts_controller.js
+++ b/app/javascript/gobierto_dashboards/modules/contracts_controller.js
@@ -96,6 +96,11 @@ export class ContractsController {
           this._updateChartsFromFilter(options);
         });
 
+        // - dc charts are drawn even if the summary page is not the page where the user lands for the first time
+        //   i.e.: they could first land on a contract page, but still this page would need to render for the filters to work
+        // - dc charts sizes are calculated automatically, but if the page is not visible it won't calculate sizes properly
+        // - Given all that: when we go from a page that is not summary to summary for the first time, the sizes must
+        //   be calculated and the charts redrawn. This is why this event only needs to be listened once.
         EventBus.$once('moved_to_summary', () => {
           this._redrawCharts();
         });
--- a/app/javascript/gobierto_dashboards/modules/contracts_controller.js
+++ b/app/javascript/gobierto_dashboards/modules/contracts_controller.js
@@ -200,16 +200,12 @@ export class ContractsController {
     // Calculations
     const amountsArray = _contractsData.map(({final_amount = 0}) => parseFloat(final_amount) );
     const sortedAmountsArray = amountsArray.sort((a, b) => b - a);
-    const savingsArray = _contractsData.map(({initial_amount = 0, final_amount = 0}) =>
-      1 - (parseFloat(final_amount) / parseFloat(initial_amount))
-    );
 
     // Calculations box items
     const numberContracts = _contractsData.length;
     const sumContracts = d3.sum(amountsArray);
     const meanContracts = d3.mean(amountsArray);
     const medianContracts = d3.median(amountsArray);
-    const meanSavings = d3.mean(savingsArray);
 
     // Calculations headlines
     const lessThan1000Total = _contractsData.filter(({final_amount = 0}) => parseFloat(final_amount) < 1000).length;
@@ -233,10 +229,6 @@ export class ContractsController {
     document.getElementById("mean-contracts").innerText = money(meanContracts);
     document.getElementById("median-contracts").innerText = money(medianContracts);
 
-    document.getElementById("mean-savings").innerText = meanSavings.toLocaleString(I18n.locale, {
-      style: 'percent',
-      minimumFractionDigits: 2
-    });
     document.getElementById("less-than-1000-pct").innerText = lessThan1000Pct.toLocaleString(I18n.locale, {
       style: 'percent'
     });
--- a/app/javascript/gobierto_dashboards/modules/contracts_controller.js
+++ b/app/javascript/gobierto_dashboards/modules/contracts_controller.js
@@ -6,6 +6,7 @@ import { scaleThreshold } from 'd3-scale';
 
 const d3 = { scaleThreshold, sum, mean, median, max }
 
+import * as dc from 'dc'
 import crossfilter from 'crossfilter2'
 
 import { getRemoteData } from '../webapp/lib/utils'
@@ -19,7 +20,7 @@ Vue.use(VueRouter);
 Vue.config.productionTip = false;
 
 // Global variables
-let data, reduced, ndx, _r, vueApp;
+let data, reduced, ndx, _amountRange, vueApp, charts = {};
 
 export class ContractsController {
   constructor(options) {
@@ -40,48 +41,48 @@ export class ContractsController {
       const TendersIndex = () => import("../webapp/containers/tender/TendersIndex.vue");
       const TendersShow = () => import("../webapp/containers/tender/TendersShow.vue");
 
-      const router = new VueRouter({
-        mode: "history",
-        routes: [
-          { path: "/dashboards/contratos", component: Home,
-            children: [
-              { path: "resumen", name: "summary", component: Summary },
-              { path: "contratos", name: "contracts_index", component: ContractsIndex },
-              { path: "contratos/:id", name: "contracts_show", component: ContractsShow },
-              { path: "licitaciones", name: "tenders_index", component: TendersIndex },
-              { path: "licitaciones/:id", name: "tenders_show", component: TendersShow },
-            ]
-          }
-
-        ],
-        scrollBehavior() {
-          const element = document.getElementById(selector);
-          window.scrollTo({ top: element.offsetTop, behavior: "smooth" });
-        }
-      });
-
-      const baseTitle = document.title;
-      router.afterEach(to => {
-        // Wait 2 ticks
-        Vue.nextTick(() =>
-          Vue.nextTick(() => {
-            let title = baseTitle;
-
-            if (to.name === "contracts_show" || to.name === "tenders_show") {
-              const { item: { title: itemTitle } = {} } = to.params;
-
-              if (itemTitle) {
-                title = `${itemTitle}${baseTitle}`;
-              }
-            }
-
-            document.title = title;
-          })
-        );
-      });
-
       Promise.all([getRemoteData(options.contractsEndpoint), getRemoteData(options.tendersEndpoint)]).then((rawData) => {
         this.setGlobalVariables(rawData)
+
+        const router = new VueRouter({
+          mode: "history",
+          routes: [
+            { path: "/dashboards/contratos", component: Home,
+              children: [
+                { path: "resumen", name: "summary", component: Summary},
+                { path: "contratos", name: "contracts_index", component: ContractsIndex },
+                { path: "contratos/:id", name: "contracts_show", component: ContractsShow },
+                { path: "licitaciones", name: "tenders_index", component: TendersIndex },
+                { path: "licitaciones/:id", name: "tenders_show", component: TendersShow },
+              ]
+            }
+
+          ],
+          scrollBehavior() {
+            const element = document.getElementById(selector);
+            window.scrollTo({ top: element.offsetTop, behavior: "smooth" });
+          }
+        });
+
+        const baseTitle = document.title;
+        router.afterEach(to => {
+          // Wait 2 ticks
+          Vue.nextTick(() =>
+            Vue.nextTick(() => {
+              let title = baseTitle;
+
+              if (to.name === "contracts_show" || to.name === "tenders_show") {
+                const { item: { title: itemTitle } = {} } = to.params;
+
+                if (itemTitle) {
+                  title = `${itemTitle}${baseTitle}`;
+                }
+              }
+
+              document.title = title;
+            })
+          );
+        });
 
         vueApp = new Vue({
           router,
@@ -90,6 +91,14 @@ export class ContractsController {
 
         EventBus.$on('summary_ready', () => {
           this._renderSummary();
+        });
+
+        EventBus.$on('filter_changed', (options) => {
+          this._updateChartsFromFilter(options);
+        });
+
+        EventBus.$once('moved_to_summary', () => {
+          this._redrawCharts();
         });
       });
     }
@@ -122,12 +131,12 @@ export class ContractsController {
       }
     }
 
-    // Contracts precalculations
-    _r = {
+    // Contracts precalculations and normalizations
+    _amountRange = {
       domain: [501, 1001, 5001, 10001, 15001],
       range: [0, 1, 2, 3, 4, 5]
     };
-    var rangeFormat = d3.scaleThreshold().domain(_r.domain).range(_r.range);
+    var rangeFormat = d3.scaleThreshold().domain(_amountRange.domain).range(_amountRange.range);
 
     for(let i = 0; i < contractsData.length; i++){
       const contract = contractsData[i];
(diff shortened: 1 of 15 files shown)


eltercero:
@ferblape @Crashillo ready for review

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -6,6 +6,7 @@ import { scaleThreshold } from 'd3-scale';
 
 const d3 = { scaleThreshold, sum, mean, median, max }
 
+import * as dc from 'dc'
```
Crashillo: Is there any chance to code splitting `dc`?

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -137,23 +146,25 @@ export class ContractsController {
       contract.final_amount = final_amount;
       contract.initial_amount = initial_amount;
       contract.range = rangeFormat(+final_amount);
+      contract.start_date_year = contract.start_date != undefined && contract.start_date != '' ? (new Date(contract.start_date).getFullYear()) : contract.start_date;
```
Crashillo: To check for `undefined` and length, you can leave it as:

```suggestion
      contract.start_date_year = contract.start_date ? (new Date(contract.start_date).getFullYear()) : contract.start_date;
```
Undefineds, nulls, and empty lengths are evaluated to `false`

Ay, ay, ay, that destructuring...

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -258,10 +263,13 @@ export class ContractsController {
...
       dimension: dimension,
-      onFilteredFunction: () => this._refreshData(dimension.top(Infinity))
+      onFilteredFunction: (chart, filter) => {
+        this._refreshData(dimension.top(Infinity))
+        EventBus.$emit('dc_filter_selected', {title: filter, id: 'contract_types'})
+      }
     }
 
-    new GroupPctDistributionBars(renderOptions);
+    charts['contract_types'] = new GroupPctDistributionBars(renderOptions);
```
Crashillo, line 276: I'm seeing you're declaring this file as a `Class`, then, why do you use *global* variables? Declara them in the constructor, and use them with `this`

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -270,19 +278,58 @@ export class ContractsController {
...
+    if (options.all) {
+      container.filter(null);
+      container.filter([options.titles]);
+    } else {
+      container.filter(options.title);
     }
 
-    new GroupPctDistributionBars(renderOptions);
+    Object.values(charts).forEach((chart) => chart.container.redraw());
+  }
```
Crashillo, line 320: I hope `options` brings all these properties.... 😇

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -270,19 +278,58 @@ export class ContractsController {
...
   _currentDataSource(){
     return reduced || data
   }
 
-  _formalizedContracts(){
-    return this._currentDataSource().contractsData.filter(({status}) =>
-      status === 'Formalizado' || status === 'Adjudicado'
+  _formalizedContractsData(contractsData){
+    return contractsData.filter(({status}) =>
+      status == 'Formalizado' || status == 'Adjudicado'
```
Crashillo: What? What? What? Removing a best practice? OMG.... Goku, kill him!

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,138 @@
 <template>
   <div class="pure-u-1 pure-u-lg-1-4">
     <aside class="dashboards-home-aside--gap">
-      wip filtros
+      <div
+        v-for="filter in filters"
+        :key="filter.id"
+        class="dashboards-home-aside--block"
+      >
+        <Dropdown @is-content-visible="filter.isToggle = !filter.isToggle">
```
Crashillo: Are we using isToggle somewhere?

Crashillo: Maybe in `@toggle=toggle(filter)`? Plz, confirm

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,138 @@
...
+
+    EventBus.$on('dc_filter_selected', ({title, id}) => {
+      const filter = this.filters.find(filter => filter.id === id)
+
+      filter.options.forEach(option => {
+        if (option.title === title) {
+          option.isOptionChecked = !option.isOptionChecked;
+        }
+      })
+    });
```
Crashillo: What if the `Aside` parent component is who listen the events and pass the props down? Asking

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,138 @@
...
+      this.contractsData.forEach((contract) => {
+        counter.process_types[contract.process_type] = counter.process_types[contract.process_type] || 0
+        counter.process_types[contract.process_type]++
+
+        counter.contract_types[contract.contract_type] = counter.contract_types[contract.contract_type] || 0
+        counter.contract_types[contract.contract_type]++
+
+        counter.dates[contract.start_date_year] = counter.dates[contract.start_date_year] || 0
+        counter.dates[contract.start_date_year]++
+      })
```
Crashillo: I've seen this before....

app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
```diff
@@ -1,17 +1,19 @@
...
         ></Nav>
         <main class="dashboards-home-main">
-          <keep-alive include="Summary">
-            <router-view :key="$route.fullPath"></router-view>
-          </keep-alive>
+          <Summary v-show="isCurrentPath('summary')"/>
+          <ContractsIndex v-show="isCurrentPath('contracts_index')"/>
+          <ContractsShow v-if="isCurrentPath('contracts_show')"/>
+          <TendersIndex v-show="isCurrentPath('tenders_index')"/>
+          <TendersShow v-if="isCurrentPath('tenders_show')"/>
```
Crashillo: It isn't a good practice adding function calls in the DOM. Every single time the component is updated, therefore, rendered; is calling to these functions. That penalizes the performance.

You should declare data properties or computed functions. Due to the name you're using, I'm assuming `isCurrentPath` runs at mounted, and stills.
If so, declare boolean flags to handle this.

app/javascript/gobierto_dashboards/webapp/lib/config.js
```diff
@@ -16,3 +16,35 @@ export const assigneesColumns = [
...
+    id: 'process_types',
+    title: I18n.t('gobierto_dashboards.dashboards.contracts.process_type'),
+    isToggle: true,
+    options: [
+        {id: 0, title: 'Abierto', counter: 0, isOptionChecked: false},
+        {id: 1, title: 'Abierto simplificado', counter: 0, isOptionChecked: false},
+        {id: 2, title: 'Basado en Acuerdo Marco', counter: 0, isOptionChecked: false},
+        {id: 3, title: 'Negociado con publicidad', counter: 0, isOptionChecked: false},
+        {id: 4, title: 'Negociado sin publicidad', counter: 0, isOptionChecked: false}
+      ]
```
Crashillo: Untraslated stuff? Oh lord

Crashillo (changes requested):

app/javascript/gobierto_dashboards/modules/contracts_controller.js
eltercero: It's actually not needed here, I forgot to remove it. I'll delete it.

app/javascript/gobierto_dashboards/modules/contracts_controller.js
eltercero: Ahh, that's why there was a git clonflict there 😅

app/javascript/gobierto_dashboards/webapp/lib/config.js
eltercero: That's the title coming from the csv. I did it this way so make things simple, but maybe I should generate it automatically based on the csv :\

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
eltercero: It was `@is-content-visible="toggle(filter)"`. Thanks!

app/javascript/gobierto_dashboards/webapp/lib/config.js
ferblape: @eltercero yes, I think you should make these lists dynamic from the CSV, there might be more types

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
eltercero: In the first case I think it makes sense, I changed it. I don't think it'd work well for the second case, though, it'll be more complex.

app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
eltercero: Changed to computed properties 👍

app/javascript/gobierto_dashboards/webapp/lib/config.js
eltercero: Changed!

app/javascript/gobierto_dashboards/modules/contracts_controller.js
eltercero: Yeah, you're right. I'll do it in the latest branch to avoid conflicts, since this file has a lot of changes.

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -91,6 +91,14 @@ export class ContractsController {
...
           this._renderSummary();
         });
+
+        EventBus.$on('filter_changed', (options) => {
+          this._updateChartsFromFilter(options);
+        });
+
+        EventBus.$once('moved_to_summary', () => {
+          this._redrawCharts();
+        });
```
Crashillo: Try to include an explanatory comment for each event. Specially the events. You may do it in following PRs

Crashillo: 🤔 Indeed, why the last event is `$once` for? I mean, I know what `once` does, but what purpose?

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,169 @@
...
+    this.updateCounters(true);
+
+    EventBus.$on('dc_filter_selected', ({title, id}) => {
+      const filter = this.filters.find(filter => filter.id === id)
+
+      filter.options.forEach(option => {
+        if (option.title === title) {
+          option.isOptionChecked = !option.isOptionChecked;
+        }
+      })
```
Crashillo: High error possibility

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,169 @@
...
+    });
+  },
+  watch: {
+    contractsData: function (newContractsData, oldContractsData) {
+      this.updateCounters();
+    }
+  },
+  methods: {
+    initFilterOptions(){
+      const contractTypesOptions = [], processTypesOptions = [], dateOptions = [];
```
Crashillo: Different lines

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,169 @@
...
+          filter.options = dateOptions;
+        } else if (filter.id === 'contract_types') {
+          filter.options = contractTypesOptions;
+        } else if (filter.id === 'process_types') {
+          filter.options = processTypesOptions;
+        }
+      })
+    },
+    updateCounters(firstUpdate=false) {
+      const counter = {process_types: {}, contract_types: {}, dates: {}};
```
Crashillo, line 125: Different lines

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,169 @@
...
+
+      debugger
```
Crashillo: Hello?

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,169 @@
...
+      this.contractsData.forEach(({process_type, contract_type, start_date_year}) => {
+        counter.process_types[process_type] = counter.process_types[process_type] || 0
+        counter.process_types[process_type]++
+
+        counter.contract_types[contract_type] = counter.contract_types[contract_type] || 0
+        counter.contract_types[contract_type]++
+
+        counter.dates[start_date_year] = counter.dates[start_date_year] || 0
+        counter.dates[start_date_year]++
+      })
```
Crashillo, line 139: Explain this function

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -1,14 +1,169 @@
...
+
+      EventBus.$emit('filter_changed', {all: true, titles: titles, id: filter.id});
+    },
+    handleCheckboxStatus({ id, value, filter }) {
+      const option = filter.options.find(option => option.id === id)
+      EventBus.$emit('filter_changed', {all: false, title: option.title, id: filter.id});
+    },
+    toggle(filter){
+      this.filters.forEach(_filter => {
+        if (_filter.id == filter.id) {
```
Crashillo: ===

app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
```diff
@@ -21,25 +25,56 @@
...
       activeTabIndex: store.state.currentTab || 0,
+      contractsData: this.$root.$data.contractsData,
     }
   },
+  computed: {
+    isSummary: function() { return this.$route.name === 'summary' },
+    isContractsIndex: function() { return this.$route.name === 'contracts_index' },
+    isContractsShow: function() { return this.$route.name === 'contracts_show' },
+    isTendersIndex: function() { return this.$route.name === 'tenders_index' },
+    isTendersShow: function() { return this.$route.name === 'tenders_show' },
```
Crashillo: 1. ES6
2. Are these values changing while the component is mounted? It looks like they don't

Crashillo (changes requested):

app/javascript/gobierto_dashboards/modules/contracts_controller.js
eltercero: I'll add an explanation for that.

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
eltercero: What do you suggest?

app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
eltercero: 1. I think you can't use arrow functions in computed properties, or do you mean something else?

2. I think they are changing. Home is mounted just once and every time the tab changes these properties change too.

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
Crashillo:
```suggestion
      const { options = [] } = this.filters.find(( { id: i } ) => id === i) || {};

      options.forEach(option => {
        if (option.title === title) {
          option.isOptionChecked = !option.isOptionChecked;
        }
      })
```

eltercero: 👏 thanks, it's much better and it works great.

app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
Crashillo: I'm just saying so, keep consistency in the code format

```suggestion
    isSummary() { return this.$route.name === 'summary' },
    ...etc
```

Crashillo (approved):

app/javascript/gobierto_dashboards/webapp/containers/shared/Home.vue
eltercero: Ahh. Understood, thanks!

PR #3092, opened 2020-05-21
Dashboard open data link

### LINKED ISSUE #3077: Dashboards / Add link to open data
We are going to include a reference to open data in dashboards ([in context](https://www.figma.com/proto/gyMSROblmDqjkKt1TT8BcL5G/Gobierto-%2F-Contratos-Subvenciones?node-id=464%3A0&viewport=607%2C-881%2C0.25&scaling=min-zoom)): 

image

Maybe we can extend the configuration options and add something like: "data_download_source: URL", and the URL can be internal or external. 

If option is present we show the box, otherwise we ignore it.

eltercero:
Closes #3077

## :v: What does this PR do?

Enables the option to define a data source in the settings like:

```json 
    {
      "dashboards_config": {
        "dashboards": {
          "contracts": {
            "enabled": true,
            "data_urls": {},
            "data_download_source": "http://data.source.url"
          }
        }
      }
    }
```

If added, a download button will appear (see screenshot).

I added the option in Staging, you can check it in the summary page: 

https://getafe.gobify.net/dashboards/contratos/resumen

If removed, it shouldn't appear.

## :eyes: Screenshots

Captura de pantalla 2020-05-21 a las 15 20 20

No

No

### DIFF
--- a/app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
+++ b/app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
@@ -150,6 +150,7 @@ export default {
           filter.options = filter.options.sort((a, b) => a.counter > b.counter ? -1 : 1)
         }
       })
+
     },
     handleIsEverythingChecked({ filter }) {
       const titles = filter.options.map(option => option.title);
--- a/app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
+++ b/app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
@@ -30,6 +30,17 @@
           </div>
         </Dropdown>
       </div>
+
+      <div
+        v-if="showDownloadDataLink()"
+        class="dashboards-home-aside--download-open-data"
+      >
+        <i class="fas fa-table"></i>
+        <a :href="dataDownloadEndpoint" target="blank">
+          {{ labelDownloadData }}
+        </a>
+      </div>
+
     </aside>
   </div>
 </template>
@@ -48,6 +59,8 @@ export default {
   },
   data() {
     return {
+      labelDownloadData: I18n.t('gobierto_dashboards.dashboards.contracts.download_data'),
+      contractsData: this.$root.$data.contractsData,
       filters: filtersConfig
     }
   },
@@ -55,6 +68,10 @@ export default {
     contractsData: {
       type: Array,
       default: {}
+    },
+    dataDownloadEndpoint: {
+      type: String,
+      default: ''
     }
   },
   created(){
@@ -168,6 +185,9 @@ export default {
           _filter.isToggle = !_filter.isToggle;
         }
       })
+    },
+    showDownloadDataLink(){
+      return(this.dataDownloadEndpoint != '' && this.dataDownloadEndpoint != undefined)
     }
   }
 }
--- a/app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
+++ b/app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
@@ -32,7 +32,7 @@
       </div>
 
       <div
-        v-if="showDownloadDataLink()"
+        v-if="isShowDownloadDatalinkShown"
         class="dashboards-home-aside--download-open-data"
       >
         <i class="fas fa-table"></i>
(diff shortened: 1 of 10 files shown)


eltercero:
@ferblape @Crashillo this is now ready to review 👍

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -30,6 +30,17 @@
           </div>
         </Dropdown>
       </div>
+
+      <div
+        v-if="showDownloadDataLink()"
+        class="dashboards-home-aside--download-open-data"
+      >
+        <i class="fas fa-table"></i>
+        <a :href="dataDownloadEndpoint" target="blank">
```
Crashillo: is not `target="_blank"` ?

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -167,6 +185,9 @@ export default {
           _filter.isToggle = !_filter.isToggle;
         }
       })
+    },
+    showDownloadDataLink(){
+      return(this.dataDownloadEndpoint != '' && this.dataDownloadEndpoint != undefined)
```
Crashillo: 1. ejem ejem.... `!==`
2. Look what are you doing: firstly, you check if not empty, then check if not undefined. That's nonsense. If the first condition pass, the second one also will. If you'd have to, you'd have to do it backwards
3. Coolest way: `return !!this.dataDownloadEndpoint`

config/locales/gobierto_dashboards/views/ca.yml
```diff
@@ -13,6 +13,7 @@ ca:
         dates: Dates
         description: Visualització i anàlisi dels contractes adjudicats i les licitacions
           del %{entity_name}
+        download_data: Descarregueu les dades en format reutilitzable
```
Crashillo, line 16: Descàrregueu...

Crashillo (changes requested):

config/locales/gobierto_dashboards/views/ca.yml
ferblape: Not correct,  no es esdrújula si no plan

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
ferblape: I was going to ask the same

ferblape:
This is linking to Gobierto data api but in the future it's going to be a CSV file, so the download will make more sense

config/locales/gobierto_dashboards/views/ca.yml
Crashillo: Sorry, es de ver *Descàrrega*

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
```diff
@@ -30,6 +30,17 @@
           </div>
         </Dropdown>
       </div>
+
+      <div
+        v-if="showDownloadDataLink()"
```
Crashillo: Giving another look at this, it looks weird to me.
So, you're calling a function in the DOM (bad practice) to get a Boolean. This boolean is about `dataDownloadEndpoint` whether exists or not.
`dataDownloadEndpoint` is a prop, and it has a default value, so always gonna exist, at least empty. 
When does `dataDownloadEndpoint` change? I presume never in the life. In such case, why is not declarated in `created` lifecycle? I'm saying:

```
created() {
   this.isShowDownloadDatalinkShown = !!dataDownloadEndpoint
}
```

End of history.

What if the `dataDownloadEndpoint` prop changes?

```
watch: {
  dataDownloadEndpoint(old, new) {
    if (old !== new) {
      this.isShowDownloadDatalinkShown = !!new
    }
  }
}
```

But I don't think that be the case, but keep in mind for future approachs

Crashillo (changes requested):

app/javascript/gobierto_dashboards/webapp/containers/shared/Aside.vue
eltercero: Yeah, makes much more sense. I changed it (added the watcher too).

Crashillo (approved):
Adding the watcher is only necessary if the props gonna change. Keep in mind

eltercero:
> Adding the watcher is only necessary if the props gonna change. Keep in mind

Uhm, ok, I'll delete it then. I don't think it will change. Thanks.

PR #3093, opened 2020-05-21
Order by site name in the admin dropdown and menu CSS adjustment

furilo:
Closes unexpected

## :v: What does this PR do?

- Order the site list in the admin dropdown by site name (the org), and adds the site title
- Enables scroll if the dropdown list height is greater than the viewport, and a max-height for the container
- It adds inline styles, which may come as an abobination, but since I'm becoming a fan of Tailwind and the likes, and this is only going to be used here, it makes sense to me ;)

- [ ] new site template?

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -109,7 +109,7 @@ class Site < ApplicationRecord
   translates :name, :title
 
   def self.alphabetically_sorted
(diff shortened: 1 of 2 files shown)


ferblape:
Deployed to staging

PR #3097, opened 2020-05-22
Dashboards / small details pending

eltercero:
Closes PopulateTools/issues#1039

## :v: What does this PR do?

Several fixues related to literals, styling, routes changes, etc. Please refer to the issue for the specifics.

@Crashillo I added a class because table column needed to have more width. What I did may not be the way it's usually done, so please take a look and let me know if there's a better way. Thanks!

The summary and contracts urls have both changed, so please go here to start browsing the module: 

https://burjassot.gobify.net/dashboards/contratos/

No

No


ferblape:
Careful, I think you need to run i18-tasks normalize. There's a failing test about that

eltercero:
> Careful, I think you need to run i18-tasks normalize. There's a failing test about that

you were right :) should be fixed now.

Crashillo:
Shouldn't we merge the `husky` PR about that?

config/locales/gobierto_dashboards/views/ca.yml
```diff
@@ -23,9 +25,9 @@ ca:
           contracts: Contracts
           summary: Resum
           tenders: Licitacions
-        permalink: enllaç cap a la web oficial
+        permalink: Consulta aquest contracte en la Plataforma de Contractació de l'Sector Públic
```
ferblape: Consulta aquest contracte en la Plataforma de Contractació del Sector Públic

config/locales/gobierto_dashboards/views/en.yml
```diff
@@ -1,6 +1,8 @@
 ---
 en:
   gobierto_dashboards:
+    shared:
+      loading: Carregant dades...
```
ferblape: Loading data

config/locales/gobierto_dashboards/views/es.yml
```diff
@@ -1,6 +1,8 @@
 ---
 es:
   gobierto_dashboards:
+    shared:
+      loading: Loading data...
```
ferblape: Cargando datos 😆

ferblape (changes requested):
Small details, fix them and 👍

config/locales/gobierto_dashboards/views/es.yml
eltercero: 🙄

eltercero:
@Crashillo ready to review ✌️

app/javascript/gobierto_dashboards/webapp/lib/config.js
```diff
@@ -1,20 +1,20 @@
 export const contractsColumns = [
-  {field: 'assignee', translation: I18n.t('gobierto_dashboards.dashboards.contracts.assignee'), format: null},
-  {field: 'title', translation: I18n.t('gobierto_dashboards.dashboards.contracts.contractor'), format: 'truncated'},
-  {field: 'final_amount_no_taxes', translation: I18n.t('gobierto_dashboards.dashboards.contracts.final_amount_no_taxes'), format: 'currency'},
-  {field: 'start_date', translation: I18n.t('gobierto_dashboards.dashboards.contracts.date'), format: null},
+  {field: 'assignee', translation: I18n.t('gobierto_dashboards.dashboards.contracts.assignee'), format: null, cssClass: ''},
```
Crashillo, line 2: This file is becoming a jumble. What is this file for?

app/views/gobierto_dashboards/dashboards/contracts.html.erb
```diff
@@ -12,6 +12,10 @@
     </div>
   </div>
 
+  <div id="loading">
```
Crashillo: Quite general. Indeed, it's very likely something like this is already defined somewhere else. Two things here:

1. What if you load the Vue app with no data. Loading becomes a Vue component (shared). Once the data is fetched, you hide the Loading. In that way, you won't mix "reactive stuff" with static views.

2. If you need to access to the DOM, a good practice is about adding a css class like `js-loading` instead of using a very generic name

config/locales/gobierto_dashboards/views/ca.yml
```diff
@@ -23,9 +23,9 @@ ca:
           contracts: Contracts
           summary: Resum
           tenders: Licitacions
-        permalink: enllaç cap a la web oficial
+        permalink: Consulta aquest contracte en la Plataforma de Contractació del Sector Públic
```
Crashillo: Consulteu aquest contracte a la....

Crashillo (changes requested):

app/javascript/gobierto_dashboards/webapp/lib/config.js
eltercero: Yeah, I know :( we decided earlier that it will be separated in two files (one config for each dashboard) as soon as there isn't as many pending branches to merge.

app/views/gobierto_dashboards/dashboards/contracts.html.erb
eltercero: Thanks, I changed it so it uses a css class.

I saw the Loading component was already done in other modules, but those were getting the data within the vue app. Since the data here is fetched from outside vue, I thought this would be a more straightforward solution.

ferblape (approved):

app/views/gobierto_dashboards/dashboards/contracts.html.erb
Crashillo: That's why I say to load the full app with no data, an empty array. Then, when it's populated, you pass the data as a prop, and handle the states as usual

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -103,6 +103,8 @@ export class ContractsController {
         EventBus.$once('moved_to_summary', () => {
           this._redrawCharts();
         });
+
+        document.getElementsByClassName("js-loading")[0].classList.add('hidden');
```
Crashillo: Puff, ~disgusting~ a kitten has died.
```suggestion
        const elem = document.querySelector(".js-loading");
        if (elem) {
          elem.classList.add('hidden')
        }
```

If you need more than one, `querySelectorAll` gives you a `NodeList` (iterable)

config/locales/gobierto_dashboards/views/ca.yml
```diff
@@ -23,9 +23,9 @@ ca:
           contracts: Contracts
           summary: Resum
           tenders: Licitacions
-        permalink: enllaç cap a la web oficial
+        permalink: Consulteu aquest contracte en la Plataforma de Contractació del Sector Públic
```
Crashillo: ...contracte **a** la Plataforma...

Crashillo (changes requested):

Crashillo (approved):

PR #3099, opened 2020-05-23
2977 schema errors

### LINKED ISSUE #2977: Data / Invalid schema type shouldn't cause an error
If you provide an invalid type in the schema the application the error is not captured and causes a 500.

I'd expect the application would inform the user about the wrong type in the schema 

You can reproduce it if you provide an schema with "string" instead of "text" type.

entantoencuanto:
Closes #2977

## :v: What does this PR do?

* Avoids exceptions trying to upload datasets when:
  * schema JSON file is malformed
  * schema JSON file declares unexisting column types

No

No


ferblape (approved):
Works as expected 👍

PR #3103, opened 2020-05-25
1027 historic charges

### LINKED ISSUE #1027: Title and meta descriptions
The other day Alvaro commented that there was pending title and meta description for the pages. More resources should be completed:
- [ ] Issues: index/show
- [ ] Processes:  index/show
- [ ] Contribution containers:  index/show
- [ ] Events in Participation:  index/show
- [ ] Pages in Participation:  index/show
- [ ] Attachments in Participation:  index/show
- [ ] Activities in Participation:  index/show

entantoencuanto:
Closes PopulateTools/issues#1027
Closes PopulateTools/issues#1029
Closes https://github.com/PopulateTools/issues/issues/1110
Related to PopulateTools/issues#1031

## :v: What does this PR do?

* Adds a new model `GobiertoPeople::Charge` as described in the issue
* The charge has a `start_date` and an `end_date`. nil values are accepted: A nil value means that the interval has no limit for the corresponding boundary. For example, a charge with blank `start_date` and `end_date` implies that is in effect for all the time
* Adds new associations of gifts, invitations, trips and events with departments by using a concern. This allows us for gifts, invitations, trips or events to get the charge and department of the person in the date or start date of the resource
* Adds associations to departments with these resources to get the items of a department (all of them belongs to a person and this person has a charge in the department in a period in which the date or start date is included)
* Changes the `charge` method of people to accept a date and obtain the charge among the charges of the person compatible with the date. If no date is provided, the charge is obtained from the current date or if there is no charge at current date the last charge. If there are no charges created for the person the `charge` attribute of the model is taken (in a future this field probably will be removed)
* For pages which allows the selection of an interval of dates, the charge used currently for the people is the charge at the end of the interval, but it has been added to the person serializer new attributes with the charges of the person in the interval ordered by date. The charges for each user can be passed to the serializer from the controller to save requests to the database
* For events, the charge associated with it and displayed is the charge at the beginning of the event
* Adds some fixtures for charges and changes some tests with the new fixtures scenarios.
* Fixes some queries used by the custom engine to take into account events related with people belonging to a department via charge.
* I've tried to put the charge associated to each item in the lists and person page but since the names are sometimes very long it looks weird.
* Generates autolinks with the URLs present in the trips descriptions

For the moment it can't be tested in staging, so it should be done locally

No

No


app/models/gobierto_people/person.rb
```diff
@@ -64,6 +65,24 @@ def owned_attending_events
...
+      end
+    end
+
+    def charge(date = nil)
+      if historical_charges.exists?
+        historical_charge(date)&.name
+      else
+        return if charge_translations.blank?
+
+        charge_translations.fetch(I18n.locale, charge_translations&.values&.first)
```
ferblape, line 87: If `charge_translations` is a regular translate attribute why not replace the whole `else` block by a plain `super`? Would it work?

app/queries/gobierto_people/people_with_activities_query.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+module GobiertoPeople
+  class PeopleWithActivitiesQuery < RowchartItemsQuery
+
+    def initialize(params = {})
+      @site = params[:site]
+      @conditions = params[:conditions].symbolize_keys
+      @relation = (params[:relation] || model.all).where(id: people_with_activities.map(&:id)).left_outer_joins(events_association)
```
ferblape: Do we need the `all`?

app/queries/gobierto_people/people_with_activities_query.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      joins = ASSOCIATIONS.keys.map do |association_model|
+        "LEFT OUTER JOIN #{association_model}_relation on #{table_name}.id = #{association_model}_relation.id"
+      end.join(" ")
+
+      where = ASSOCIATIONS.keys.map do |association_model|
+        "#{table_name}.id = #{association_model}_relation.id"
+      end.join(" OR ")
+
+      "WITH #{cte} SELECT DISTINCT(#{table_name}.*) FROM #{table_name} #{joins} WHERE #{where}"
```
ferblape: Could we replace the `.* ` by `.id`? This query is only used here and you are only using the ids of the people

app/queries/gobierto_people/people_with_activities_query.rb
```diff
@@ -0,0 +1,106 @@
...
+        append_condition(:ends_at, conditions[:end_date], "<=", true)
+      end
+    end
+
+    def table_name
+      model.table_name
+    end
+
+    def activities_sql
+      cte = ASSOCIATIONS.keys.map do |association_model|
```
ferblape: What does cte stands for? The variable name could be improved a bit 😝

app/queries/gobierto_people/people_with_activities_query.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
```
ferblape, line 1: I've been reviewing this class during at least 10 minutes and I don't fully follow the queries you are generating. I think you could include some comments with the queries the method `activities_sql` generates

app/serializers/gobierto_people/person_serializer.rb
```diff
@@ -28,11 +34,43 @@ def political_group
...
     def position
-      object.charge
+      filtered_positions.first&.name
+    end
+
+    def all_positions_html
+      object.historical_charges.reverse_sorted.join("<br>")
+    end
+
+    def filtered_positions
```
ferblape, line 40: memoize please, it's used in many methods

app/views/gobierto_people/people/_person.html.erb
```diff
@@ -16,7 +16,9 @@
     <%= link_to gobierto_people_person_path(person.slug, date_range_params) do %>
       <h2><%= person.name %></h2>
     <% end %>
-    <div class="soft"><%= person.charge %></div>
+    <div class="soft tipsit" title="<%= person.historical_charges.reverse_sorted.join("<br>") %>">
```
ferblape: - This is a N + 1, are you including the historical_charges relation when loading the person?
- could you do a helper or move it to a presenter?

ferblape (changes requested):
Looks great, but it could be improved a bit ;)

app/models/gobierto_people/person.rb
entantoencuanto: Actually `charge` is overrided but it's not given by a superclass, is introduced with `translates` call in the model, calling super in this context generates an error saying that the method is not defined in any superclass. But if we create a superclass with the `translate  :charge` statement and the method is inherited from it, the `super` call raises an exception because the charges superclass doesn't accept arguments and this accepts an optional argument.

app/queries/gobierto_people/people_with_activities_query.rb
entantoencuanto: No, hehe.

app/queries/gobierto_people/people_with_activities_query.rb
entantoencuanto: Changed

app/queries/gobierto_people/people_with_activities_query.rb
entantoencuanto: It stands for `common_table_expression`. I've changed `cte` with it, it's much more expressive ;)

app/queries/gobierto_people/people_with_activities_query.rb
entantoencuanto: I've added some comments

app/serializers/gobierto_people/person_serializer.rb
entantoencuanto: Totally agree.

app/views/gobierto_people/people/_person.html.erb
entantoencuanto: Done

ferblape (approved):

ferblape:
@entantoencuanto can you check the conflicts?

ferblape:
tipsy wasn't imported in people module and it was causing a JS error in staging, it should be fixed now

PR #3141, opened 2020-06-08
CSV response generation improvement

ferblape:
## :v: What does this PR do?

This PR replaces how CSV results are generated by using the `\COPY` command from Postgres, which already returns a CSV and we skip Ruby to build it.

## Times

The improvement is very good when the result contains a big number of rows, starting at 10k

```
ab -n 50 -c 1 https://demo-datos.gobify.net/api/v1/data/data.csv?sql=select%20*%20from%20global_mobility_report%20limit%20100000
```

Rows | Before | After
-- | -- | --
100 | 50 | 80
1000 | 70 | 60
10000 | 624 | 151
100000 | 2436 | 1053
224147 | 5708 | 2374

### DIFF
--- a/app/controllers/gobierto_data/api/v1/base_controller.rb
+++ b/app/controllers/gobierto_data/api/v1/base_controller.rb
@@ -21,6 +21,8 @@ module GobiertoData
           {}.tap do |options|
             if (separator = params[:csv_separator]).present?
               options[:col_sep] = separator_tr.fetch(separator, separator)
+            else
+              options[:col_sep] = separator_tr["comma"]
             end
           end
         end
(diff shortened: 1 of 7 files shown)


ferblape:
When the column name is wrong there's an exception.

Error: https://appsignal.com/populate/sites/5eb0fcfc74782044d1198606/exceptions/incidents/162?timestamp=2020-06-12T06%3A14%3A37Z

URL: https://getafe.gobify.net/api/v1/data/data.csv?sql=SELECT%20*%20FROM%20censo_barrios_estudios_sexo%20where%20barrio=%27Centro%27%20and%20rango_edad%20=%20%270%20-%204%27

ferblape:
@entantoencuanto please review and merge if ok

entantoencuanto (approved):
Very good improvement

PR #3152, opened 2020-06-11
Circle ci move asset compilation

ferblape:
## :v: What does this PR do?

This PR improves the performance of the tests and the execution in CircleCI

- node_modules are persisted, so we save ~ 2 mins in `yarn install`
- assets are cached in the branch, so if a branch is being executed many times and there are no changes in the assets, the execution will be also ~ 1 min faster
- I've removed the check of normalized I18n files, keeping only missing and unused. This saves around ~ 30 secs, and now we have Husky. Besides that, normalization is not such a reason to raise a failure and won't cause a failure in production
- I've removed a extra step to logout admins, not necessary because we reset the session

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,19 +28,32 @@ jobs:
       - restore_cache:
           name: "[Bundler] Restore cache"
           keys:
-            - gobierto-{{ checksum "Gemfile.lock" }}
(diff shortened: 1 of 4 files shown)


ferblape:
I'm going to merge it into staging, I hope not to break CircleCI

PR #3164, opened 2020-06-17
Explicit class name in relation (try to avoid random errors in CI)

ferblape:
##  :v: What does this PR do?

Forces class name in relation to try to avoid CI random errors


entantoencuanto (commented):
wow!

PR #3179, opened 2020-06-23
Gobierto Data: check if the user loads an invented dataset

### LINKED ISSUE #3154: Gobierto data / Don't load invalid datasets
When you load an invented dataset, i.e https://getafe.gobify.net/datos/blat-is-the-best the UI is loading the dataset as if it'd exist. I'd expect to be redirected to the list of datasets

Notice that the api call to the meta of that dataset is returning 404 but you are not checking that response code.

jorgeatgu:
Closes #3154 

## :v: What does this PR do?

Check if the user loads an invented dataset, in this case, redirect the user to dataset list

Staging

### After this PR
after

### DIFF
--- a/app/javascript/gobierto_data/lib/factories/datasets.js
+++ b/app/javascript/gobierto_data/lib/factories/datasets.js
@@ -12,18 +12,8 @@ const headers = {
 // Dataset-endpoint factory to get/post/put/delete API data
 export const DatasetFactoryMixin = {
   methods: {
-    async getDatasetMetadata(id) {
+    getDatasetMetadata(id) {
       return axios.get(`${endPoint}/${id}/meta`, { headers })
-        .then(response => {
-          return response
-        })
-        // eslint-disable-next-line no-unused-vars
-        .catch(error => {
-          if (error.response.status === 404) {
-            // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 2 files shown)


ferblape (approved):

app/javascript/gobierto_data/lib/factories/datasets.js
```diff
@@ -12,8 +12,18 @@ const headers = {
...
+        .then(response => {
+          return response
+        })
+        // eslint-disable-next-line no-unused-vars
+        .catch(error => {
+          if (error.response.status === 404) {
+            // eslint-disable-next-line no-unused-vars
+            this.$router.push('/datos/').catch(error => {})
+          }
+        })
```
Crashillo: Why must a `factory` know the error response calls a router action? You're mixing unrelated concepts. A couple of thing here:

1. `getDatasetMetadata` shouldn't be async. It only returns a promise
2. `.then(response => { return response })` REALLY? 😒 Remove that clause
3. No catches here. Handle the catch error where `getDatasetMetadata` is being called, wrapping it into a try-catch block.

Crashillo (changes requested):

ferblape:
Ping @Crashillo

Crashillo (approved):

PR #3182, opened 2020-06-24
Gobierto Data: Render markdown

### LINKED ISSUE #3160: Gobierto Data / Datasets descriptions
URL: https://getafe.gobify.net/datos/calidad-del-aire

While datasets description supports markdown in the admin, the front doesn't render markdown.

Screenshot 2020-06-16 at 08 52 18

Two issues:

- [ ] The dataset page should also render the description as markdown

Screenshot 2020-06-16 at 08 49 42

- [ ] According to the mockups the text should be truncated in the list of datasets:

Screenshot 2020-06-16 at 08 56 38

But now the whole text is displayed:

Screenshot 2020-06-16 at 08 56 15

jorgeatgu:
Closes #3160 

## :v: What does this PR do?

Dataset descriptions can use Markdown, so we need to render Markdown in the front.
When the description contains a long text, we truncate the text and show a button.

Staging

### After this PR

after


ferblape (changes requested):
Could we show the link "Ver más" only if there is hidden content?

jorgeatgu:
> Could we show the link "Ver más" only if there is hidden content?

Done!

ferblape (approved):

app/javascript/gobierto_data/webapp/components/commons/Info.vue
```diff
@@ -25,15 +25,42 @@
...
+            </span>
+          </template>
+          <template v-else>
+            <span
+              class="gobierto-data-summary-header-description-link"
+              @click="scrollDetail"
+            >
+              {{ seeLess }}
+            </span>
+          </template>
```
Crashillo: Since you conditionally render just an `span`, you may remove the `templates` and add the directives `v-if`/`v-else` in the spans directly

Crashillo (approved):

app/javascript/gobierto_data/webapp/components/commons/Info.vue
jorgeatgu: Done!

PR #3189, opened 2020-06-29
Remove fingerprint from assets paths in processes fixtures

entantoencuanto:
## :v: What does this PR do?

Fixes path of some assets using a fingerpring declared in a fixture file. After some recent update these paths was failing and producing errors in some related tests

No

No


ferblape:
Good catch!

PR #3193, opened 2020-06-30
3162 locale change

### LINKED ISSUE #3162: User locale on sites with different locales
When a user visits Gobierto, her preferred locale is stored in a cookie.

When you visit a different Gobierto site, no matter what languages it has configured, the locale of the cookie is applied, even if that locale is not in the site available locales.

I'd expect that if the locale is not in the list of available locales, the user locale should be changed to the site default locale.

entantoencuanto:
Closes #3162

## :v: What does this PR do?

* Changes the locale to the default locale of a site if the cookie locale is not available for the site
* Adds an integration test

No

No


ferblape:
Verified in staging, thanks!

PR #3196, opened 2020-07-01
Gobierto Data / Update URL when saving a new query or visualization

### LINKED ISSUE #3154: Gobierto data / Don't load invalid datasets

jorgeatgu:
Closes #3154 https://github.com/PopulateTools/issues/issues/1014

## :v: What does this PR do?

Update the URL when saving a new query or visualization.

User-cases:

- Save a new query/viz from scratch.
- Load an own query/viz, modify it, and save it as new.
- Load a query/viz from another user.

### Before this PR

**Query:**
before_query

**Visualization**
before_viz

### After this PR

**Query**
after_query

**Visualization:**
after_viz


app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -451,7 +449,7 @@ export default {
       this.isPublicVizLoading = false
     },
     async getPrivateVisualizations() {
-      const userId = getUserId()
+      const userId = Number(getUserId());
```
Crashillo, line 510: Keep in mind if `userId` was `0` the conditional clause would be falsy, so you wouldn't call to getVisualizations.
By the way, another "cool" way to cast an integer is: `      const userId = +getUserId();`

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -610,15 +610,26 @@ export default {
...
         this.isQueryModified = false
         this.isQuerySavingPromptVisible = false
 
-        this.setPublicQueries(await this.getPublicQueries());
         this.setPrivateQueries(await this.getPrivateQueries());
+
+        //Get the old list
+        const oldQueries = this.publicQueries
+        //Get the list with the new query
+        const newQueries = this.privateQueries
```
Crashillo: This is confusing. `publicQueries` and `privateQueries` store different stuff, why now is `old` and `new`? Wha's the point?

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -726,9 +737,38 @@ export default {
         this.vizName = null
 
         await this.getPrivateVisualizations()
+
+        //Get the list before the newViz is saved
+        const oldVizs = this.publicVisualizations
+        //Get the list with the new visualization
+        const allNewVizs = this.privateVisualizations
```
Crashillo: Same on this

app/javascript/gobierto_data/webapp/components/sets/VisualizationsItem.vue
```diff
@@ -46,6 +49,7 @@
   </div>
 </template>
 <script>
+import Spinner from "./../commons/Spinner.vue";
```
Crashillo: Question: is this Spinner different from the `Loading` component? (lib/vue-components)

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: For example, you save a new query, at that moment, that query is not yet in `this.publicQueries`, but it's in `this.privateQueries` since it has been updated a [few lines before](https://github.com/PopulateTools/gobierto/blob/e0b7450a19612fbe6fc688db810d819603db0e99/app/javascript/gobierto_data/webapp/components/Dataset.vue#L619). 

```Javascript
console.log(this.publicQueries.length) 5

//Contains the new query
console.log(this.privateQueries.length) 6 
```

So I can compare the list before saving with the list containing the query I just kept. Then I filter them, and so I get the new saved query.

And now we [updated](https://github.com/PopulateTools/gobierto/blob/e0b7450a19612fbe6fc688db810d819603db0e99/app/javascript/gobierto_data/webapp/components/Dataset.vue#L632) `this.publicQueries` 

```Javascript
console.log(this.publicQueries.length) 6
console.log(this.privateQueries.length) 6
```

app/javascript/gobierto_data/webapp/components/sets/VisualizationsItem.vue
jorgeatgu: No. I change it.

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: Yes, if the `user` was `0` = user is not logged, so he can't see `privateVisualizations`.

Thanks for the tip!

Crashillo (approved):

ferblape:
I'm not able to make it work. Steps:

1. Go to https://getafe.gobify.net/datos/censo-barrios-edad-nacionalidad/editor
2. Use this simple query `SELECT barrio, sum(total) as total FROM censo_barrios_edad_nacionalidad group by barrio`
3. Click Visualization > X Bar chart
4. Create private visualization with a name and click Save, the URL keeps the same

Screenshot 2020-07-02 at 15 12 05

BTW, there's a request to `https://getafe.gobify.net/api/v1/data/data.csv?sql=null` which returns 400 Bad request, why is that call being done?

jorgeatgu:
I made a few changes to update the URL when saving a Visualization in the editor.

> BTW, there's a request to https://getafe.gobify.net/api/v1/data/data.csv?sql=null which returns 400 Bad request, why is that call being done?

After saving a Visualization, we update the `publicVisualization` and `privateVIsualization`, so a request is made again. In this case, the error is caused by [deleting this line.](https://github.com/PopulateTools/gobierto/pull/3196/commits/f26e8e555794e7671f19c056966bd6e8eac5067e#diff-0cf2260d7e97b413f3920a5bd03a2770L699)

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -782,9 +789,37 @@ export default {
...
+      //Changes the path depending on if we save a query or viz.
+      const pathQueryOrViz = this.savingViz ? 'v' : 'q'
+
+      /*Don't updates the URL if the component is Editor, only replace and don't reload, because in the editor we've two options, saved a query or viz, if the user saves a viz, and we update the URL, the browser reloads, and the user goes to visualization tab, and this behavior is too hacky.*/
+      //https://developer.mozilla.org/en-US/docs/Web/API/History/pushState
+      history.pushState(
+        {},
+        null,
+        `${location.origin}/datos/${slugDataset}/${pathQueryOrViz}/${newId}`
+      )
```
Crashillo, line 818: Is there any chance to do this through the vue-router?

ferblape (approved):
Ok, to me! @Crashillo confirm and we deploy

Crashillo (approved):

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: Nope, because `$router.push` and `$router.replace` reload the entire page, so if a user saves a visualization in the editor, we will redirect to the visualization tab

PR #3200, opened 2020-07-03
Gobierto Data / Improve visualization flow

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1078

## :v: What does this PR do?

- Hide private icons and label when loads a Visualization from others users.
- Increase the width of the input
- Show in a block the query related to Vizualization
- Remove all icons config from Visualizations list
- Scroll to top when user click in a Visualization
- Refactor to improve visualizations loads: only reloads visualizations when the user saves or delete a visualization

Staging


app/javascript/gobierto_data/lib/commons.js
```diff
@@ -1,24 +1,5 @@
...
-      unbind: function(el) {
-        document.body.removeEventListener('click', el.clickOutsideEvent)
-      },
-      stopProp(event) { event.stopPropagation() }
-    }
-  }
-}
-
 //TODO: extract handleOutsideClick outside directive
 export const closableMixin = {
```
Crashillo, line 4: is it quite possible the `closableMixin` was also deprecated, perhaps?

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -217,7 +227,7 @@ export default {
       }
 
       //Update only the baseTitle of the dataset that is active
-      if (to.name === 'Dataset' && this._inactive === false) {
+      if (to.path !== from.path && to.name === 'Dataset' && this._inactive === false) {
```
Crashillo: `to.path` and `from.path` should be always different, otherwise the router won't be triggered, will it?

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -18,40 +18,43 @@
...
-    <template v-if="enableForkPrompt">
-      <label
-        :for="labelPrivate"
+    <template v-if="showPrivateIcon">
+      <PrivateIcon
+        :is-closed="isPrivate"
+        :style="{ paddingRight: '.5em', margin: 0 }"
+      />
+      <div
+        v-if="isQueryModified || isVizModified"
```
Crashillo: Line 243, you've a computed property

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -238,24 +255,22 @@ export default {
...
-    isQuerySaved(newValue) {
-      if (newValue) {
-        this.disabledButton = true
-      }
-    },
-    enabledForkButton(newValue, oldValue) {
-      if (newValue !== oldValue) {
-        this.enabledForkButton = newValue
+    $route(to, from) {
+      if (to.path !== from.path) {
```
Crashillo: Same as I mentioned before

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -238,24 +255,22 @@ export default {
...
-      this.$nextTick(() => {
-         this.countInputCharacters(this.value)
-      });
+  created() {
+    if (this.showPrivateViz) {
+      this.isPrivate = this.showPrivateViz
     }
   },
+  mounted() {
+    this.$nextTick(() => {
```
Crashillo: 🤔 nextTick on mounted? Why so?

app/javascript/gobierto_data/lib/commons.js
Crashillo: I just saw it's used through `v-closable` directive

Crashillo (changes requested):

ferblape:
In a private visualization the link to the query is missing:

Screenshot 2020-07-16 at 07 10 09

ferblape:
It's just happened to me right now:

- I've edited Wadus visualization (private)
- make it public
- changed the name to Wadus 2

And it has been duplicated, now there are two visualizations: Wadus and Wadus 2

Screenshot 2020-07-17 at 12 03 19

app/javascript/gobierto_data/webapp/components/commons/Queries.vue
```diff
@@ -152,7 +153,10 @@ export default {
       this.$emit('closeQueriesModal')
     },
     clickDeleteQueryHandler(id) {
-      this.$root.$emit('deleteSavedQuery', id)
+      const answerDelete = confirm(this.labelDeleteQuery);
```
Crashillo, line 157: cheap :D

app/javascript/gobierto_data/webapp/components/commons/SavingDialog.vue
```diff
@@ -306,17 +349,43 @@ export default {
...
+      let editorInputViz
+      if (this.$refs.inputText) {
+        editorInputViz = this.$refs.inputText.placeholder.includes(this.labelVisName) ? true : false
+      }
+
+      //In the Visualizations view, we've more space, so we take advantage of it by increasing the input name width.
+      if (this.$route.name === 'Visualization') {
+        maxWidth = maxWidth * 1.4
+        minWidth = minWidth * 1.4
+      } else if (editorInputViz) {
```
Crashillo: This condition is so twisted. First off, `editorInputViz` should be initialized, in this case to `false`, but actually the real check is just `this.$refs.inputText.placeholder.includes(this.labelVisName)` which could be simplified like:
```
const { placeholder = [] } = this$ref.inputText || {};
....
else if (placeholder.includes(this.labelVisName)) {
....
}
```

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -95,7 +105,14 @@ export default {
...
     },
     toggleConfigPerspective() {
+      this.$root.$emit('showSavedVizString', false)
       this.viewer.toggleConfig()
+      //Enable save button when user interacts with Perspective columns
+      const itemPerspective = document.querySelector('perspective-viewer').shadowRoot
+      const rowPerspective = itemPerspective.querySelectorAll("perspective-row");
+
+      rowPerspective.forEach(rowMenu => rowMenu.addEventListener('drag', () => this.$emit("showSaving")));
+      rowPerspective.forEach(rowMenu => rowMenu.addEventListener('click', () => this.$emit("showSaving")));
```
Crashillo: You could merge both actions in the same loop in order to avoid run through the array twice

app/javascript/gobierto_data/webapp/components/sets/VisualizationsItem.vue
```diff
@@ -105,83 +113,85 @@ export default {
...
-    async $route(to, from) {
-      if (to.path !== from.path) {
-        await this.getDataVisualization(this.privateVisualizations);
-        await this.getDataVisualization(this.publicVisualizations);
-      }
     }
   },
-  async created() {
+  created() {
+    this.$root.$emit("isVizModified", false);
```
Crashillo: I don't see the point of calling to this events both in created and destroy hooks. Why a component should do these sort of things when there's no user interaction? Is really so necessary?

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -727,6 +733,13 @@ export default {
       }
     },
     async storeCurrentVisualization(config, opts) {
+      const {
+        params: {
+          queryId
+        }
+      } = this.$route;
+
+      let vizIdFromRoute = +queryId
```
Crashillo, line 748: Why do you redeclare `queryId`? There aren't more uses

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -927,11 +969,18 @@ export default {
     },
     eventIsVizModified(value) {
       this.isVizModified = value
+      const userId = Number(getUserId());
+      //Don't show privatePublicIcon when load a viz from other user
+      if (this.vizUserId !== userId && this.enabledForkButton) {
+        this.showPrivatePublicIconViz = false
+      } else {
+        this.showPrivatePublicIconViz = true
+      }
```
Crashillo: Why not?
```suggestion
     this.showPrivatePublicIconViz = (this.vizUserId !== userId && this.enabledForkButton)
```

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/sets/VisualizationsItem.vue
jorgeatgu: It's not necessary

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: To prevent confusion, rename `queryId` to `vizIdFromRoute `, because in this function, we'll try to save or update a Visualization not a Query.

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -809,43 +828,45 @@ export default {
...
-        this.currentQuery = `SELECT * FROM ${this.tableName} LIMIT 50`;
-        this.isQueryModified = false
-        this.runCurrentQuery()
-        this.disabledSavedButton()
-        this.disabledStringSavedQuery()
-        this.queryName = null
-      }
+    resetQuery() {
+      this.isQuerySavingPromptVisible = false
+      this.currentQuery = `SELECT * FROM ${this.tableName} LIMIT 50`;
```
Crashillo, line 844: You might want to create a computed prop (initialQuery) and replace it in both places

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -876,29 +899,51 @@ export default {
...
+          this.isForkPromptVisible = true
+        } else {
+          this.showPrivatePublicIcon = true
+          this.disabledForkButton()
+        }
+      } else if (userId !== 0 && nameComponent === 'Visualization') {
+
+        this.showLabelEdit = true
+
+        const objectViz = this.privateVisualizations.find(({ id }) => id === queryId) || {}
```
Crashillo, line 929: This whole file has become massive. Perhaps, you should think about splitting the logic in Queries and Visualizations, create both mixins where you set all the stuff related (events, attributes) and maybe you'll avoid the conditional clauses all around.

Think about it, it would deserve a PR to make it happen.

Crashillo (approved):

ferblape:
I'm going to merge this nos that got approved to unblock the following PRs

PR #3201, opened 2020-07-03
Check value of amount is NaN

eltercero:
Closes #1071

## :v: What does this PR do?

It fixes an issue that would make the subsidies page run very slow. It happened when DC tries to crossfilter data with undefined values.

https://getafe.gobify.net/dashboards/subvenciones should not crash your browser.

### DIFF
--- a/app/javascript/gobierto_dashboards/modules/contracts_controller.js
+++ b/app/javascript/gobierto_dashboards/modules/contracts_controller.js
@@ -146,8 +146,8 @@ export class ContractsController {
 
     for (let i = 0; i < contractsData.length; i++){
       const contract = contractsData[i];
-      const final_amount_no_taxes = contract.final_amount_no_taxes && !isNaN(contract.final_amount_no_taxes) ? parseFloat(contract.final_amount_no_taxes) : 0.0;
-      const initial_amount_no_taxes = contract.initial_amount_no_taxes && !isNaN(contract.initial_amount_no_taxes) ? parseFloat(contract.initial_amount_no_taxes) : 0.0 ;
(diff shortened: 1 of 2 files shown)


ferblape (approved):

eltercero:
@ferblape I've realised this could happen in contracts dashboard as well and added a commit for that too, if you could please check it again just in case 🙌

app/javascript/gobierto_dashboards/modules/contracts_controller.js
```diff
@@ -146,8 +146,8 @@ export class ContractsController {
 
     for (let i = 0; i < contractsData.length; i++){
       const contract = contractsData[i];
-      const final_amount_no_taxes = contract.final_amount_no_taxes ? parseFloat(contract.final_amount_no_taxes) : 0.0;
-      const initial_amount_no_taxes = contract.initial_amount_no_taxes ? parseFloat(contract.initial_amount_no_taxes) : 0.0 ;
+      const final_amount_no_taxes = contract.final_amount_no_taxes && !isNaN(contract.final_amount_no_taxes) ? parseFloat(contract.final_amount_no_taxes) : 0.0;
```
Crashillo: Use `Number.isNaN` better

https://medium.com/dailyjs/better-nan-check-with-es6-number-isnan-517861d32be3

Crashillo (approved):

app/javascript/gobierto_dashboards/modules/contracts_controller.js
eltercero: Changed, thanks!

PR #3209, opened 2020-07-07
Custom fields reorder

entantoencuanto:
## :v: What does this PR do?

* When custom fields are reordered, created, updated or deleted if they depend on an instance (for example, custom fields for plans) it is touched to clear cache

No

No


app/controllers/gobierto_admin/gobierto_common/custom_fields/custom_fields_controller.rb
```diff
@@ -114,6 +117,12 @@ def destroy
 
         private
 
+        def touch_instance
+          return unless @custom_field_form&.instance&.present?
+
+          @custom_field_form.instance.touch
```
ferblape: Why not? 

```
        def touch_instance
          @custom_field_form.&instance.&touch
        end
```

entantoencuanto: Uf, much better! 😅

PR #3210, opened 2020-07-07
Gobierto Data / Including a skeleton spinner

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1079

## :v: What does this PR do?

It includes a skeleton spinner that simulates content while loading the different data. 

Staging

### After this PR

after_spinner 2020-07-07 17_56_54


app/javascript/lib/vue-components/modules/SkeletonSpinner.vue
```diff
@@ -0,0 +1,65 @@
...
+    styleSquares() {
+      return {
+        //
+        width: ((100 - (this.squares - 1)) / this.squares) + '%',
+        height: this.heightSquare
+      }
+    },
+  },
+  methods: {
+    stylesLines() {
```
Crashillo, line 57: Why this method is not a computed one?

config/locales/defaults/ca.yml
```diff
@@ -15,7 +15,7 @@ ca:
     - dv
     - ds
     abbr_month_names:
-    -
+    -
```
Crashillo, line 18: Maybe could we remove these empty values?

Crashillo (approved):

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -1,111 +1,122 @@
...
   <div>
-    <div class="pure-g">
-      <div class="pure-u-1-2">
-        <h2
-          v-if="titleDataset"
-          class="gobierto-data-title-dataset"
-        >
-          {{ titleDataset }}
-        </h2>
+    <template v-if="!isDatasetLoading">
```
ferblape: I think this condition is in the wrong direction: if the dataset is loading, show the spinner.  Do you mean isDatasetLoaded?

app/javascript/lib/vue-components/modules/SkeletonSpinner.vue
jorgeatgu: It was. But this method determines the width of the lines, if it were computed the width would always be the same.

app/javascript/lib/vue-components/modules/SkeletonSpinner.vue
```diff
@@ -0,0 +1,65 @@
...
+    },
+    lines: {
+      default: '0',
+      type: String,
+    }
+  },
+  computed: {
+    styleSquares() {
+      return {
+        //
```
ferblape: empty comment?

app/javascript/lib/vue-components/modules/SkeletonSpinner.vue
```diff
@@ -0,0 +1,65 @@
...
+        width: ((100 - (this.squares - 1)) / this.squares) + '%',
```
ferblape: If `this.squares` is a String shouldn't you convert it to integer or float? Careful with zero division

app/javascript/lib/vue-components/modules/SkeletonSpinner.vue
```diff
@@ -0,0 +1,65 @@
...
+    </template>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "SkeletonSpinner",
+  props: {
+    heightSquare: {
+      default: `1rem`,
```
ferblape: Single quotes?

ferblape (changes requested):
Please review my comments

config/locales/defaults/ca.yml
jorgeatgu: This looks strange. I didn't change these files https://github.com/PopulateTools/gobierto/blob/fb23c24f82dcf53dd41ba6c65c97849f62e20f13/config/locales/defaults/ca.yml cc @ferblape @entantoencuanto

ferblape: It looks like you added a space, but you can ignore it, we can't remove the blank here, it's a Rails default.

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: You're right.

app/javascript/lib/vue-components/modules/SkeletonSpinner.vue
jorgeatgu: In the division JS always try to transforms `String` into `Number`.

> Careful with zero division

Done

PR #3220, opened 2020-07-09
Vue Components / Remove styles from the checkbox component

jorgeatgu:
This style shouldn't be specific to the checkbox component. We can use the checkbox component in other components than filters.

For example: 
imagen


Crashillo (approved):
Wouldn't be more accurate use the `gobierto-data-filters` or `gobierto-data-filters-element` classes? Just asking

jorgeatgu:
Yes, much better to use `gobierto-data-filters-element`

Crashillo (approved):

PR #3237, opened 2020-07-15
3175 remove algolia dependency

### LINKED ISSUE #3175: Remove Algolia dependency
We are going to get rid of Algolia with https://github.com/Casecommons/pg_search? in order to simplify dependencies. 

Lets start with a work plan, analyzing the steps to follow, and a plan it to release in different phases (PRs):

- existing modules (specially agendas and cms, which have most of Algolia
entries)
- budgets module
- modules pending to be indexed (I think the most important one is data)

entantoencuanto:
Closes #3175

## :v: What does this PR do?

Moves search engine from algolia to pg_search:
* Configures pg_search with multisearch option and adds some attributes and associations to the `PgSearch::Document` model:
  * `belongs_to` association with `Site`
  * Translatable attributes: `title` and `description`
  * `resource_path`
  * `meta`: Used to store some extra attributes like kind of budget_line or collection title and id of GobiertoCms::Page instances
  * `external_id` used in `GobiertoBudgets::BudgetLine` document types because the indexed elements are not stored in database
  * `searchable_udpated_at`: It copies the `updated_at` attribute from searchable
* Adds multisearch to models indexed with algolia
* Extends search support to `GobiertoPlans::Node` taking into account versions and including in search document custom fields of type text values
* Adds some tests for plans projects checking that the search documents contains the versioned values of the native fields and the custom fields
* Replaces the concern used to index budget lines in algolia to do the same job in pg_search and changes callbacks using algolia to use pg_search instead
* Adds an API endpoint to make searches using pg_search.
* Replaces global front search box and pages search in admin/cms/sections to use the API instead of algolia
* Removes the algolia js package and gem
* Fixes styles of front search box
* Removes derprecated tasks and fixes some tests

- There is a PR in populate-ansible to remove environment variables related to algolia

No


app/javascript/lib/shared/modules/module-search.js
```diff
@@ -60,73 +63,42 @@ $(document).on('turbolinks:load', function() {
...
-      });
-    } else {
-      $('<div class="result"><p>'+I18n.t("layouts.search.no_results")+'</p></div>').appendTo($resultsContainer);
-    }
-
-    if(window.searchClient.indexes.length > 1) {
-      $('<div class="result"><small>'+I18n.t("layouts.search.powered_by")+'</small></div>').appendTo($resultsContainer);
-    }
+  function formattedResult(d){
+    return '<h2>' + d['title'] + '</h2>' +
```
ferblape: You can use interpolations, i.e `<h2>${d['title']}</h2>` + ....

Crashillo:
You may take a look @entantoencuanto

app/models/concerns/gobierto_budgets/budget_line_algolia_helpers.rb
```diff
@@ -45,10 +45,10 @@ def algolia_as_json
           kind: kind,
           resource_path: resource_path,
           class_name: self.class.name
-        }.merge(translated_attributes)
+        }.merge(algolia_translated_attributes)
       end
 
-      def translated_attributes
+      def algolia_translated_attributes
```
ferblape: Why not remove algolia from the file name and from the method?

app/views/layouts/_gobierto_footer.html.erb
```diff
@@ -12,8 +12,9 @@
   <% if algoliasearch_configured? %>
     window.searchClient = {
       client: algoliasearch('<%= Rails.application.secrets.algolia_application_id %>', '<%= Rails.application.secrets.algolia_search_api_key %>'),
```
ferblape: can you remove the client option, right?

app/models/concerns/gobierto_budgets/budget_line_algolia_helpers.rb
entantoencuanto: This entire helper file is going to be removed after cleaning the algolia dependencies, but this method name conflicts with another one added by the new code

app/javascript/lib/shared/modules/module-search.js
entantoencuanto: Done!

app/models/concerns/gobierto_budgets/budget_line_algolia_helpers.rb
entantoencuanto: Al the algolia references has been deleted in this branch

app/controllers/gobierto_common/api/v1/search_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
```
ferblape, line 1: Could you add a test of this endpoint?

ferblape (changes requested):

app/controllers/gobierto_common/api/v1/search_controller.rb
entantoencuanto: Done!

ferblape (changes requested):
- in Gobierto data module the search input doesn't work
- if you open it in mobile it doesn't work either. You don't need to use a mobile, you can make the browser small to enable the mobile layout

entantoencuanto:
> * in Gobierto data module the search input doesn't work

Attached to another event, I think it's working now

> * if you open it in mobile it doesn't work either. You don't need to use a mobile, you can make the browser small to enable the mobile layout

There was a duplication in the input id, I've set two different inputs with their own autocomplete

PR #3261, opened 2020-07-24
Detect CSV encoding before UTF-8 conversion

ferblape:
Closes [#issues/1096](/PopulateTools/issues/issues/1096)

## :v: What does this PR do?

This PR improves the encoding conversion of datasets before being imported. The previous approach wasn't valid if the dataset wasn't in UTF-8 format. The best way to force a conversion to UTF-8 is providing with `#force_encoding` the current encoding, and then convert to UTF-8.

To obtain the current encoding of the data I've found a very good gem maintained by top rubyists that works really well: https://github.com/brianmario/charlock_holmes/

Try to import a CSV in a different encoding that UTF-8.


ferblape:
@entantoencuanto ping, could you do the acceptance today? we need this in production

entantoencuanto (approved):

PR #3263, opened 2020-07-24
Accept column names prefixed with Node for external_id

entantoencuanto:
## :v: What does this PR do?

Allows to import custom fields using both `external_id` and `Node.external_id` (as it is in the plan projects form) for the project external_id value

Import custom fields from a csv in admin. The result should be the same when the external_id column is named `external_id` and `Node.external_id`

No

No


ferblape:
Why there's no test?

entantoencuanto:
You're right, I'm goint to add one

PR #3280, opened 2020-08-03
3007 track gobierto data activities

### LINKED ISSUE #3007: Track gobierto data activities
At least for and admin I'd expect these activities to be tracked:

- create dataset
- update dataset
- update dataset data
- remove dataset

entantoencuanto:
Closes #3007

## :v: What does this PR do?

* Adds site activities when Gobierto data API is used to create, update attributes or import data from CSV with datasets.
* Also includes a site activity when a dataset is deleted from admin. This activity is associated with site instead of dataset because otherwise, since the dataset does not exist, the event cannot be created.
* Includes some tests for activities in API controller and admin integration tests

Visit admin and create, edit attributes and delete datasets or use the API to create, update data or change datasets attributes 

No

No

### DIFF
--- a/app/controllers/gobierto_admin/gobierto_data/datasets_controller.rb
+++ b/app/controllers/gobierto_admin/gobierto_data/datasets_controller.rb
@@ -61,8 +61,9 @@ module GobiertoAdmin
 
       def destroy
         @dataset = find_dataset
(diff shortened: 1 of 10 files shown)


ferblape (changes requested):
When I create a new dataset, could you skip the data updated event and leave just the dataset created event? Right now is a bit confusing:

Screenshot 2020-08-04 at 07 00 51

Fix it only if it's quick

entantoencuanto:
Done!

ferblape:
Perfect

PR #3283, opened 2020-08-04
3222 spread custom fields api optimization

### LINKED ISSUE #3222: Spread custom fields api optimizations to the rest of APIs
Following the work on https://github.com/PopulateTools/gobierto/pull/3202, adapt the other api endpoints using custom fields to get the speed improvement:

- [ ] investments
- [ ] data
- anywhere else?

entantoencuanto:
Closes #3222

## :v: What does this PR do?

* Extends custom fields API optimization to data and investments Gobierto modules
* Adapts front components to new API response (for vocabulary custom fields the old response was always an array of serialized vocabulary terms and now is the term id)
* Fixes an error mixing API optimization and filters

Visit affected modules. Nothing should change

No

No

### DIFF
--- a/app/javascript/gobierto_investments/webapp/components/MapTour.vue
+++ b/app/javascript/gobierto_investments/webapp/components/MapTour.vue
@@ -78,6 +78,7 @@ import {
 } from "vue-mapbox";
 import Wkt from "wicket";
 import axios from "axios";
+import { Middleware } from "lib/shared";
 import { CommonsMixin, baseUrl } from "./../mixins/common.js";
 
 export default {
@@ -142,6 +143,10 @@ export default {
       ] = responses;
 
       this.dictionary = attributesDictionary;
+      this.middleware = new Middleware({
+        dictionary: attributesDictionary
+      });
+
       this.items = this.setData(items);
 
       this.subsetItems = this.items;
--- a/app/javascript/gobierto_investments/webapp/containers/home/Home.vue
+++ b/app/javascript/gobierto_investments/webapp/containers/home/Home.vue
@@ -374,9 +374,8 @@ export default {
         // Get the items based on these new active filters
         const __items__ = this.applyFiltersCallbacks(__activeFilters__);
 
-        return __items__.filter(({ attributes }) =>(
+        return __items__.filter(({ attributes }) =>
           this.convertToArrayOfIds(attributes[key]).includes(id)
-          )
         ).length;
       };
       const { key, options = [] } = filter;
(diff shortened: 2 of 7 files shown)


entantoencuanto:
Please, @Crashillo, could you review the js changes before I merge into staging?

app/javascript/gobierto_investments/webapp/containers/home/Home.vue
```diff
@@ -374,8 +374,9 @@ export default {
         // Get the items based on these new active filters
         const __items__ = this.applyFiltersCallbacks(__activeFilters__);
 
-        return __items__.filter(({ attributes }) =>
-          attributes[key].map(g => g.id).includes(id)
+        return __items__.filter(({ attributes }) =>(
```
Crashillo: You may get rid of the wrapping parenthesis, in the filter condition

app/javascript/gobierto_investments/webapp/components/MapTour.vue
```diff
@@ -142,6 +143,10 @@ export default {
       ] = responses;
 
       this.dictionary = attributesDictionary;
+      this.middleware = new Middleware({
```
Crashillo, line 146: You're initializating the middleware, but what are you doing then?

app/javascript/gobierto_investments/webapp/mixins/common.js
```diff
@@ -89,6 +89,7 @@ export const CommonsMixin = {
         availableProjectFields
       } = CONFIGURATION;
       const { id: locationId, ...restLocationOptions } = location;
+      const phase_terms = this.convertToArrayOfIds(attributes[phases.id]).map(phase_id => this.middleware.getAttributesByKey(phases.id).vocabulary_terms.find(({ id }) => id === phase_id));
```
Crashillo: This could fail. Better split it into several steps:

```suggestion
      const phase_terms = this.convertToArrayOfIds(attributes[phases.id]).map(phase_id => {
        const { vocabulary_terms= [] } = this.middleware.getAttributesByKey(phases.id) || {}
        return vocabulary_terms.find(({ id }) => id === phase_id) || attributes[phases.id]
      }) || [];
```

The last `... || attributes[phases.id]` clause is about the case there wouldn't be a matching vocabulary_term (perhaps a wrong ID), so you return the original object at least

Crashillo (changes requested):

app/javascript/gobierto_investments/webapp/components/MapTour.vue
entantoencuanto: If I don't initialize the middleware the MapTour fails, because somewhere there is a call to `getPhases` and the `getAttributesByKey` method is not available

PR #3286, opened 2020-08-06
3251 protect api with password

### LINKED ISSUE #3251: Apply sites password protection to the API
When sites are password protected their API is still public. We should apply that authentication step also to the API, it's a http auth, so for api clients it's easy to implement and remove when the site becomes public.

entantoencuanto:
Closes #3251

## :v: What does this PR do?

* Adds basic auth to each action of API if site is protected with password.
* This filter is skipped if the user provides a valid user or admin authentication token (the user must be active or admin must have permissions on site)
* Adds tests to site navigation and API

Nothing should change in gobierto data, investments and plans

No

Suggestion sent to gobierto.readme.io

### DIFF
--- a/app/controllers/api_base_controller.rb
+++ b/app/controllers/api_base_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ApiBaseController < ActionController::API
+  include ActionController::HttpAuthentication::Basic::ControllerMethods
   include SubmodulesHelper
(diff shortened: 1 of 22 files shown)


ferblape:
- have you send changes to the documentation?
- I miss some tests

ferblape (changes requested):
I miss:

- a clarification in the documentation (both in how the module works and the api docs)
- tests

PR #3292, opened 2020-08-10
autofocus on admin modals

### LINKED ISSUE #3244: Admin / Focus on the first input when creating or editing a vocabulary term
When you go to create a new term or edit an existing one, the cursor should be focused on the first input (term name)

imagen

Crashillo:
Closes #3244 

## :v: What does this PR do?
Autofocus to the first available `input[type="text"]` when the modal opens


furilo:
Does this also apply to new item creation?

Crashillo:
Applies to the first input[type=text] the modal has. Therefore, not only in plans, but any other module that triggers a modal will be focused on its first input-text, if any.

ferblape:
👍

PR #3322, opened 2020-09-04
2686 rails upgrade 6 defaults

### LINKED ISSUE #2686: Upgrade to Rails 6
Migrate to the latest version of Rails 6 published. This guide would help: https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0

Probably it'll be very difficult to activate Zeitwerk (see https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#autoloading), so it could be splitted in two PRs:

- first, migrate to Rails 6.x with the traditional autoloading style
- second, activate _6.x settings_ to use Zeitwerk

entantoencuanto:
Closes #2686

## :v: What does this PR do?

This PR includes the Rails 6 defaults except autoloading with Zeitwerk and ActiveStorage defaults

This PR is built on top of #3291 and only enables the 6.0 defaults. This is done in a separated PR because introduces breaking changes and should be merged after checking that everything is OK with the previous PR.

- Adds an initialization file for Rails 6 defaults. At the end of the process this file can be deleted and set directly in `config/application.rb` `config.load_defaults "5.2"` 

No

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -7,4 +7,6 @@ production:
   - default
   - user_verifications
   - mailers
+  - active_storage_analysis
+  - active_storage_purge
   - algoliasearch
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -7,6 +7,4 @@ production:
   - default
   - user_verifications
   - mailers
-  - active_storage_analysis
-  - active_storage_purge
   - algoliasearch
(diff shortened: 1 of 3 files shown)


config/sidekiq.yml
```diff
@@ -7,4 +7,6 @@ production:
   - default
   - user_verifications
   - mailers
+  - active_storage_analysis
```
ferblape: these queues are not necessary for us

ferblape (changes requested):
Just review the queue names, the rest looks ok!

config/sidekiq.yml
entantoencuanto: Agh, you're right, it comes from a test I made and I missed to delete it

PR #3328, opened 2020-09-07
Update Amazon SES signature to v4

ferblape:
## :v: What does this PR do?

This PR updates the version of AWS SES gem to support v4 signature ([more info about this](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html)) and stop using v3, which will stop working on Sep 30th

Test emails are delivered properly


ferblape:
@entantoencuanto ping

entantoencuanto (approved):

PR #3343, opened 2020-09-14
Enable CORS support in API requests

ferblape:
##  :v: What does this PR do?

This PR enables CORS support in API requests. CORS is disabled by default in Rails, and you have to manually enable which urls are aimed to be used by remote requests.

You can test remote calls from a Javascript snippet to the data api of a site password protected.


ferblape:
@entantoencuanto I need to merge this because we need it in a project, so I'm going to merge it!

entantoencuanto:
Oh, it seemed to work fine in staging, sorry for not giving feedback

PR #3347, opened 2020-09-15
Parse frecuencies and categories

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1111

## :v: What does this PR do?
The error was produced due to the API changes regarding frecuencies and categories format

Check out: https://getafe.gobify.net/datos - https://demo-datos.gobify.net/datos

## :eyes: Screenshots
imagen

Incomplete fields
imagen


ferblape (changes requested):
The approach is not correct, because Category and Frequency are custom fields, that might exist, or might not. See for example https://demo-datos.gobify.net/datos/, where they don't exist and the code is broken

PR #3366, opened 2020-09-22
3353 extend api token model

### LINKED ISSUE #3353: Extend the current api tokens model to support domain specific tokens and a name
Extend the current api tokens model to include:

- name: optional, to help users identify/describe tokens
- domain: optional. When the domain is provided, the request should only perfom if the origin is the same as the token domain.

Requests that require a token but the origin is a site of the current instance won't require a token. This will allow us to remove this code for js apps: https://github.com/PopulateTools/gobierto/blob/master/app/views/layouts/_gobierto_footer.html.erb#L35-L37

entantoencuanto:
Closes #3353

## :v: What does this PR do?

Changes the token models and API behaviour with them:
* Adds a `domain` attribute allowed to be blank
* Changes the `name` attribute allowing blank names for the not primary tokens of a user (if the name is present it can't be duplicated for a same user/admin)
* When a site is protected with password:
  * If the request is internal (the origin host is in the list of sites of the Gobierto instance) no token is required for the open read endpoints (like the datasets index) and if a token is provided the domain of the token is ignored
  * If the request comes from an external site a token is required for all requests and if the token does not exist or has an associated domain which differs from the request origin host the request is unauthorized
* If the site is not protected with password the public endpoints don't require token but the domain is verified in the actions which require authentication when the token has a domain defined.
* Removes the basic_auth_token references in js front apps and application footer because it's because it is no longer needed after consider these calls as internal
* Adds some API tests

Currently the tokens don't have interface, WIP in an upcoming PR

No

- Pending: Documentation has to be updated

### DIFF
--- a/app/controllers/concerns/user/api_authentication_helper.rb
+++ b/app/controllers/concerns/user/api_authentication_helper.rb
@@ -21,8 +21,10 @@ module User::ApiAuthenticationHelper
     raise_unauthorized unless user_authenticated?
   end
 
+  # The host is not checked if request is internal or the site is not password
+  # protected
   def check_host
-    return if internal_site_request? || !site_protected? || token.present? && valid_token_domain?
+    return if internal_site_request? || !site_protected? || api_token_with_host.present?
 
     raise_unauthorized
   end
@@ -33,25 +35,15 @@ module User::ApiAuthenticationHelper
     authenticate_user_in_site unless internal_site_request?
   end
 
-  def valid_token_domain?
-    return unless api_token.present?
-
-    api_token.domain.blank? || api_token.domain == request.host
-  end
-
   def internal_site_request?
-    Site.where(domain: request.host).exists?
+    @internal_site_request ||= Site.where(domain: request.host).exists?
   end
 
   def find_current_user
-    return unless internal_site_request? || valid_token_domain?
-
     current_site.users.confirmed.joins(:api_tokens).find_by(user_api_tokens: { token: token })
   end
 
   def find_current_admin
-    return unless internal_site_request? || valid_token_domain?
-
     @current_admin ||= ::GobiertoAdmin::Admin.joins(:api_tokens).find_by(admin_api_tokens: { token: token })
   end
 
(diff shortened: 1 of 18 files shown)


app/controllers/concerns/user/api_authentication_helper.rb
```diff
@@ -21,17 +21,37 @@ def authenticate_user!
...
+    authenticate_user_in_site unless internal_site_request?
+  end
+
+  def valid_token_domain?
+    return unless api_token.present?
+
+    api_token.domain.blank? || api_token.domain == request.host
+  end
+
+  def internal_site_request?
```
ferblape, line 38: In a request to gobierto_data this query is executed twice (the second with cache), but it'll be faster if you memoize the method:

```
D, [2020-09-25T07:52:36.042772 #652962] DEBUG -- : [getafe.gobify.net] [3fdb38ee-62cc-4e63-8c5b-24130b49fed0]   Site Exists? (0.8ms)  SELECT 1 AS one FROM "sites" WHERE "sites"."domain" = $1 LIMIT $2  [["domain", "getafe.gobify.net"], ["LIMIT", 1]]
D, [2020-09-25T07:52:36.044871 #652962] DEBUG -- : [getafe.gobify.net] [3fdb38ee-62cc-4e63-8c5b-24130b49fed0]   CACHE Site Exists? (0.0ms)  SELECT 1 AS one FROM "sites" WHERE "sites"."domain" = $1 LIMIT $2  [["domain", "getafe.gobify.net"], ["LIMIT", 1]]
```

app/controllers/concerns/user/api_authentication_helper.rb
```diff
@@ -21,17 +21,37 @@ def authenticate_user!
     raise_unauthorized unless user_authenticated?
   end
 
+  def check_host
```
ferblape, line 26: The method name is confusing to me, I don't understand the difference between it and `authenticate_in_site`

ferblape (changes requested):
Ping me

ferblape (approved):

PR #3385, opened 2020-09-30
Solve accessibility issues

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1115

## :v: What does this PR do?

- Remove duplicate `id`
- Include the `alt` attribute in those images where it's necessary. If the image is only decorative isn't required, for example, illustrations.
- Add `role="presentation"` to some `tables`, and remove in `th`.
- Refactor HTML, the tag `a` should always go inside the tag `button`
- Make the dropdown/select accessible so that it can be navigated with the keyboard. 
- Remove non-existent attributes like `aria-controlledby`

### Before this PR

navigation-not-work

### After this PR

select_navigation_keyboard

### DIFF
--- a/app/javascript/gobierto_budgets/modules/application.js
+++ b/app/javascript/gobierto_budgets/modules/application.js
@@ -108,6 +108,28 @@ $(document).on('turbolinks:load', function() {
     $('.line_browser').velocity("fadeOut", { duration: 50 });
   });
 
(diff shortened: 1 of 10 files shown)


Crashillo (approved):

ferblape:
I'm going to merge this, if you find new issues in plans let's open a new issue @jorgeatgu

PR #3390, opened 2020-10-01
Gobierto Data / Include visualizations in dataset home

### LINKED ISSUE #3364: Data / Include visualizations in dataset home
We whould add a block "Visualizations" in the dataset has some, the same we do with Consultas

https://getafe.gobify.net/datos/censo-edad-nacionalidad-procedencia/resumen

jorgeatgu:
Closes #3364 

## :v: What does this PR do?

- Include visualizations as a section in dataset home

Staging
### After this PR


furilo:
Can we keep consistency of which headers we show? It seems in Queries we don't show the "Yours" if you don't have any, but in visualizations we do the contrary. 

imagen

jorgeatgu:
Done! 

Screenshot 2020-10-02 at 18 53 47

app/assets/stylesheets/module-data.scss
```diff
@@ -1263,6 +1263,13 @@
 
 }
 
+.gobierto-vue-dropdown--content {
+  .gobierto-data-sets-nav--tab-container {
```
Crashillo: Avoid nesting if you can

Crashillo (approved):

ferblape (commented):
Code is ok, but the way that visualization block loads is a bit confusing, because the spinner doesn't appear until you know there are visualizations. See this video:

dataset visualizations

The user doesn't know that visualizations are loading there...

jorgeatgu:
I made a few changes, now includes a spinner until the visualizations are loaded.
Screen Recording 2020-10-07 at 11 45 49 2020-10-07 11_51_27

ferblape (approved):

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -25,6 +25,26 @@
       <!-- Only is mounted where there are attributes -->
       <SummaryTab
         v-if="activeDatasetTab === 0 && attributes"
+        :dataset-id="datasetId"
+        is-viz-saving-prompt-visible="isVizSavingPromptVisible"
```
Crashillo: Is this working? You miss the colon

Crashillo (approved):

app/javascript/gobierto_data/webapp/components/Dataset.vue
jorgeatgu: Yes, it's working because this prop is unnecessary.

PR #3395, opened 2020-10-05
Gobierto Data / Update URL when selecting a category

### LINKED ISSUE #3363: Data / Permalink when selecting a category
When selecting a category/vocab term in one of the blocks in the sidebar, update the browser URL with a permalink for that category. Think how the URL should work, as you can have selected a combination of various terms from the same vocab, or from various vocabs (and remember that vocabs can be dinamically configured). 

URL might be something like

https://getafe.gobify.net/datos/terms/123:124:125

Where 123, 124 etc are the term IDs (which might belong to different vocabs - if more info is needed to reference the vocab that's ok)

(using ? or / doesn't matter much, whatever is easier)

imagen

jorgeatgu:
Closes https://github.com/PopulateTools/gobierto/issues/3363

## :v: What does this PR do?

When the user clicks on one or more of the sidebar filters, we update the URL, including the ids of the selected filters.

Staging

### After this PR

Screen Recording 2020-10-07 at 12 18 27 2020-10-07 12_19_32

### DIFF
--- a/app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
+++ b/app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
@@ -111,10 +111,6 @@ export default {
       this.checkSelectedCheckbox()
     },
     updateURLwithSelectedCategories(values) {
-      if (this.$route.name === 'Dataset') {
-        // eslint-disable-next-line no-unused-vars
-        this.$router.push('/datos/').catch(err => {})
-      }
 
       const { options: optionsChecked } = values
       const { key } = values
@@ -142,16 +138,9 @@ export default {
       routeItems = [...new Set(routeItems)];
       routeItems = routeItems.toString().replace(/,/gi, '')
 
-      //If the user accesses through a permalink we change the route
-      let backSlashRoute = this.$route.fullPath === '/datos' ? `${this.$route.path}/terms/` : `${this.$route.path}terms/`
-      let urlTerms = this.isPermalinkActive ? `${location.origin}/datos/terms/` : backSlashRoute
-      urlTerms = routeItems.length === 0 ? `${this.$route.path}` : urlTerms
+      // eslint-disable-next-line no-unused-vars
+      this.$router.push(`/datos/terms/${routeItems}`).catch(err => {})
 
-      history.pushState(
-        {},
-        null,
-        `${urlTerms}${routeItems}`
-      )
     },
     getElements(ids) {
       let list = []
@@ -166,15 +155,16 @@ export default {
     reFilterElements(category) {
       let elements = []
       elements = this.filters.filter(({ key }) => key === category)
-      const { options } = elements[0]
+      const [{ options } = {}] = elements || []
       const isItemCounter = options.filter(({ isOptionChecked, counter }) => isOptionChecked === true && counter > 0)
       const getIdFromItemsCounter = [...new Set(isItemCounter.map(({ id }) => id))]
       elements = this.getElements(getIdFromItemsCounter)
       return elements
     },
     selectedCheckbox(values) {
-      //When the user accesses through a permalink we capture the selected elements, select them and filter the datasets
+      //Remove undefined values to prevent error when converting the rest into an array of numbers.
       const categoriesSelected = values.filter(item => item).map(item => +item);
+      //When the user accesses through a permalink we capture the selected elements, select them and filter the datasets
       for (let item of this.filters) {
         item.options.forEach((d) => {
           if (categoriesSelected.includes(d.id)) {
--- a/app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
+++ b/app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
@@ -5,7 +5,7 @@
       class="gobierto-data-filters"
     >
       <div
-        v-for="filter in filtersModify"
+        v-for="filter in filters"
         :key="filter.title"
         :class="!filter.isToggle ? 'gobierto-data-filters-element-no-margin' : ''"
         class="gobierto-data-filters-element"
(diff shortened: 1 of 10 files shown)


furilo:
If you access to getafe.gobify.net/datos (without the last /) the new parameters are added without the needed / ->

getafe.gobify.net/datosterms/4629:4620:

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
```diff
@@ -79,22 +85,114 @@ export default {
...
+      //If the user accesses through a permalink we change the route
+      let backSlashRoute = this.$route.fullPath === '/datos' ? `${this.$route.path}/terms/` : `${this.$route.path}terms/`
+      let urlTerms = this.isPermalinkActive ? `${location.origin}/datos/terms/` : backSlashRoute
+      urlTerms = routeItems.length === 0 ? `${this.$route.path}` : urlTerms
+
+      history.pushState(
+        {},
+        null,
+        `${urlTerms}${routeItems}`
+      )
```
Crashillo: Is vue-router notified somehow? Mixing these things causes issues often

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
```diff
@@ -79,22 +85,114 @@ export default {
...
+          let item = `${ids[index]}:`
+          list.push(item)
+        }
+      }
+      return list
+    },
+    reFilterElements(category) {
+      let elements = []
+      elements = this.filters.filter(({ key }) => key === category)
+      const { options } = elements[0]
```
Crashillo: Destructuration level 2 (with fallbacks)
```suggestion
      const [{ options } = {}] = elements || []
```

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
```diff
@@ -79,22 +85,114 @@ export default {
...
+      const isItemCounter = options.filter(({ isOptionChecked, counter }) => isOptionChecked === true && counter > 0)
+      const getIdFromItemsCounter = [...new Set(isItemCounter.map(({ id }) => id))]
+      elements = this.getElements(getIdFromItemsCounter)
+      return elements
+    },
+    selectedCheckbox(values) {
+      //When the user accesses through a permalink we capture the selected elements, select them and filter the datasets
+      const categoriesSelected = values.filter(item => item).map(item => +item);
```
Crashillo: `filter(item => item)` ?

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
jorgeatgu: Nope, to avoid errors i've replaced it.

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
jorgeatgu: Is to prevent any error after split array returns:
`paramsRouter Array(4) [ "3665", "3680", "3671", "" ]`

 If I don't filter when converting the array of strings into an array of numbers js it converts `""` into `0`

`categoriesSelected Array(4) [ 3665, 3680, 3671, 0 ]`

I should have commented on this in the code, sorry.

Crashillo: use `reduce` then
```suggestion
      const categoriesSelected = values.reduce((acc, item) => { if (item) acc.push(+item); return acc}, [])
```

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
```diff
@@ -23,19 +23,18 @@
               <a
                 class="gobierto-block-header--link"
                 @click.stop="e => selectAllCheckbox_TEMP({ ...e, filter })"
-              >{{
-                filter.isEverythingChecked ? labelNone : labelAll
-              }}</a>
+              >{{ checkIfEverythingChecked(filter.isEverythingChecked) }}</a>
```
Crashillo, line 26: I don't get this: you're calling a function<sup>1</sup> to check if a boolean value (filter.isEverythingChecked) is true o false, in order to display one label or another

Why don't you simply use a `<template v-if="filter.isEverythingChecked">...`??


<sup>1</sup> Bad practice. You should never call functions in the template

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
```diff
@@ -79,22 +85,108 @@ export default {
       return ((this.filters[0] && (this.filters[0] || {}).count >= 1)
         || (this.filters[1] && (this.filters[1] || {}).count >= 1));
     },
-    filtersModify() {
-      return this.filters.length ? this.filters.map(d => ({ ...d, isToggle: true })) : []
+    isAllChecked() {
+      return this.routeItemsFrequency.length !== 0 || this.routeItemsCategory.length !== 0
```
Crashillo, line 89: `length` is always numeric for objects, then you can just:

```suggestion
      return this.routeItemsFrequency.length || this.routeItemsCategory.length
```

app/javascript/gobierto_data/webapp/components/sidebar/SidebarCategories.vue
```diff
@@ -79,22 +85,108 @@ export default {
...
+    }
+  },
+  created() {
+    if (this.$route.params.pathMatch !== undefined) {
```
Crashillo, line 93: Similar
```suggestion
    if (this.$route.params.pathMatch) {
```

Crashillo (changes requested):

PR #3430, opened 2020-10-20
Reduce JS translations bundle size

### LINKED ISSUE #2687: Investigate if we can reduce translations file size in JS
@Crashillo has been investigating a bit how to reduce the size of the JS packs by reducing the translations file:

- check the namespaces in ignore-missing and ignore-unused. These options were filled with keys used in JS because Ruby wasn't able to detect them. It seems that it has been improved and we can check again

- investigate if we can skip English translations when dumping the JS file

mllocs:
Closes https://github.com/PopulateTools/gobierto/issues/2687

## :v: What does this PR do?

It creates a javascript translation file for each locale. Then, using webpack's dynamic loading it creates a bundle for each locale and only loads the translations of the current locale. 

By doing this we'll be saving around **86Kb of gzipped Javascript** (downloading and parsing). On the other hand we are going to do an extra request since the translations bundle will be loaded "on demand" (handled by webpack). 

- Check that the frontend translations work (make sure that you go to a place where frontend translations are used, most of the views use backend provided translations).

- Check changing your locale.

## :eyes: Screenshots

These screenshots were generated with webpack-analyzer-plugin in production mode.

### Before this PR

Screen Shot 2020-10-20 at 09 00 21

### After this PR

Screen Shot 2020-10-20 at 09 05 31
Screen Shot 2020-10-20 at 09 05 38
Screen Shot 2020-10-20 at 09 05 44

🧘‍♂️ 

🧘‍♂️

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,8 +53,7 @@ yarn-debug.log*
 .yarn-integrity
 
 # I18njs
-app/javascript/i18n/modules/translations.js
-app/javascript/lib/i18n/modules/translations.js
+app/javascript/lib/i18n/modules/*
 
 # Gobierto Engines
 vendor/gobierto_engines/*
--- a/app/javascript/lib/i18n/index.js
+++ b/app/javascript/lib/i18n/index.js
@@ -1,1 +1,1 @@
-import './modules/translations.js'
+import(`./modules/translations_${I18n.locale}`);
--- a/config/i18n-js.yml
+++ b/config/i18n-js.yml
@@ -27,4 +27,5 @@
 #
 export_i18n_js: false
 translations:
-  - file: "app/javascript/lib/i18n/modules/translations.js"
+  - file: "app/javascript/lib/i18n/modules/translations_%{locale}.js"
+    only: '*'


Crashillo:
Hi there, @mllocs 

Thanks for your contribution.

Unfortunately, we cannot merge this PR, because javascript literals are missing. It's kinda tricky, but this application gathers some new developments (JS-modules) with legacy server-side code (as ERB files). And there, is where the current locale is set. So, you won't noticed which actual value have `I18n.locale` after _DOMContentLoaded_ event finishes, therefore, in `import(`./modules/translations_${I18n.locale}`);`, _I18n.locale_ will be always the default value (`en`) until the DOM is loaded.
Those literals which have been fetched from ERB files, they will display properly, but the javascript code splitting process will fetch only `config/translations_en.js` file, however the locale has been changed onload, therefore the javascript literals now point to an inexistent file.
In order to make your contribution work, we would need to refactor old legacy modules (their way of loading)

Anyways, thanks for your interest & happy coding!

mllocs:
Hello @Crashillo, thanks for your answer! I didn't pick an easy issue to start with but it helped me a lot understanding the architecture of the app, plus, I did a similar task recently in another project. I'll try checking out what you said and see if there's something that can be done to make it work. 😊

PR #3446, opened 2020-10-30
Add GeoJSON world countries

jorgeatgu:
Closes #N

## :v: What does this PR do?

Add a GeoJSON with the countries of the world, format: ISO 3166-1 alpha 2 code.


ferblape (changes requested):
Instead of loading it from public, with Webpacker it could be included with a require using a loader, right? This way we could move the file to gencat-engine and avoid an extra request when the page is loaded

Can you confirm if it's possible?

jorgeatgu:
Another way without WebPacker, import `geojson` as a module.

ferblape:
Ok

PR #3511, opened 2020-11-25
Gobierto Data / Transforms Booelan values and show errors

### LINKED ISSUE #3434: Gobierto Data / Saving an invalid query should show the error to the user
When a user tries to save a query but the SQL is invalid the query can't be saved because of an error in the server. That error should be captured and displayed to the user, which right now dowesn't receive any feedback.

Steps to reproduce:

1. login as use in any gobierto instance
2. go to a dataset
3. write an incorrect query
4. click "Save" and see nothing happens

### LINKED ISSUE #3436: Gobierto Data / Boolean column is always represented with false in perspective
No matter the value of the boolean column, Perspective renders a "false" in the cell. See this screenshoot, where I'm filtering by rows which `is_forecast=TRUE`:

Screenshot 2020-10-22 at 07 46 26

I've also checked that in the dataset schema the column type is boolean

jorgeatgu:
Closes #3436 
Closes #3434

## :v: What does this PR do?

Postgresql returns boolean values as `t` and `f`, doesn't understand that `f` is `false` and `t` is `true,` so we have to transform those values.

When the user tries to save an invalid query, nothing happens. Now we catch the error and show it.

Staging

## Screenshots

Show error

Screen Recording 2020-11-26 at 08 15 43 2020-11-26 08_19_30

### DIFF
--- a/app/javascript/gobierto_data/webapp/components/Dataset.vue
+++ b/app/javascript/gobierto_data/webapp/components/Dataset.vue
@@ -657,7 +657,7 @@ export default {
     },
     setPublicQueries(response) {
       const {
-        data: { data: items = [] },
+        data: { data: items = [] }
       } = response;
 
       this.publicQueries = items
@@ -710,16 +710,7 @@ export default {
         try {
           ({ status } = await this.putQuery(queryId, { data }));
         } catch (error) {
-          const {
-            response: {
-              data: {
-                errors: arrayError
-              }
-            }
-          } = error;
-          const [ sqlError ] = arrayError
-          const { detail } = sqlError
-          this.queryError = detail
+          this.queryError = this.destructuringStoreQueryError(error)
         }
 
         //Update revert query
@@ -729,16 +720,7 @@ export default {
         try {
           ({ status, data: { data: newQuery } } = await this.postQuery({ data }));
         } catch (error) {
-          const {
-            response: {
-              data: {
-                errors: arrayError
-              }
-            }
-          } = error;
-          const [ sqlError ] = arrayError
-          const { detail } = sqlError
-          this.queryError = detail
+          this.queryError = this.destructuringStoreQueryError(error)
         }
       }
 
@@ -784,16 +766,7 @@ export default {
         this.isQueryRunning = false;
         this.getColumnsQuery(this.items)
         this.queryError = null
-      } catch (error) {
-        const {
-          response: {
-            data: {
-              errors: arrayError
-            }
-          }
-        } = error;
-        const [ sqlError ] = arrayError
-        const { sql: stringError } = sqlError
+      } catch ({ response: { data: { errors: [{ sql: stringError } = {}] = [] } } } ) {
         this.queryError = stringError
         this.isQueryRunning = false;
         this.enabledQuerySavedButton = false
@@ -1085,6 +1058,10 @@ export default {
       this.isVizModified = false
       this.activatedSavedVizButton(value)
       this.showPrivatePublicIconViz = true
+    },
+    destructuringStoreQueryError(error) {
+      const { response: { data: { errors: [{ detail } = {}] = [] } } } = error
+      return detail
     }
   },
 };
--- a/app/javascript/gobierto_data/webapp/components/Dataset.vue
+++ b/app/javascript/gobierto_data/webapp/components/Dataset.vue
@@ -715,10 +715,14 @@ export default {
         } catch (error) {
           const {
             response: {
-              statusText
+              data: {
+                errors: arrayError
+              }
             }
           } = error;
-          this.queryError = statusText
+          const [ sqlError ] = arrayError
+          const { detail } = sqlError
+          this.queryError = detail
         }
 
         //Update revert query
@@ -730,10 +734,14 @@ export default {
         } catch (error) {
           const {
             response: {
-              statusText
+              data: {
+                errors: arrayError
+              }
             }
           } = error;
-          this.queryError = statusText
+          const [ sqlError ] = arrayError
+          const { detail } = sqlError
+          this.queryError = detail
         }
       }
 
--- a/app/javascript/gobierto_data/webapp/components/Dataset.vue
+++ b/app/javascript/gobierto_data/webapp/components/Dataset.vue
@@ -648,10 +648,7 @@ export default {
     },
     setPrivateQueries(response) {
 
-      const {
-        data: { data: items },
-        status
-      } = response;
+      const { data: { data: items } } = response;
       this.privateQueries = items;
     },
     async getPublicQueries() {
--- a/app/javascript/gobierto_data/webapp/components/Dataset.vue
+++ b/app/javascript/gobierto_data/webapp/components/Dataset.vue
@@ -647,8 +647,10 @@ export default {
       });
     },
     setPrivateQueries(response) {
+
       const {
         data: { data: items },
+        status
       } = response;
       this.privateQueries = items;
     },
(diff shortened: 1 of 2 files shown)


ferblape:
Hmmm, about the error in a query, I don't agree with what you call the error message ;)

For example, if I write this wrong query the error message is No Reason Phrase (??):

Screenshot from 2020-11-25 22-01-00

While the API returns a different error message:

Screenshot from 2020-11-25 22-01-11

Some screenshots in the PR would help be appreciated too

jorgeatgu:
My mistake. I thought the error in executing an invalid query was the same as trying to save an invalid query, but it is a little different.

Update:
Screen Recording 2020-11-26 at 08 15 43 2020-11-26 08_19_30

ferblape (approved):

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -708,13 +710,39 @@ export default {
...
+          const {
+            response: {
+              data: {
+                errors: arrayError
+              }
+            }
+          } = error;
+          const [ sqlError ] = arrayError
+          const { detail } = sqlError
+          this.queryError = detail
```
Crashillo: In one line 😄 
```suggestion
        } catch ({ response: { data: { errors: [{ detail } = {}] = [] }}}) {
          this.queryError = detail
```

app/javascript/gobierto_data/webapp/components/Dataset.vue
```diff
@@ -708,13 +710,39 @@ export default {
...
```
Crashillo: DRY

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -92,10 +92,12 @@ export default {
       /* We compare the columns, if it returns true we pass the schema to Perspective, if false Perspective will convert the columns*/
       const sameColumns = arrayColumnsFromAPI.length === arrayColumnsFromQuery.length && arrayColumnsFromAPI.every(column => arrayColumnsFromQuery.includes(column))
 
+      //Postgresql returns booleans values as "t" and "f", so we need to transform them, "t" as "true" and "f" as false.
+      const replaceBooleanValues = this.items.replace(/"t"/g, '"true"').replace(/"f"/g, '"false"')
```
Crashillo: Does`this.items` contain only Booleans?

Crashillo (changes requested):
Actually I don't like this way to face it the Boolean issue (it happened the same for ISCIII). We shouldn't be parsing the Boolean values all the time. I think this problem belongs to postgres response or perhaps the CSV conversion.

Thoughts, @ferblape @jorgeatgu ?

Crashillo:
FYI, I edited the OP to fix the _Closes_ triggers. They should be in the first line of the issue you wanna close. At the sidebar, now you can _link_ these issues as well

jorgeatgu:
> Actually I don't like this way to face it the Boolean issue (it happened the same for ISCIII). We shouldn't be parsing the Boolean values all the time. I think this problem belongs to postgres response or perhaps the CSV conversion.
> 
> Thoughts, @ferblape @jorgeatgu ?

Postgresql always return `true` as `t` and `false` as `f` @entantoencuanto 

I don't like this solution either, but I think it's the only one.

Crashillo:
What if? `SELECT true::boolean::text, false::boolean::text`

https://www.postgresql.org/message-id/CACfv+pJbG5fLWOgSfZaTKR0_w3Rbx3XqTphD7Pndp3ygvEOB-A@mail.gmail.com

Crashillo:
It's OK that PostgresQL deals with 't'/'f'. That's not the point. But the issue I believe it is at the conversion to JSON/CSV

jorgeatgu:
Another approach:

The first step, check if the columns contain `Boolean` values https://github.com/PopulateTools/gobierto/blob/74552faadd1858e6b0387124bac25cbcefbb0423/app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue#L96

If columns contain `Boolean`, goes to transform only these columns

https://github.com/PopulateTools/gobierto/blob/74552faadd1858e6b0387124bac25cbcefbb0423/app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue#L214-L234

And the last step: convert `JSON` to `CSV`

Test:

Screenshot 2020-11-26 at 12 38 46

Crashillo (commented):
So we're converting String -> JSON -> CSV, just for check Boolean values. Interesting. Could you please, test perfomance with large files?

Crashillo:
I would save those transformations. I'd replace it straight outta file 🏋🏾‍♀️. For example, `this.items` has:
```csv
\"España\",\"espana\",\"8460261\",\"0.0\",,,,,,,,,,,,,,,\"f\"
```
So, try something like `this.items.replaceAll("\"f\"", false)` instead of all those hacky-tricky-diabolic transformations

ferblape:
Ping me to discuss

Context: we can't change how postgres returns booleans, and we can't put a Ruby layer in the middle, because we lose all the speed of Postgres, so it must be done in the client.

My proposal: would it make sense to update the boolean values here only if the column type is boolean? https://github.com/PopulateTools/gobierto/pull/3511/commits/7653b4669c30b7ecc67ac269df063740c242eb72#diff-39e9cd514806cd05c4df2cddbce712c0761c3b080515df459d07bc60442b35afL107-L120

Crashillo:
> would it make sense to update the boolean values here only if the column type is boolean?

@jorgeatgu do we know what sort of columns (datatypes) we have BEFORE initialize perspective? If so, the only thing (I believe) we can do is about checking whether any column is boolean, therefore, we apply the `replace` clause. In other words, only replace if some column is boolean.
But I'm afraid we cannot replace ONLY booleans because the input is a large string (plain text)

jorgeatgu:
@ferblape 

@Crashillo Yes, the API returns the type of the columns.

Right now this [function](https://github.com/PopulateTools/gobierto/blob/74552faadd1858e6b0387124bac25cbcefbb0423/app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue#L204-L212) checks if some column has a `boolean` value, if contains some Boolean columns another function _hacky-tricky-diabolic transformations_  to `t` and `f` to `true` and `false`.

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -92,10 +92,16 @@ export default {
       /* We compare the columns, if it returns true we pass the schema to Perspective, if false Perspective will convert the columns*/
       const sameColumns = arrayColumnsFromAPI.length === arrayColumnsFromQuery.length && arrayColumnsFromAPI.every(column => arrayColumnsFromQuery.includes(column))
 
+      //Check if any columns contains Boolean values
+      const datasetsHasBooleanColumns = this.checkIfBooleanExist(this.objectColumns)
+
+      //If columns contains Boolean values goes to replace them
+      let replaceItems = datasetsHasBooleanColumns.length > 0 ? this.replaceBooleanValues(this.items) : this.items
```
Crashillo: You complicate things a lot

```suggestion
      let replaceItems = this.items
      if (this.objectColumns.some(({ value }) => value === "boolean" )) {
        replaceItems = this.items.replace(/"t"/g, '"true"').replace(/"f"/g, '"false"')
      }
```

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
Crashillo: And remove the `checkIfBooleanExist`'n'`replaceBooleanValues` functions, sure

Crashillo (approved):

ferblape (approved):

Issue #3421, opened 2020-10-16
Data / Hide JSON and XLS options from custom queries download menu


furilo:
Leave only the CSV option

![imagen](https://user-images.githubusercontent.com/24740/96237281-698c0c00-0f9d-11eb-8542-361e0158b826.png)

Issue #3482, opened 2020-11-10
Data / Disable JSON and XLSX formats


ferblape:
Related with #3421 

The generation of these dataset dumps is causing memory leaks in production servers, and since we are not going to use them I'd suggest to disable them from the code (don't remove, just keep the code but don't use the format):

- [ ] Remove those formats from the API responses
- [ ] Remove from the dataset cache generation
- [ ] Update docs

PR #3512, opened 2020-11-25
Gobierto Data / Hide JSON and XLS options from the download menu

jorgeatgu:
Closes 

https://github.com/PopulateTools/gobierto/issues/3421

## :v: What does this PR do?

Hide JSON and XLS options from the download menu

Staging

### DIFF
--- a/app/javascript/gobierto_data/webapp/components/Dataset.vue
+++ b/app/javascript/gobierto_data/webapp/components/Dataset.vue
@@ -327,10 +327,11 @@ export default {
       formats: arrayFormats
     } = attributes;
 
+    const { csv: csvURL } = arrayFormats
     this.titleDataset = titleDataset;
     this.tableName = tableName;
     this.objectColumns = objectColumns;
-    this.arrayFormats = arrayFormats;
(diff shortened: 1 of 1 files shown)


ferblape:
Maybe it wasn't necessary that you updated anything, because when https://github.com/PopulateTools/gobierto/issues/3482 is implemented the formats will be read automatically from the API, right?

jorgeatgu:
Yes, the list is automatically created with the API formats.

Now that we only have one format available, maybe it doesn't make much sense to show a list with only one item. This way, we save the user a click.

Screenshot 2020-11-26 at 08 06 12

furilo:
We'll have more formats in the future, so leave it as it is. If this starts being used a lot we can reconsider your suggestion (which makes total sense)

jorgeatgu:
OK.

ferblape:
Hey, did you revert this changes in staging? To avoid conflicts in the future

jorgeatgu:
> Hey, did you revert this changes in staging? To avoid conflicts in the future

Done!

PR #3542, opened 2020-12-04
1152 custom fields tanslatable attributes

entantoencuanto:
Closes https://github.com/PopulateTools/issues/issues/1152

## :v: What does this PR do?

* Avoids errors in API and admin when custom field of type text/markdown changes from not translatable to translatable and vice versa.
* The API always returns a string and detects for not translatable fields if the payload is a hash and selects a value according to locale
* In the admin interface, in the edit form the values are taken from the payload (if not saved nothing changes in the record):
  * For fields changed from not translatable to translatable the content is stored in current locale and the others are kept blank
  * For fields changed from translatable to not translatable the value for the current locale is used in the text field.
These changes 

Create translatable and not translatable text and long text custom fields. Fill the values on some resource using them. Edit the custom fields and flip the translatable feature.

No

No

### DIFF
--- a/app/controllers/concerns/gobierto_common/custom_fields_api.rb
+++ b/app/controllers/concerns/gobierto_common/custom_fields_api.rb
@@ -40,7 +40,7 @@ module GobiertoCommon
       md_uids = md_custom_fields.pluck(:uid)
       raw_values.transform_values do |attributes|
         transform(attributes, localized_uids, :translate)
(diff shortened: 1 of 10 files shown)


ferblape:
This is already released to staging but there are some tests failing

ferblape (commented):
Code is OK, just review the tests please

PR #3550, opened 2020-12-10
Navigation between list and dashboard

### LINKED ISSUE #3548: Include vue-router in dashboards
If there are multiple dashboards in a module, they render as cards. In order to implement toggling between such view and the specific one, it requires to include the `vue-router`.

Note that dashboards should emit some kind of url-updates events to be managed from the container (usually a different module)

Crashillo:
## :v: What does this PR do?
Implements a toggle between the list and the dashboard.

#### Caveats
Cannot integrate vue-router. Further explanation: https://github.com/PopulateTools/gobierto/issues/3548#issuecomment-742505743


app/javascript/gobierto_dashboards/webapp/ViewerManager.vue
```diff
@@ -1,19 +1,22 @@
...
-        />
-      </div>
-    </template>
-    <!-- display directly the dashboard otherwise -->
-    <template v-else-if="dashboards.length === 1">
-      <Viewer :config="dashboards[0]" />
-    </template>
+    <transition name="fade">
+      <!-- show list if there's more than one -->
+      <template v-if="!currentDashboard && dashboards.length > 1">
```
jorgeatgu: Replace this with `computed`

jorgeatgu (changes requested):

app/javascript/gobierto_dashboards/webapp/ViewerManager.vue
Crashillo: Done

ferblape (approved):

PR #3553, opened 2020-12-10
3370 limit queries

### LINKED ISSUE #3370: Data / Dataset default LIMIT when dataset is not too big
The other day makind a demo I found myself modifying the default query (removing the LIMIT 50 and executing again) because when going to browse the data in the date or do a visualization, the LIMIT 50 makes no sense. 

This got me thinking that if we are going to cache the default dataset (which equals to the default query without the LIMIT) and if the dataset is not too big (so we don't make the user download lots of data) maybe it makes sense to remove the LIMIT 50. 

For example, if a complete dataset is less than XX MBs, we can remove the LIMIT 50 so you already have the complete dataset and can start visualizing it. If its bigger, we leave the LIMIT 50. 

Also, if we have the LIMIT 50 I think we should warn the user (ie. "This is a preview of the data. Download it or modify the default query to work with all the data"). 

And if we do the LIMIT 50 removal, we should change the Perspective call in order to use the cached URL. 

Thoughts?

entantoencuanto:
Closes #3370

## :v: What does this PR do?

* Makes a change to not send the internal `exposed_in_public_api` setting in module meta endpoint
* Adds some settings to gobierto data API which can be managed in admin:
  * `max_dataset_size_for_queries`: The maximum size in MB to not limit by default the results of the initial data query of a dataset. The size of the dataset will be taken from the csv result of `SELECT *`. When 0 the size is not considered
  * `default_limit_for_queries`: The limit to be applied if the size of the dataset is higher than max_dataset_size_for_queries
  * `exposed_in_public_api`: When true the previous settings are exposed in the module meta endpoint
* Adds a new jsonb column to datasets table to store the size of cached files of a dataset to download. The value consist on a hash with the size of each file for different formats created. This attribute is updated after cached files creation
* Adds 2 new attributes to `meta` response of a dataset:
  *  `size`: The size of each cached file of dataset in MB
  * `default_limit`: The limit to apply if the size of the dataset is too large, considering the API settings. If no limit is required the value is `null`
* Changes the front application to fetch the limit from the API and use it if present

WIP: Pending to add some tests

### After this PR

3370_after

No

Pending: Documentation update


ferblape:
@furilo take a look to the UI to confirm if it's ok to you

furilo:
Several issues: 

- Why are we making "exposed_in_public_api" an option? What's the need for this? I don't see any, thus this only introduces unneeded complexity in the code and UI.
- What does "Configuración públicamente expuesta por la API" mean? I have struggled to understand it only seeing the admin UI -which is what all people will have available. It's difficult to understand without context. Please offer other (shorter) options. Same for the other literals. 
- You have not provided any working example of this feature. I have updated here the max limit (https://getafe.gobify.net/admin/gobierto_data/configuration/settings/edit) but it doesn't seem to be applied in the front end (https://getafe.gobify.net/datos/censo-vehiculos/editor). Please test things properly with examples.

entantoencuanto:
I've leaved only the maximum dataset size configuration in gobierto data API settings, and the limit is fixed to 50. Also the cache expires when the module settings are updated so the changes on settings take effect immediately.

furilo:
> Also the cache expires when the module settings are updated so the changes on settings take effect immediately.

Only? Or also when a dataset is updated? A good way to provide proof (and force you to test it in staging) is to do a Loom...

entantoencuanto:
Yes, also when dataset is updated, but doing the check on staging it seems that in some cases updates of datasets don't trigger the refresh of the cached data correctly and the jobs apparently are lost. If I call the jobs with `perform_now` instead of enqueueing them the default query limit changes as expected

entantoencuanto:
There was a problem with the sidekiq service, once restarted it seems that the updates are done correctly

ferblape (approved):

PR #3562, opened 2020-12-14
Remove schema from version control

ferblape:
## :v: What does this PR do?

Removes schema.rb / structure.sql from the version control, we won't keep or commit that file anymore

No need to test it, just make sure you are aware of it in your topic branches


eltercero (approved):

Crashillo (approved):

entantoencuanto (approved):
It works OK on my local environment

PR #3576, opened 2020-12-17
Routes updated in GobiertoPlans

### LINKED ISSUE #3569: GobiertoPlans vue-router nested childs
The new tabs require another level of nesting for the vue-router in plans.

Currently, the `Home` component includes all nested components, but after including the new tabs, both `Home` (it should be renamed as *Plan*) and `Dashboards` (and another future tabs) must be at the same level.

Crashillo:
Closes #3569 

## :v: What does this PR do?
Handle the router UX in Plans properly:
- Updates the url accordingly the dashboards selected or not
- Load an specific dashboard if comes from url

Besides, it includes a normalization of the routes. The aim is about avoiding to deal with plain-text route names in the code, but only in one place: the router definition.


jorgeatgu (approved):

ferblape (approved):

config/routes.rb
```diff
@@ -460,6 +460,7 @@
         get ":slug(/:year)/proyecto/:id" => "plan_types#show", as: :project
         get ":slug(/:year)/tabla/:uid" => "plan_types#show"
         get ":slug(/:year)/tabla/:uid/:term_id" => "plan_types#show"
+        get ":slug(/:year)/dashboards(/:dashboard_id)" => "plan_types#show"
```
ferblape, line 463: This change should be included in the documentation in order for the module Dashboards to work, right?

app/javascript/gobierto_plans/webapp/components/GroupsByTermTableRow.vue
```diff
@@ -10,7 +10,7 @@
       >
         <i :class="['fas', nestedGroups.length ? `fa-caret-${isOpen ? 'down' : 'right'}` : '']" />
         <router-link
-          :to="{ name: 'term', params: { ...params, term: slug } }"
+          :to="{ name: routes.TERM, params: { ...params, term: slug } }"
```
ferblape, line 13: I like this way of naming routes

app/javascript/gobierto_plans/webapp/lib/router.js
```diff
@@ -22,42 +33,66 @@ export const router = new VueRouter({
...
+                },
+              ]
+            }
+          ],
         },
         {
-          path: "dashboards/:dashboardid?",
-          name: "dashboards",
-          component: Dashboards
+          path: "dashboards/:dashboardId?",
```
ferblape, line 91: Maybe it'd be necessary to document the list of steps to add dashboards to an existing Gobierto module?

1 - add routes to the vue component
2 - create a tab component
3 - add rails routes
...

app/javascript/gobierto_plans/webapp/components/GroupsByTermTableRow.vue
Crashillo: Two best practices here:
1. Avoid text literals in the code
2. At the moment that one variable is reused somewhere else..... move it to shared place

app/javascript/gobierto_plans/webapp/lib/router.js
Crashillo: It'll depend on the module the dashboards are gonna be included. I mean, let's say a new module provides a route, `site.gobify.net/dummy/dashboards`, to render directly the dashboards. You won't need anything (but the rails routes) to make it work.
Vue-router and Tabs are specifications of the one single design, not general.

config/routes.rb
Crashillo: You may take a look https://gobierto.readme.io/docs/gobierto-dashboards

ferblape: 👍🏾

PR #3584, opened 2020-12-18
Dashboards module

### LINKED ISSUE #3453: Dashboards API / dashboards_data
API structure defined here https://docs.google.com/document/d/1goWjuJI73qXMFjENe54-_V-AkYaQED7xAG9M5_ICE9o/edit

## GET /api/v1/dashboards_data?context=module::instance/ID&data_pipe=PIPE_TYPE

* parameters
  * context: module + instance type
  * data_pipe: a type of dashboard_data, for which we define a structure, and the needed transformations to fill this dashboard_data from the data inside the corresponding module. A data_pipe has a corresponding data structure and a widget type that will be used when building a dashboard

Examples of data_pipes

- **&data_pipe=project_metrics**
  * data structure (the one we are using for indicators): name, value, objective, date
  * widget template: metric (what we call "indicator" in the design, pending to be changed)
  * for the moment, we would fill this will all indicators across all projects from a given plan
  * (potential) we could store the progress of the plan, and the progress of each category. For this, on top of adding the project indicators, we would add the global progress, and the progress of each category. Examples: 
    * name, value, objective, date
    * Progreso del plan, 45%, 100%, [current date]
    * Progreso de Área de Acción Climática, 23%, 100%, [current date]
- **&data_pipe=budget_lines**: (this is just an example, we are not integrating it for the time being)
  * (pending)

## GET /api/v1/dashboards_data?context=module::instance/XX&data_pipe=project_metrics

- To populate the indicators tab in a Plan
- Example if context is GobiertoPlans::Plan/123, it would retrieve all indicators from across all plan projects

### LINKED ISSUE #3454: Dashboards API / Dashboards resource
API defined here https://docs.google.com/document/d/1goWjuJI73qXMFjENe54-_V-AkYaQED7xAG9M5_ICE9o/edit

(read first the rest of the issues to understand requirements)

Things the Vue/front-end will need: 

## Dashboards: GET /api/v1/dashboards?context=module::instance/id

Options for URL: 
- To populate the Dashboards list
- As this is not too much information, and the full Dashboard definition will probably be retrieved in an specific endpoint, we may add this in the Plan main endpoint (/api/v1/plans/XX)

## Dashboard POST /api/v1/dashboard (create) - PUT /api/v1/dashboard/XX (update)

To create / update a dashboard.
- requires admin permissions (see #3455)
- dashboard configuration: title, visibility, display of widgets (this depends on the grid system we use) 
- dashboard belongs_to dashboardable (module::instance/id i.e GobiertoPeople::Person/123)
- elements (widgets):
  - HTML: we need to save just the HTML content the user inputs
  - Indicator: we need to save the indicator ID, and the type. We should have "type" templates that represent the values of an indicator according to the template (see proto)
  - in the future we may have other types of widgets that need to store different bits of info. Think of a flexible structure where to save info (JSONB?)
- To Discuss for when we save a dashboard. Options:
  - when a user adds a widget, edits it, or resizes it, it should be saved via ajax as things happen
  - user can add elements, resize them, etc. but are only saved is user explicitly clicks on a SAVE button (design pending for this scenario)

## Dashboard: GET /api/v1/dashboard/XX
To build a dashboard: 
- elements (widgets)
  - HTML
  - Indicator: we'll have to retrieve the actual values of the indicator, and pass them to the template type to paint them. 
- dashboard configuration

entantoencuanto:
Closes #3454
Closes #3453

## :v: What does this PR do?

* Creates a `GobiertoDashboards` module in gobierto
* Adds a `GobiertoDashboards::Dashboard` model
* Adds an dashboards API with basic CRUD of dashboards (only admins can do that) and index, show and data endpoints of a dashboard (the data endpoint is under develompent, it may have change):
  * index: `GET /api/v1/dashboards`. This index allows filter dashboards by context
  * show: `GET /api/v1/dashboards/1`
  * data: `GET /api/v1/dashboards/1/data` This endpoint displays a list of the custom field records values of the fields used by the widgets
  * new: `GET /api/v1/dashboards/new` This endpoint provides a JSON which can be used with values to send to create or update a dataset
  * create: `POST /api/v1/dashboards`
  * update: `PATCH/PUT /api/v1/dashboards/1`
  * delete: `DELETE /api/v1/dashboards/1`
* Adds a `dashboard_data` endpoint which accepts a `context` and a `data_pipe`. For the moment the only available data_pipe is `project_metrics`. This pipe, returns all custom field records of a plan when is set as the context (as described in #3453). The endpoint has this form: `GET /api/v1/dashboard_data?context=GobiertoPlans::Plan/1&data_pipe=project_metrics`
* Adds tests for the dashboards API

WIP: Pending to add some missing tests

Adds the dashboards model and API. 

- [x] new entry in `config/application.yml`? The new module has an entry in this file

Pending documentation of API:

### DIFF
--- a/app/serializers/gobierto_dashboards/base_serializer.rb
+++ b/app/serializers/gobierto_dashboards/base_serializer.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module GobiertoDashboards
+  class BaseSerializer < ActiveModel::Serializer
+    def current_site
+      Site.find(object.site.id)
+    end
+
+    def exclude_links?
+      instance_options[:exclude_links]
+    end
+
+    def exclude_relationships?
+      instance_options[:exclude_relationships]
+    end
+
+    def with_translations?
+      instance_options[:with_translations]
+    end
+
+    def include_draft?
+      instance_options[:include_draft]
+    end
+  end
+end
--- a/app/serializers/gobierto_dashboards/base_serializer.rb
+++ b/app/serializers/gobierto_dashboards/base_serializer.rb
@@ -3,7 +3,7 @@
 module GobiertoDashboards
   class BaseSerializer < ActiveModel::Serializer
     def current_site
-      Site.find(object.site.id)
+      object.site
     end
 
     def exclude_links?
--- a/app/services/gobierto_common/context_service.rb
+++ b/app/services/gobierto_common/context_service.rb
@@ -21,7 +21,7 @@ module GobiertoCommon
     private
 
     def fill_missing_parts(context)
-      return context if context =~ %r{^gid://gobierto//}
+      return context if context =~ %r{^gid://gobierto/}
 
       missing_part = "gid://"
       missing_part += "gobierto/" unless context =~ %r{^gobierto/}
--- a/app/services/gobierto_common/context_service.rb
+++ b/app/services/gobierto_common/context_service.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module GobiertoCommon
+  class ContextService
+    attr_reader :context
+
+    delegate :present?, to: :resource
+
+    def initialize(context)
+      @context = fill_missing_parts(context.to_s.strip)
+    end
+
+    def resource
+      @resource ||= begin
+                      GlobalID::Locator.locate @context
+                    rescue ActiveRecord::RecordNotFound
+                      nil
+                    end
+    end
+
+    private
+
+    def fill_missing_parts(context)
+      return context if context =~ %r{^gid://gobierto//}
+
+      missing_part = "gid://"
+      missing_part += "gobierto/" unless context =~ %r{^gobierto/}
+
+      "#{missing_part}#{context}"
+    end
+  end
+end
--- a/app/controllers/gobierto_dashboards/api/v1/base_controller.rb
+++ b/app/controllers/gobierto_dashboards/api/v1/base_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module GobiertoDashboards
+  module Api
+    module V1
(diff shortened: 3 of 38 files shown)


app/serializers/gobierto_dashboards/base_serializer.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module GobiertoDashboards
+  class BaseSerializer < ActiveModel::Serializer
+    def current_site
+      Site.find(object.site.id)
```
ferblape: Wouldn't be enought `object.site`?

app/services/gobierto_common/context_service.rb
```diff
@@ -0,0 +1,32 @@
...
+                      nil
+                    end
+    end
+
+    private
+
+    def fill_missing_parts(context)
+      return context if context =~ %r{^gid://gobierto//}
+
+      missing_part = "gid://"
```
ferblape, line 26: You should better use `URI::GID` API, i.e `URI::GID.create('app', model).to_s` https://github.com/rails/globalid/blob/master/lib/global_id/uri/gid.rb

app/serializers/gobierto_dashboards/base_serializer.rb
entantoencuanto: I was very inspired there 🙃

app/services/gobierto_common/context_service.rb
entantoencuanto: In this case the service defines this method to add missing parts to the gid full URI string if required, but the URI::GID class doesn't seem to provide a method to do this, and it doesn't also provide tools to manage a global id as a string it provides a mehtod to create a global id from fragments so I'd have to split a GID string previously.

PR #3588, opened 2020-12-21
Dashboard maker bugs #2

### LINKED ISSUE #3532: Dashboard maker: Mouse selection disabled in cards' textareas
When you're editing an HTML card, the mouse selection is buggy:
- You cannot click wherever you want
- Nor selecting any piece of text

Keyboard works as expected, though.

### LINKED ISSUE #3533: Dashboard maker: drag items from outside
The example provided by the library is quite sloppy and it doesn't work properly for dashboard maker.
Anyway, there're some code commented out in Maker.vue to face the problem.
https://github.com/PopulateTools/gobierto/blob/164c97a63e8c71e6e34ce85f4b79b8ff5c479758/app/javascript/gobierto_TEMP/webapp/Maker.vue#L134-L152

Original source: https://github.com/jbaysolutions/vue-grid-layout/blob/master/website/docs/.vuepress/components/Example10DragFromOutside.vue

Crashillo:
Closes #3532 
Closes #3533

## :v: What does this PR do?
It solves the dashboard-maker bugs
- mouse selection
- drag from outside

#### Caveats
~Drag event behaviour is odd. If you play around without releasing the item, it moves on his own until it's placed.~ Wrapping the `drag` event with the `debounce` helper seems to solve the issue


app/javascript/gobierto_dashboards/webapp/Maker.vue
```diff
@@ -87,91 +87,122 @@ export default {
...
+      this.dirty = false;
     },
     dragoverPosition({ clientX, clientY }) {
-      this.x = clientX
-      this.y = clientY
+      // current mouse position
+      this.x = clientX;
+      this.y = clientY;
+    },
+    debounce(func, timeout) {
```
ferblape, line 129: Not a big deal, but could this be a directive for easly reuse it? It's a common pattern

ferblape (approved):

app/javascript/gobierto_dashboards/webapp/Maker.vue
Crashillo: Yeah, actually I was thinking about it, but I did'nt know where to place it, as it's not a vue-filter, not a mixin... I'm more inclined to [directive](https://es.vuejs.org/v2/guide/custom-directive.html), but look at the common declarations... I've to investigate further

PR #3592, opened 2020-12-22
Migrate stylelint

### LINKED ISSUE #2563: Sass-lint is no longer maintained
As we can see at its website: https://github.com/sasstools/sass-lint, the project is unstable right now.

`stylelint` is a good option, it has Hound & IDEs integration, and better docs. https://stylelint.io/

Crashillo:
Closes https://github.com/PopulateTools/gobierto/issues/2563

## :v: What does this PR do?
Removes `sass-lint` deprecated library, in favour of `stylelint`

- HoundCI config updated
- NPM scripts updated

### Considerations
The following rules have been disabled. Perhaps we should satisfy them in a future.

-   at-rule-no-unknown: null
-   no-descending-specificity: null
-   selector-type-no-unknown: null


.stylelintrc.yml
```diff
@@ -0,0 +1,6 @@
+extends:
```
jorgeatgu, line 1: We can add our own file of stylelint rules, including comments in the rules to give context to the errors.

[Example](https://github.com/jorgeatgu/base/blob/master/.stylelintrc)

jorgeatgu (approved):
We can add our own file of stylelint rules, including comments in the rules to give context to the errors.

[Example](https://github.com/jorgeatgu/base/blob/master/.stylelintrc)

Crashillo:
Hahahaha, funny error messages

PR #3593, opened 2020-12-22
Lint staged: checks pre-commit

Crashillo:
### What does this PR do?
Includes a new git hooks, to check whether the staged files fulfill the rules, before do any commit


config/locales/gobierto_dashboards/views/es.yml
```diff
@@ -13,6 +13,7 @@ es:
 
       Luego podrás configurarla."
     objective: Objetivo
+    ole: 1
```
ferblape: Careful

ferblape (commented):
Will the linter affect to old code?

config/locales/gobierto_dashboards/views/es.yml
Crashillo: It means: **O**bject **l**inted **e**xactly = 1 (true) :trollface:

Crashillo:
No, that's the key. `lint-staged` seeks only in the staged files.
In order to check all files, you simply run: `npm run eslint` or `npm run stylelint`, but.... be pacient

Crashillo:
CircleCI shows multiple failing tests, but apparently they're not related at all 🤷‍♂️

PR #3617, opened 2021-01-12
Use award date instead of start date

ferblape:
## :v: What does this PR do?

This PR updates the contracts visualization to use the award_date field instead of the start_date field

Award date was introduced by @eltercero in https://github.com/PopulateTools/issues/issues/1151 but never got integrated in the visualization

Just check in any of the visualizations that contracts are put in their proper year.


eltercero (approved):
looks good!

jorgeatgu (approved):

Issue #3635, opened 2021-01-19
Data / No LIMIT in default sql for big dataset


furilo:
This https://github.com/PopulateTools/gobierto/issues/3370 is not working OK. In this dataset https://datos.gobierto.es/datos/contratos/resumen which is biiig, if you go to the Data tab, you get the default SELECT * FROM without limit

💥

![image](https://user-images.githubusercontent.com/24740/105094178-2d88ab80-5aa4-11eb-88bb-d68cb27918ac.png)

Why this dataset is not getting a LIMIT? 

Also, looking at this, I can't understand what this means:

![image](https://user-images.githubusercontent.com/24740/105094215-409b7b80-5aa4-11eb-946c-c7d49c9ad175.png)

Please propose other options.

ferblape:
the limit has been fixed in contracts dataset but you need to remove your local storage, because the queries are stored to keep the history

PR #3640, opened 2021-01-20
Return by default 50 rows when size can't be obtained / is nil

ferblape:
Related with #3635

## :v: What does this PR do?

While in that issue you should investigate why a big big dataset doesn't have the size recorded, this PR patches the behaviour of the query editor to add a default limit of 50 rows when the size is nil.

In staging

- update the size of a dataset to nil
- check the default query contains a LIMIT 50


ferblape:
The failing test is a random flaky test, I'm merging to use it in production

PR #3657, opened 2021-01-26
Allow gobierto-data to be startpage

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1155

## :v: What does this PR do?
Run gobierto-data frontend when path is also the root folder

https://demo-datos.gobify.net/


ferblape:
@furilo is this duplicated content? / and /datos respond with the same info. Maybe if it's the home module redirect /datos to / ?

jorgeatgu (approved):

PR #3663, opened 2021-01-28
Sql permalink

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1150

## :v: What does this PR do?
- When the user runs a query, immediately the url is updated with such query (encoded)
- When the url contains `sql=` parameter, it's parsed into the editor

Test: https://getafe.gobify.net


ferblape (changes requested):
Hmmmm, something is broken, if you load a query URL such as https://getafe.gobify.net/datos/calidad-del-aire/q/160, the SQL query is not loaded in the editor and nothing happens:

Screenshot 2021-02-01 at 07 02 19

Also, if you go to a dataset, play with queries and then click on a saved query it breaks with a JS error

Crashillo:
Errors fixed and deployed

furilo:
It seems OK.

app/javascript/gobierto_data/modules/gobierto_data_controller.js
```diff
@@ -3,6 +3,7 @@ import { router } from "../lib/router";
 import 'codemirror/lib/codemirror.css'
 
 Vue.config.productionTip = false;
+Vue.config.devtools = true;
```
ferblape, line 6: Is it dangerous for production? (less performance?)

app/javascript/gobierto_data/webapp/components/sets/DownloadsTab.vue
```diff
@@ -36,14 +36,7 @@ export default {
...
     return {
-      directionLeft: true,
-      filesDataset: [{
-          file: {
-            name: 'Actuaciones Bomberos Enero 2019.xls',
-            size: '232kb'
-          }
-        }
-      ]
+      directionLeft: true
```
ferblape, line 39: 🙀

ferblape (approved):

app/javascript/gobierto_data/modules/gobierto_data_controller.js
Crashillo: Currently is enabled in some other modules. But yes, it's not a good practice, the ideal is about to have something like this: https://webpack.js.org/plugins/environment-plugin/, then just replace that `true` with `process.env.DEBUG`

Crashillo: Here we are https://github.com/PopulateTools/gobierto/issues/3680

PR #3671, opened 2021-02-01
Hide dashboards if disabled

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1189

## :v: What does this PR do?
Hide dashboards in plans by default (navbar)

### DIFF
--- a/app/controllers/concerns/gobierto_common/module_helper.rb
+++ b/app/controllers/concerns/gobierto_common/module_helper.rb
@@ -4,8 +4,13 @@ module GobiertoCommon
 
     private
 
+    def module_enabled?(module_namespace, site = nil)
+      site ||= current_site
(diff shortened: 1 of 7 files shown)


Crashillo:
We need to update the directive below, based on two conditions:
1. dashboards-module enabled
2. dashboards > 0

https://github.com/PopulateTools/gobierto/blob/59dcfb1089dba3ada6acd78906f8e758cfb9a3ac/app/views/gobierto_plans/plan_types/show.html.erb#L42

cc/ @entantoencuanto

entantoencuanto:
Done, I've added a helper method which returns true only if module is enabled and there are any public dashboard for the plan.

jorgeatgu (approved):

PR #3675, opened 2021-02-02
Performance contracts

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1188

## :v: What does this PR do?
Improve performance of contracts website

- filter minor contracts
- reduce the ticks to the 1/4 of the previous value

Test: https://alcobendas.gobify.net/visualizaciones/contratos/


Crashillo:
@ferblape this hack https://github.com/PopulateTools/gobierto/pull/3675/commits/392f5ac4da0f8bcbd9de0c9b6823978567aaf115 is due to, otherwise the API blocks the requests in localhost when user is not logged. Is there any way to avoid this?

app/javascript/gobierto_visualizations/webapp/lib/utils.js
```diff
@@ -5,7 +5,7 @@ import axios from "axios";
 const endPointGobiertoData = `/api/v1/data/data.json`
 
 export function getRemoteData(endpoint) {
-  return csv(endpoint, { headers: { Authorization: "Bearer FmC7jKbXJbzLn1UnmWMXbNXe" } });
+  return csv(endpoint);
```
ferblape: :(

ferblape: I've just expired this token

ferblape:
> @ferblape this hack [392f5ac](https://github.com/PopulateTools/gobierto/commit/392f5ac4da0f8bcbd9de0c9b6823978567aaf115) is due to, otherwise the API blocks the requests in localhost when user is not logged. Is there any way to avoid this?

Yes, in the configuration of the visualization add the token to the urls as a get parameter

ferblape:
This is ok for the moment and it can be merged.

Should we notice in a message that minor contracts are excluded from the visualizations @furilo ?

PR #3681, opened 2021-02-03
Gobierto Visualizations / Bug on resize and don't paint some elements in the DOM

jorgeatgu:
## :v: What does this PR do?

Resize:
When resizing the browser, we lose the value(final_amount) that gives the circles' size. Its happens because transforming the treemap data, the value was removed, and the main array was updated without this value.

DOM:
We do not build the elements that are not seen. If the text doesn't fit in the rectangle, we don't show it, so there's no point in building something that won't be seen.

Staging

### Before this PR
Screen Recording 2021-02-03 at 12 53 56 2021-02-03 12_59_18

### After this PR

Screen Recording 2021-02-03 at 13 13 26 2021-02-03 13_14_14

### DIFF
--- a/app/javascript/gobierto_visualizations/webapp/containers/contracts/CategoriesTreeMapNested.vue
+++ b/app/javascript/gobierto_visualizations/webapp/containers/contracts/CategoriesTreeMapNested.vue
@@ -88,7 +88,6 @@ export default {
           }
           if (key === value_key) {
             root.value = parseFloat(root[value_key]);
-            delete root[value_key];
           }
         }
         return root;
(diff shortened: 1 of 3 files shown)


ferblape (approved):

Crashillo (commented):
Code now is OK, but you're not filtering the elements in any way, aren't you?

ferblape:
Why needs to filter elements?

Crashillo:
Because the PR title says "don't paint some elements in the DOM", but it is painting all of them indeed 😄 

That's about the perfomance we saw, it's nonsense adding elements to the DOM if they cannot be seen, that's why they should be filtered by the first N elements.

PR #3706, opened 2021-02-12
Gobierto Visualizations / Remove nodes from treemap's

### LINKED ISSUE #3686: Remove DOM nodes
Follows #3681

Pending to implement: 

> DOM:
> We do not build the elements that are not seen. If the text doesn't fit in the rectangle, we don't show it, so there's no point in building something that won't be seen.

jorgeatgu:
Closes PopulateTools/issues#1205
Closes #3686

## :v: What does this PR do?

- To improve the treemap's performance, we are only going to build the HTML of those elements that are going to be visible. When a level has many elements, we will not make it directly because these elements will not be visible, and to build the HTML is to add 4 HTML nodes per element.

- DRY for the functions that build the HTML of the different depths.
- Reducing the height of the treemaps

Staging


furilo:
Has this broken https://esplugues.gobify.net/visualizaciones/subvenciones ?

imagen

Crashillo (approved):

PR #3715, opened 2021-02-17
Number of columns of inversions summary table 

Crashillo:
## :v: What does this PR do?
Allow that another category fits on the screen

imagen


ferblape (approved):
OK to me!

@entantoencuanto all yours, you decide if it can be deployed now or has to wait

PR #3716, opened 2021-02-17
Gobierto Contracts / Show permalink on hover

jorgeatgu:
Closes PopulateTools/issues#1207

## :v: What does this PR do?

Show the permalink on hover if the table row contains a link. 

If the table contains a link, the rows will be wrapped in a tag `<a>`, and if not, they will be wrapped in a `<tr>`.

Staging

### After this PR

Screen Recording 2021-02-17 at 11 38 42 2021-02-17 11_40_44

### DIFF
--- a/app/assets/stylesheets/comp-vue-table.scss
+++ b/app/assets/stylesheets/comp-vue-table.scss
@@ -130,12 +130,32 @@
 
   .is-clickable {
     cursor: pointer;
-    display: table-row;
-    vertical-align: middle;
     text-decoration: none;
     color: $color_text;
     &:hover {
       background-color: rgba(var(--color-base-string), .1);
+      .gobierto-table__a {
+        color: $color_text;
+      }
+    }
+    .gobierto-table__td {
+      height: auto;
+      padding: 0;
+      min-height: 30px;
+    }
+    .gobierto-table__a {
+      text-decoration: none;
+      display: block;
+      height: 100%;
+    }
+    .ellipsis {
+      .gobierto-table__a {
+        width: 430px;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        display: inline-block;
+      }
     }
   }
 }
--- a/app/assets/stylesheets/comp-vue-table.scss
+++ b/app/assets/stylesheets/comp-vue-table.scss
@@ -130,6 +130,10 @@
 
   .is-clickable {
     cursor: pointer;
+    display: table-row;
+    vertical-align: middle;
+    text-decoration: none;
+    color: $color_text;
     &:hover {
       background-color: rgba(var(--color-base-string), .1);
     }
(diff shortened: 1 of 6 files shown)


furilo (approved):

Crashillo:
It's still pending some changes we made y'day

app/javascript/lib/vue/components/modules/Table.vue
```diff
@@ -148,9 +194,17 @@ export default {
...
+      type: String,
+      default: ''
+    },
+    routingName: {
+      type: String,
+      default: ''
+    },
+    routingComponent: {
+      type: String,
+      default: ''
```
Crashillo: No longer necessary

app/javascript/lib/vue/components/modules/Table.vue
```diff
@@ -82,7 +115,20 @@
                   :class="cssClass"
                   class="gobierto-table__td"
                 >
-                  <span>{{ item[id] }}</span>
+                  <template v-if="hasPermalink">
+                    <a
+                      :href="item.href"
```
Crashillo: Instead of enforcing an existing `href` in the data, this could be configurated as a prop:
```
    href: {
      type: String,
      default: 'href'
    },
```
And you can use it simply as `item[this.href]`, if parent pass the property, you'll use such value, otherwise it looks for `href` in the data object

app/javascript/lib/vue/components/modules/Table.vue
```diff
@@ -226,6 +280,9 @@ export default {
     filterColumns(columns) {
       this.mapColumns = columns
       this.arrayColumnsFiltered = Array.from(this.mapColumns).filter(([,{ visibility }]) => !!visibility)
+    },
+    handleTableItem(item) {
+      this.$emit('send-item', item)
```
Crashillo: why `send-item`? where to? :D on-href-click or similar

Crashillo (changes requested):

app/javascript/lib/vue/components/modules/Table.vue
```diff
@@ -164,6 +210,9 @@ export default {
     };
   },
   computed: {
+    hasPermalink() {
+      return this.data.some(element => element['href'])
```
Crashillo:
```suggestion
      return this.data.some(element => element[this.href])
```

app/assets/stylesheets/comp-vue-table.scss
```diff
@@ -130,8 +130,33 @@
...
+      height: 100%;
+    }
+    .ellipsis {
+      .gobierto-table__a {
+        width: 430px;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        display: inline-block;
+      }
```
Crashillo: Don't nest

Crashillo (changes requested):

Crashillo (approved):

PR #3717, opened 2021-02-17
Update default investments filters

entantoencuanto:
## :v: What does this PR do?

Changes some filters in invesments index page:
* Removes nom-servei-gestor
* Adds any-partida
* Adds tipus (category)

Frontend changes resquested by client also applied:
- No element title in the cards (gallery & map). Less fields displayed.
- Table view uses the gobierto table vue-component (enables pagination & column sorting)
- Refactorized project view

No

No


jorgeatgu (approved):

Crashillo:
Still broken, same point, in a non-related test: https://app.circleci.com/pipelines/github/PopulateTools/gobierto/4097/workflows/bdf446e3-33a9-423d-922d-1ebf4248d801/jobs/33984

ferblape:
@Crashillo in this PR you have modified components used also by the visualizations, such as the table. Are you sure these changes haven't break the table and therefore the visualization ? 🤔

Crashillo:
I've just tried to run tests in local, I got errors but they look different:
```
Error:
GobiertoVisualizations::VisualizationsContractsTest#test_contracts:
Net::ReadTimeout: Net::ReadTimeout
```
I do not know what I'm missing. In any case, CircleCI shows the following error:
```
Expected false to be truthy.
test/integration/gobierto_visualizations/visualizations_contracts_test.rb:122:in `block in test_contracts'
```
These lines points to:
https://github.com/PopulateTools/gobierto/blob/master/test/integration/gobierto_visualizations/visualizations_contracts_test.rb#L119-L122

They're about if a table cell contains some text, what they have 😕

Crashillo:
Aaaah, OK, I think I get it. The test points to: `all("tr.gobierto-table__tr")[4]` what I guess it means the 4th row, isn't it? But in the contracts.csv, `SANIVIDA, S.L.` is not at line 4/5, but much downward. Perhaps is this related?

ferblape:
Fifth because start couting in 0. Maybe tests have changed order? Ask @eltercero for help, he's the man behind those tests

eltercero:
yes, it's because the elements are in an array. 

If the order of the data has changed the test has to be updated, but it looks like you haven't changed anything from visualizations, right? Maybe the merges from master went wrong?

Crashillo:
Now the table we're using there, comes sorted by a column (in contracts, it's import), so I would say the test should consider the table order before accessing any row

eltercero:
> Now the table we're using there, comes sorted by a column (in contracts, it's import), so I would say the test should consider the table order before accessing any row

In the tests we have mock data (there are some csvs files in the repository). If the data order has changed, the test has to be updated.

Crashillo:
It was the table column order, it was ascendent (which is the default), so it requires to enforce the sort direction if you wanna display the items from the highest value down

Crashillo:
Wait! Is this the first time in a loooooooong time the tests are worth? This is the first time the tests actually worth!

Crashillo:
Yippee ki-yay, mother****er!

jorgeatgu (approved):

PR #3719, opened 2021-02-18
3653 cache data queries

### LINKED ISSUE #3653: Gobierto Data / Cache query requests
Query requests (`/queries/<id>`) should be cached as the endpoint `/api/v1//data/data`

entantoencuanto:
Closes https://github.com/PopulateTools/gobierto/issues/3653

## :v: What does this PR do?

* Changes queries controller to cache data in csv format for show and download actions
* Removes a method which is no longer required with the previous changes

Currently the performance improvement can't be seen in the front because the app doesn't make use of the queries show or download endpoints, but it can be observed calling directly to the API:

* `GET /api/v1/data/queries/ID/download.csv` 
* `GET /api/v1/data/queries/ID`

No

No

### DIFF
--- a/app/controllers/gobierto_data/api/v1/base_controller.rb
+++ b/app/controllers/gobierto_data/api/v1/base_controller.rb
@@ -26,17 +26,6 @@ module GobiertoData
           end
         end
 
-        def csv_from_query_result(result, options = {})
-          return if result.blank?
-
-          CSV.generate(**options) do |csv|
-            csv << result.fields
-            result.each_row do |row|
-              csv << row
-            end
-          end.force_encoding("utf-8")
-        end
-
         def csv_from_relation(relation, options = {})
           with_serialized_data_from_relation(relation) do |data, new|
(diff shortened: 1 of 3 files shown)


ferblape:
Sorry I wasn't explicit enough in the issue, but the endpoint /api/v1/data caches both JSON and CSV formats, and in this PR you are not caching the JSON response of the queries endpoint, and you should.

> Currently the performance improvement can't be seen in the front because the app doesn't make use of the queries show or download endpoints, but it can be observed calling directly to the API:

You are wrong here, if you go to visualizations page of a dataset, the visualizations use queries. There is where we noticed the performance degradation. See for example the XHR requests in https://getafe.gobify.net/datos/licitaciones/visualizaciones to the queries endpoint (i.e https://getafe.gobify.net/api/v1/data/queries/168)

That's request we were aiming to improve in this issue.

ferblape (approved):
Much fastter now

PR #3720, opened 2021-02-18
Split test by modules

Crashillo:
## :v: What does this PR do?
Split circleci logic into different tests by modules

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,11 @@
 ---
+version: 2.1
+
+aliases:
+  - &working_directory ~/gobierto
+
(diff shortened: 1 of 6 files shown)


ferblape:
@Crashillo @entantoencuanto review please, this is ready to me

Crashillo:
Wow, only 4' to run admin tests 😯 wunderbar!

Crashillo:
Three last times`tests_engines` are failing in the same tests. Perhaps r they actually broken?

Crashillo (commented):
✅ I cannot aprove it 'cause I'm the creator. So... check!

PR #3721, opened 2021-02-18
Reduce requests in Gobierto Data

### LINKED ISSUE #3658: Gobierto Data / Reduce the number of requests
When we're requesting the queries or visualizations, currently two XHR are being done:
imagen

The first one, is returning the public queries, and the latter the private ones. Since the app manages user tokens that approach is quite inefficient. Therefore:

1. API must check the `Authorization` token of the request. Cases:
    1.1 If none, it just returns the public items
    1.2 If token, it means the user is logged, return both public and private (their own) items
2. Front must refactor all the public/private stuff. In terms of XHR requests, it won't distinguish between public or private, it only make ONE call, once it fetches the results, it will seek the `relationships/user` properties to know which results are privates.

entantoencuanto:
Closes #3658

## :v: What does this PR do?

- [x] Changes filtered relation of visualizations and queries controllers to always show the closed items owned by logged in users compatible with the other filters.
- [x] Add some tests to visualizations controller to check this
- [x] Refactorized the number of API calls all around the module (Both queries & visualizations)
- [x] Removed multiple functions no longer needed
- [x] Visualizations pointing to same query, reduced (refactorized) the number of XHR calls, to make it unique
- [x] BUG: when update an already saved query with some other content, saved query comes again (vue $watch)


ferblape:
@Crashillo  All good except a little detail with a sql parameter added to the URL that shouldn't:

1. Go to https://getafe.gobify.net/datos/v/visualizaciones
2. Click in any visualization, i.e Monóxido de nitrógeno..
3. The url contains a SQL parameter that shouldn't https://getafe.gobify.net/datos/calidad-del-aire/v/98?sql=SELECT%2520%2a%2520FROM%2520calidad_del_aire%2520LIMIT%252050 it should be just https://getafe.gobify.net/datos/calidad-del-aire/v/98

jorgeatgu (approved):

Crashillo:
> @Crashillo All good except a little detail with a sql parameter added to the URL that shouldn't:
> 
>     1. Go to https://getafe.gobify.net/datos/v/visualizaciones
> 
>     2. Click in any visualization, i.e Monóxido de nitrógeno..
> 
>     3. The url contains a SQL parameter that shouldn't https://getafe.gobify.net/datos/calidad-del-aire/v/98?sql=SELECT%2520%2a%2520FROM%2520calidad_del_aire%2520LIMIT%252050 it should be just https://getafe.gobify.net/datos/calidad-del-aire/v/98

Fixed

ferblape (approved):
Thanks, now is fixed

furilo:
This keeps generating more queries than it should? https://datos.gobierto.es/datos/contratos-exploracion/v/3

Crashillo:
The problems comes up because the fourth tab shows both all vizzs and a particular one.
https://github.com/PopulateTools/gobierto/blob/90078660cfc83fb2af960c9fdb9e88b03f8f381f/app/javascript/gobierto_data/webapp/components/Dataset.vue#L504-L512

But solution is kinda hacky. I'd need to look up further.

PR #3730, opened 2021-02-23
Add unit type options to numeric custom field form

entantoencuanto:
## :v: What does this PR do?

* Adds options to numeric custom field type to set the unit type. For the moment only `generic` (default) and `currency`
* Note:  The i18n-task hook has automatically added changes in config/locales/defaults/ translations removing some trailing spaces

From admin edit/create a custom field of numeric type. A unit type section should appear

### Before this PR

Screen Shot 2021-02-23 at 22 47 47

### After this PR

Screen Shot 2021-02-23 at 22 48 28

No

* [Suggestion added](https://dash.readme.com/project/gobierto/v0.1/suggested/update/60357a99eedb840038f074e8)

### DIFF
--- a/app/controllers/gobierto_admin/gobierto_common/custom_fields/custom_fields_controller.rb
+++ b/app/controllers/gobierto_admin/gobierto_common/custom_fields/custom_fields_controller.rb
@@ -133,6 +133,7 @@ module GobiertoAdmin
             :vocabulary_type,
             :plugin_type,
             :date_type,
+            :unit_type,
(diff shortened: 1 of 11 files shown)


furilo:
Do we have a default currency? Would we able to change it? I only anticipate needing €'s, but to understand it.

entantoencuanto:
For the moment we are going to assume that the default currency is €, and in the future, if other currency is required, we will add an extra param and configuration interface to admin

PR #3745, opened 2021-03-03
Gobierto Visualizations / Translate categories

jorgeatgu:
Closes #N

## :v: What does this PR do?

- Translate the categories titles.
- Change the color of the treemap.

[Staging](https://santfeliu.gobify.net/visualizaciones/contratos?locale=ca)

### After this PR

translate


app/javascript/gobierto_visualizations/webapp/lib/config/category_titles.js
```diff
@@ -0,0 +1,74 @@
+export const categoryTitleEng = [
```
eltercero: these are not exactly the titles in English, these are the translation keys. I'm saying it because maybe in the future there are real English translations and they object should have this name (just like `categoryTitlesCat` and `categoryTitlesEsp`).

ferblape (changes requested):
Why aren't we using Rails I18n files here? You are re-inventing the wheel

app/javascript/gobierto_visualizations/modules/contracts_controller.js
```diff
@@ -207,6 +208,25 @@ export class ContractsController {
...
   }
 
+  _translateCategoriesTitle(contractsDataMap) {
+
+    const categoriesLanguage = I18n.locale === 'es' ? categoryTitlesEsp : categoryTitlesCat
+
+    contractsDataMap.map(d => {
+      const { category_title } = d
+      /*If the contract isn't categorized, we include it in other*/
+      if (!category_title) {
```
eltercero: We don't need this in the end, since they will never come as null (and if they do, we should know to fix it)

eltercero (changes requested):

app/javascript/gobierto_visualizations/modules/contracts_controller.js
jorgeatgu: Ok. Done!

jorgeatgu:
> Why aren't we using Rails I18n files here? You are re-inventing the wheel

Ouch. Fixed.

app/javascript/gobierto_visualizations/webapp/visualizations/treeMapNested.vue
```diff
@@ -309,7 +309,7 @@ export default {
             } else if (deepLevel === 3 && scaleColor && depth === 2) {
               return this.colors(contracts[this.firstDepthForTreeMap])
             } else {
-              return '#12365b'
+              return '#118e9c'
```
furilo: Should this be dynamic to avoid what just happened?

app/javascript/lib/shared/modules/createScaleColors.js
```diff
@@ -1,7 +1,7 @@
 import { scaleOrdinal } from 'd3-scale';
 
 export function createScaleColors(values, arrayDomain) {
-  let colorsGobiertoExtend = ["#12365b", "#118e9c", "#ff766c", "#f7b200", "#158a2c", "#94d2cf", "#3a78c3", "#15dec5", "#6a7f2f", "#55f17b"]
+  let colorsGobiertoExtend = ["#118e9c", "#12365b", "#ff766c", "#f7b200", "#158a2c", "#94d2cf", "#3a78c3", "#15dec5", "#6a7f2f", "#55f17b"]
```
furilo: Shouldn't we be using here CSS variables of the brand colors? I guess this are already defined.

app/javascript/gobierto_visualizations/webapp/visualizations/treeMapNested.vue
jorgeatgu: Done.

app/javascript/lib/shared/modules/createScaleColors.js
jorgeatgu: Yes. I've added the Gobierto colors to the `css-confs` as CSS vars.

ferblape (approved):
Much better!

PR #3771, opened 2021-03-17
Gobierto Visualilzations / Check if contracts comes with gobierto_start_date

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1237

## :v: What does this PR do?

In some contracts, `gobierto_start_date` is `undefined`. We add an exception, if `gobierto_start_date` is `undefined`, it will be `formalized_date`. In case `gobierto_start_date` and `formalized_date` come `undefined`, we use `award_date`.

[Staging](https://alcobendas.gobify.net/visualizaciones/contratos/adjudicaciones)

### After this PR

Screenshot 2021-03-17 at 11 37 43

### DIFF
--- a/app/javascript/gobierto_visualizations/modules/contracts_controller.js
+++ b/app/javascript/gobierto_visualizations/modules/contracts_controller.js
@@ -174,7 +174,15 @@ export class ContractsController {
       .domain(this._amountRange.domain)
       .range(this._amountRange.range);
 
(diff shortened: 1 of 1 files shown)


ferblape:
As I said in https://github.com/PopulateTools/issues/issues/1237#issuecomment-800991795 it was a data issue, so we are not solving it in the visualization. Closing.... The commit should be reverted in staging

jorgeatgu:
> The commit should be reverted in staging

@ferblape changes reverted

PR #3782, opened 2021-03-24
Gobierto Contracts / review accesibility

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1200

### Level A and Level AA

**img elements must have an accessible name.**

Some illustrations don't have an alt attribude.

**The element main must not appear as a descendant of the main element.**

In contracts we use two tag main.

**Clickable controls should be keyboard accessible. Clickable controls should have an ARIA role.**

When we need a button we should use tag button instead tag a or whatever.

**Do not use the meta viewport tag to disable zoom**

Setting user-scalable=no or maximum-scale to less than 2.0 prevents zooming small text to make it readable.

### After this PR


ferblape:
@jorgeatgu could you rebase master?

jorgeatgu:
👉🏻  https://github.com/PopulateTools/gobierto/pull/3895

PR #3785, opened 2021-03-25
Update terser version

ferblape:
## :v: What does this PR do?

This PR updates terser version to try to fix problems in CircleCI

Everything in staging should work as expeced

CircleCI should finish the build


Crashillo:
In webpack v5, this dependency won't be longer necessary, will it?

Crashillo (approved):

PR #3517, opened 2020-11-27
Performance plugins

Crashillo:
### What does this PR do?
Includes two webpack plugins to measure performance: `BundleAnalyzerPlugin` ~and `SpeedMeasurePlugin`~. Both are enabled only for development environment.

_BundleAnalyzerPlugin_ was already there, but disabled. Now, it's always enabled at  _http://localhost:8888/_ once you run the application.

On the other side, `terser` plugin, only useful to circleci developments (it decreases a lot the hot-module-reloading) has been activated only for _test_ environment.

Besides, _webpacker_ configuration (`config/webpack`) has been simplified, separating some webpack modifications to its own files. It also allows to add libraries to different environments.

Development assets output has been reduced, through [stats webpack config](https://v4.webpack.js.org/configuration/stats/), set to `errors-warnings`.

PR #3788, opened 2021-03-26
No terser

Crashillo:
## :v: What does this PR do?
Remove `terser` dependency and sequence the minification process in circleci (less memory leaks)

### DIFF
--- a/config/webpack/optimization/terser.js
+++ b/config/webpack/optimization/terser.js
@@ -4,7 +4,7 @@ module.exports = {
   optimization: {
     minimizer: [
       new TerserPlugin({
-        parallel: 2
+        parallel: false
       })
     ]
   }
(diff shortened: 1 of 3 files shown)


ferblape:
@Crashillo I've an idea, why we don't remove the whole Terser thing in test environment?

Crashillo:
Initially, the idea was about using the terser only in the test environment. Indeed, in the performance-PR https://github.com/PopulateTools/gobierto/pull/3517/files, the terser optimization only applies to test; but the current configuration doesn't tell enviroments apart, they're all the same.

According the [plugin docs](https://webpack.js.org/plugins/terser-webpack-plugin/), no configuration would set `parallel: true` (default), which was the original circleci issue (it requires a number)

PR #3795, opened 2021-03-30
Only enable extensions if not enabled

ferblape:
Unexpected

## :v: What does this PR do?

Only enable database extensions if haven't been enabled


eltercero (approved):
I guess I'll do the same for contratos

ferblape:
Gut

PR #3798, opened 2021-03-31
Update case-sensitive-paths-webpack-plugin version (performance)

ferblape:
## :v: What does this PR do?

This PR updates the plugin case-sensitive-paths version because of this release has some improvements in performance


Crashillo:
Humm, look at this commit I pushed y'day, disabling the plugin in development environment
https://github.com/PopulateTools/gobierto/pull/3517/commits/18b7ad320ea3859d3315f216a0e0d4e3aa33f5f0

ferblape:
Still failing... With this upgrade, the analytics in my laptop says that the compile time is 30 secs again and not 1 min

BTW I also manually updated the plugin first in `webpack-analytics` branch, so the commit is a bit duplicated 😅

Crashillo (approved):

PR #3804, opened 2021-03-31
Remove needless gems

### LINKED ISSUE #3515: Evaluate Gemfile front-end dependencies
```
# Frontend
gem "bourbon", "~> 7.0.0"
gem "sass", "~> 3.4"
gem "flight-for-rails"
gem "i18n-js", ">= 3.0.0.rc11"
gem "jquery-rails"
gem "sassc"
gem "therubyracer"
gem "turbolinks"
gem "uglifier", ">= 1.3.0"
gem "chroma"
gem "font-awesome-sass", "~> 5.6"

# Webpack
gem "webpacker", "~> 5.0"
```

Due to legacy code, Gemfile defines a bunch of front-end dependencies which actual use is quite suspicious.
For instance, webpacker gem requests for 5.x version, but we're using webpack 4.x.

Check out, one by one, if they're could be safely removed.

The final purpose is separation of concepts: `package.json` -> front-end / `Gemfile` -> back-end

Crashillo:
Closes https://github.com/PopulateTools/gobierto/issues/3515

## :v: What does this PR do?
Clean Gemfile from frontend libraries

- [x] bourbon
- [x] sass
- [x] flight-for-rails
- [ ] ~18n-js~ required for rails commands
- [x] jquery-rails
- [x] sassc
- [x] therubyracer
- [x] turbolinks
- [x] uglifier
- [x] chroma
- [ ] ~font-awesome-sass~ it should be moved to npm


Gemfile
```diff
@@ -26,17 +26,8 @@ gem "truncate_html"
...
-gem "bourbon", "~> 7.0.0"
-gem "sass", "~> 3.4"
-gem "flight-for-rails"
-gem "i18n-js", ">= 3.0.0.rc11"
-gem "jquery-rails"
-gem "sassc"
-gem "therubyracer"
-gem "turbolinks"
-gem "uglifier", ">= 1.3.0"
-gem "chroma"
```
ferblape, line 38: I think chroma is needed (see the error in citizien charts module)

The rest of tems might not be needed as they are now depedencies in npm (flight, jquery, turbolinks...)

Crashillo: I dug into Chroma gem and I've found it only in this file:
https://github.com/PopulateTools/gobierto/blob/4de026646f6d55153c4ba10a7b2fa82bb7e34543/app/views/gobierto_citizens_charters/charters/_single_charter_stylesheets_hook.html.erb#L9-L18

Looking at what it does, it seems only applies some opacity to an hex-color inside a CSS snippet, does it worth it keeping the gem? Perhaps it'd be easy replace it with no deps
cc/ @entantoencuanto

furilo:
Please replace it, or even reduce the functionality, it's probable that we
delete this module.

Gemfile
Crashillo: code updated to remove this gem, so perhaps some test fail if they're testing Chroma somehow

ferblape:
@Crashillo did you release this to staging?

Crashillo:
I did 04c2a228e9283c9cb4fd19331cd8d78e3e018957

PR #3810, opened 2021-04-07
Refactor DC, fix filters & update data queries

Crashillo:
## :v: What does this PR do?
- [x] Refactorization of old code, it seems `dc` has migrated to `v4.x` breaking some files.
- [x] Fix the UX interaction: filtering
- [ ] ~Update the data queries based on the new datasets~

Test: https://getafe.gobify.net/observatorio/mapa

### DIFF
--- a/app/assets/stylesheets/module-observatory.scss
+++ b/app/assets/stylesheets/module-observatory.scss
@@ -339,7 +339,7 @@
     padding: 1rem 2rem;
   }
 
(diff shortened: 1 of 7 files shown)


Crashillo:
Since the new datasets breaks the current visualization, I'd merge this PR (I removed the references to the issues) to avoid keep this open over the time we can adapt the datasets.

PR #3816, opened 2021-04-09
Gobierto Visualizations / Replaces the path to avoid redirect to blank page.

jorgeatgu:
## :v: What does this PR do?

If the URL doesn't contain a slash at the end and the user clicks on the beeswarm circle, it redirects to a blank page. Replaces the path to avoid redirect to blank page.
Don't get the HREF data from parent to deal with undefined.

### Before this PR

before

### After this PR

after

### DIFF
--- a/app/javascript/gobierto_visualizations/webapp/containers/contracts/Summary.vue
+++ b/app/javascript/gobierto_visualizations/webapp/containers/contracts/Summary.vue
@@ -8,7 +8,7 @@
     <EntityTreeMapNested
       v-if="activeTab === 0"
       id="gobierto-visualizations-treemap-entity"
-      :data="visualizationsData"
+      :data="visualizationsDataEntity"
       class="mt4"
     />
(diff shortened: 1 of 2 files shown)


app/javascript/gobierto_visualizations/webapp/containers/contracts/Summary.vue
```diff
@@ -269,7 +273,7 @@ export default {
     goesToItem(event) {
       const { id } = event
       // eslint-disable-next-line no-unused-vars
-      this.$router.push(`adjudicaciones/${id}`).catch(err => {})
+      this.$router.push(`/visualizaciones/contratos/adjudicaciones/${id}`).catch(err => {})
```
Crashillo, line 276: Try always to parametrize route paths as it's done in some other modules- This literal string should be only present at the router configuration.

You can create a new issue for refactor all this, since there's no route parametrized yet

Crashillo (approved):

PR #3817, opened 2021-04-13
Gobierto Data / Create a plugin map for Perspective

### LINKED ISSUE #3800: Data / Implement geospatial view
We should provide a simple map view of a dataset if it contains geospatial columns. Perspective doesn't provide a native map view, so we may implement something based on Leaflet.

The idea is that if a table contains some column/s with an specific format, we generate the map visualization. 

- define the format of the columns we accept (lat+lon, geojson, points, polygons...?). Start with something simple
- if the front detect this column/s, it shows the map. Decide where we show it. As a new tab? A block in the dataset home?

jorgeatgu:
Closes #3800 

## :v: What does this PR do?

Creates a perspective plugin from scratch to be able to create maps when datasets contain geospatial data.

[Staging](https://getafe.gobify.net/datos/secciones-censales)

### After this PR

[Loom](https://www.loom.com/share/0373f10b57b84eb9b9e8ec27dda4c4d1)


app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -10,6 +10,10 @@ import "@finos/perspective-viewer";
 import "@finos/perspective-viewer-datagrid";
 import "@finos/perspective-viewer-d3fc";
 import "@finos/perspective-viewer/themes/all-themes.css";
+/* eslint-disable */
+import { MapPlugin } from "../../../../lib/perspective-viewer-map/plugin.js";
```
furilo: We'll vendorize this later?

jorgeatgu: Yes!

furilo:
Looks good ! Some comments: https://www.loom.com/share/442f5ff6390a4eaebcd669e80fb70f7e cc @ferblape

ferblape:
- +1 to load more datasets with geo data (points also)

- I'd merge this PR yet at least until there's an easy way to select the variable to be displayed in the map, because otherwise it's not very useful

- I wouldn't mess with joins (although with a join it works fine https://getafe.gobify.net/datos/habitantes-distritos-secciones/editor?sql=SELECT%2520h.varon%252C%2520h.mujer%252C%2520s.geometry%250AFROM%2520habitantes_por_distritos_secciones%2520h%250AINNER%2520JOIN%2520secciones_censales%2520s%2520ON%2520h.seccion%253A%253Ainteger%2520%253D%2520s.csec) but I'd assume that the table has to have a geometry column, so we should add it in the ETLs when loading the data

- the choropleth algorithm can be adjusted a bit. See the example above, I can't distinguish between blue levels

- to do in another issue, categorical variables, not just numeric

Crashillo:
> the choropleth algorithm can be adjusted a bit. See the example above, I can't distinguish between blue levels

Something is happen there. The maximum category color has, at least, some [transparency](https://getafe.gobify.net/datos/secciones-censales/editor?sql=SELECT%2520%2a%2520FROM%2520secciones_censales). Ranges begin at 0.1 to 0.9, and the opaque color in the [previous query](https://getafe.gobify.net/datos/habitantes-distritos-secciones/editor?sql=SELECT%2520h.varon%252C%2520h.mujer%252C%2520s.geometry%250AFROM%2520habitantes_por_distritos_secciones%2520h%250AINNER%2520JOIN%2520secciones_censales%2520s%2520ON%2520h.seccion%253A%253Ainteger%2520%253D%2520s.csec) actually has the 3rd category, which implies a 0.5 transparency

Crashillo:
Indeed, the query fetches repeated results that overlap each other. 

The [fixed query](https://getafe.gobify.net/datos/habitantes-distritos-secciones/editor?sql=SELECT%2520h.varon%252C%2520h.mujer%252C%2520s.geometry%250AFROM%2520habitantes_por_distritos_secciones%2520h%250AINNER%2520JOIN%2520secciones_censales%2520s%2520ON%2520h.seccion%253A%253Ainteger%2520%253D%2520s.csec%250Aand%2520h.distrito%253A%253Ainteger%2520%253D%2520s.cdis%250Aand%2520fecha%253D%271%252F4%252F2021%27) filters also by _district_ and _fecha_

Perhaps we could compare before rendering if the GeoJSON layer was previously added.

imagen

Crashillo:
> Perhaps we could compare before rendering if the GeoJSON layer was previously added.

Ñeeee, thinking this further.. I don't like it. Each GeoJSON contains specific _properties_ referring to its geometry. This geometry could be doubled, but their props couldn't. Therefore, IMO you shouldn't filter stuff just because a shape is repeated, their values ARE different.

Crashillo:
> I'd merge this PR yet at least until there's an easy way to select the variable to be displayed in the map, because otherwise it's not very useful

What do you mean by _easy way_? Selecting the first numeric value is being done also by the other perspective plugins. But regarding this, if we gonna depict categorical variables, we won't need anymore to find the first numeric column, but we'll face a similar issue instead: which column? the first one? I don't think this matter must be solved by the plugin beyond distinguish the variable types: _numeric -> continuous vs. text -> discrete_

The main difference between both kinds will be the palette: a color gradient doesn't work for discretes, so we should provide the current color palette just for continuous (numeric) values; and a contrasted palette for categories.

Crashillo:
In the _summary_ tab, I'd show just a static map with the geometry column. That is, no distintion between discrete/continuous variables, no legend, no tooltip, no interaction... a plain color to tell the user: "hey, this dataset has geospatial data, look at this". In the way **dbeaver** does: 
imagen
 
However, in the Map tab, I'd show an interactive plugin based on the `currentQuery` values, full-featured (legend, tooltip, zoom)... actually the same behaviour as the mapplugin in the editor tab has, but well, with no chance to change the plugin type or any setting.

ferblape:
> What do you mean by easy way? Selecting the first numeric value is being done also by the other perspective plugins. But regarding this, if we gonna depict categorical variables, we won't need anymore to find the first numeric column, but we'll face a similar issue instead: which column? the first one? I don't think this matter must be solved by the plugin beyond distinguish the variable types: numeric -> continuous vs. text -> discrete

Is it easy to pass an argument to the plugin with which column should use as data? We are considering using a special custom field at dataset level which could be the data column to be loaded by default in the map.

Crashillo:
Ok, done. I've set everything up to receive a custom-field called _geom_, string-type, to tell the visualization which column contains the geospacial data.

Therefore, from the backoffice, just add another field (similar to _slug_, _category_, _frequency_...) called geom with the column name. If none is passed, the plugin looks for a column called _geometry_.

Crashillo:
From now on the legend, instead of searching the first numeric value, it gets the first element in the dataset<sup>1</sup>, whatever it is. If numeric, it uses a continuous scale, otherwise, it uses a categorical scale.

Both scales define a set of 5 colors each, which they can be overwritten with CSS custom properties, `--perspective-map-category-1` (discretes) and `--perspective-map-range-1` (continuous) in the site backoffice.

<sup>1</sup> This could be similar to _geometry_ column, get a custom-field from the backoffice (not implemented yet)

Crashillo:
The missing tests are no longer activated, but still here present

ferblape:
> The missing tests are no longer activated, but still here present

Yes, see #3826

Crashillo:
Now the map has an additional tab whenever the dataset contains a column called _geometry_, and also Summary tab, includes another map.

When you edit something at _Map_ tab (change filters or columns), the value is stored in sessionStorage. Therefore you can go to _Datos_ tab, and restore such changes.

ferblape:
Hey, now the npm package is ready, do you need to remove code from the PR?

jorgeatgu:
> Hey, now the npm package is ready, do you need to remove code from the PR?

Yes. Now we uses our package to render the map :)

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -45,134 +44,103 @@ export default {
...
-          schema[key] = 'string'
-        } else if (schema[key] === 'decimal') {
-          schema[key] = 'float'
-        } else if (schema[key] === 'inet') {
-          schema[key] = 'integer'
-        } else if (schema[key] === 'date') {
-          schema[key] = 'datetime'
-        }
-      });
+      // Object.keys(schema).forEach((key) => {
```
ferblape: Can we remove these comments?

ferblape (approved):
Code is ok!

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
```diff
@@ -45,134 +44,91 @@ export default {
...
+
+      // Clean any previous config
+      this.resetConfig();
+
+      this.viewer.setAttribute("columns", JSON.stringify(columns.split(",")))
+      if (this.configMap) {
+        const { metric, zoom = false } = this.configMap
+        this.viewer.setAttribute("metric", metric || "")
+        this.viewer.setAttribute("zoom", zoom || "")
+      }
```
Crashillo: Event though we don't do nothing else yet, I think would be more appropriate set just one attribute, for instance, `config-map=""{\"zoom\":2,\"metric\":\"column\"}""`, i.e. a stringified object, in terms of a future enhancement where we need extra stuff.
Allowing an object from now on, we won't have such problem later

Crashillo (changes requested):

app/javascript/gobierto_data/webapp/components/commons/Visualizations.vue
jorgeatgu: Done!

Crashillo: That's done here. But is not in https://github.com/PopulateTools/perspective-map/blob/main/src/index.js. You will must adapt this attribute to be parsed from the plugin, and remove the old attributes

Crashillo (approved):

PR #3843, opened 2021-04-24
3837 add rake tasks to run every module only

### LINKED ISSUE #3837: Improve test run by module as rake task
I'd like to have modularized the test run by gobierto module. Something like:

```
bin/rails test:admin # which run gobierto_admin module 
bin/rails test:attachments
bin/rails test:budget
bin/rails test:calendars
bin/rails test:cms
bin/rails test:common
bin/rails test:core
bin/rails test:dashboard
bin/rails test:data
bin/rails test:export
bin/rails test:indicators
...
```
because this is tedious run and error prone

```
bin/rails test test/controller/gobierto_admin
bin/rails test test/models/gobierto_admin
bin/rails test test/systemgobierto_admin
....
```

I would take advantage of circle ci config.yml

stbnrivas:
Closes #3837

## :v: What does this PR do?

add new rake task to run test by every gobierto module

```
bin/rails -T # should show new task like:

bin/rails test:gobierto_admin
...
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -113,6 +113,7 @@ group :development, :test do
   gem "i18n-tasks"
   gem "spring"
   gem "puma"
+    gem "open3", "~> 0.1.1"
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -113,7 +113,7 @@ group :development, :test do
   gem "i18n-tasks"
   gem "spring"
   gem "puma"
-    gem "open3", "~> 0.1.1"
+  gem "open3", "~> 0.1.1"
 end
 
 group :test do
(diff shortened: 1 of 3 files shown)


ferblape (approved):

ferblape:
What's pending here?

stbnrivas:
I am not sure about:

- namespace location...
- use the full name of module example: gobierto_admin
- if I will better use a gem like Open3 instead of puts

ferblape:
> * namespace location...

Yes, use a new  file (test.rake?). Let's have a single namespace per file

> * use the full name of module example: gobierto_admin

But you are already using the whole module name, right?

> * if I will better use a gem like Open3 instead of puts

The experience would be better if the output is shown instantly so Open3 might help, but don't  lose too much time

stbnrivas:
ok, I changed :
- move to lib/test.rake
- the command to output 
- add --fail-fast as params to task with
```
bin/rails test:gobierto_plans[true]
```

I will include Open3 next time

PR #3845, opened 2021-04-25
Update ruby version

juancata:
### Before this PR
gobierto_snyk_findings

### After this PR
gobierto_snyk_ruby_version_updated

- updated entry in `Dockerfile`


ferblape (changes requested):
Thanks for your PR @juancata 

Could you change also https://github.com/PopulateTools/gobierto/blob/master/.ruby-version and [CircleCI config](https://github.com/PopulateTools/gobierto/blob/master/.circleci/config.yml)

juancata:
Sorry for the delay, I've updated .ruby-version and CircleCI configuration

PR #3847, opened 2021-04-27
Visualizations performance

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1276

## :v: What does this PR do?
- Hide the spinner when the vue application is actual mounted
- Mount in the DOM only for the current route path


ferblape:
I don't fully understand how has it improved, but it loads much faster now

Crashillo:
Wait there, there's a bug https://santfeliu.gobify.net/visualizaciones/contratos. The spinner doesn't hide.

> I don't fully understand how has it improved, but it loads much faster now

https://github.com/PopulateTools/gobierto/blob/5afe580baacf4cc07f0e063812f2c864eea73817/app/javascript/gobierto_visualizations/webapp/containers/subsidies/Home.vue#L15-L17

It's about the `v-if` directive. `v-show` simply applies a `display:none` to the markup, but all stuff is in the DOM. There's no reason to render a component if the user is not at such page. Those components are acting like a router, actually.

ferblape:
> Wait there, there's a bug https://santfeliu.gobify.net/visualizaciones/contratos. The spinner doesn't hide.

It's working fine to me, what do you mean?

Crashillo:
imagen

Don't you see this?

Crashillo:
I figured out the problem. I fixed in subsidies, but I didn't in costs & contracts.

❤️ Lovely repeated code.

PR #3859, opened 2021-05-04
add custom field gobierto-default-geometry-data-column

### LINKED ISSUE #3841: Data / Add new custom field in the seeds
Update gobierto data module seeds and add a new custom field of type string (no translatable) called  "Gobierto default geometry data column" and with uid  `gobierto-default-geometry-data-column`

This field will be used by the maps visualization to fetch the data to show in the map from that column

stbnrivas:
Closes #3841 

## :v: What does this PR do?
add a custom field with uid `gobierto-default-geometry-data-colum` into GobiertoData::Dataset, this field will be used by the maps visualization to fetch the data to show 

it need to run:

```
bin/rails c
site = Site.find_by(domain: "**********")
GobiertoSeeds::GobiertoData::Recipe.run(site) if site.configuration.gobierto_data_enabled?
```

### DIFF
--- a/db/seeds/gobierto_seeds/gobierto_data/recipe.rb
+++ b/db/seeds/gobierto_seeds/gobierto_data/recipe.rb
@@ -122,6 +122,14 @@ module GobiertoSeeds
           }
           license.save
         end
+
+        dataset_default_geometry = site.custom_fields.string.where(class_name: "GobiertoData::Dataset").find_or_initialize_by(uid: 'gobierto-default-geometry-data-column')
+        if dataset_default_geometry.new_record?
+          dataset_default_geometry.name_translations = { ca: "Columna per llegir dades de geometria", en: "Column to read geometry data", es: "Columna para leer datos de geometría" }
+          dataset_default_geometry.position = 7
+          dataset_default_geometry.options = { configuration: {} }
+          dataset_default_geometry.save
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


db/seeds/gobierto_seeds/gobierto_data/recipe.rb
```diff
@@ -122,6 +122,14 @@ def self.run(site)
           }
           license.save
         end
+
+        dataset_default_geometry = site.custom_fields.where(class_name: "GobiertoData::Dataset").find_or_initialize_by(uid: 'gobierto-default-geometry-data-column')
+        if dataset_default_geometry.new_record?
+          dataset_default_geometry.name_translations = { ca: "Recinte", en: "Enclosure", es: "Recinto" }
```
ferblape: Hahahah, what are these names?

ferblape: Also, you need to tell the type of custom field. See in the above code that the query is `site.custom_fields.string.where` (the **.string.** is missing)

stbnrivas: I though about the field will contains coords or similar, a poligon of leaflet ... an enclosure where values dataset belongs to....
no problem geometry better?

ferblape: Sorry, I read "Reciente", I though you copied from a catalan translation.

This field will be used by the maps visualization to fetch the data to show in the map from that column. You can call it "Column to read geometry data"

PR #3879, opened 2021-05-13
Gobierto Visualizations / Filters not working

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1289

## :v: What does this PR do?

If the table is active, the filters don't work. When filtering the data, it will not find the dc.js chart because the component has been destroyed with the `v-if`.

Staging


eltercero:
Test added.

ferblape (approved):

PR #3880, opened 2021-05-13
995 GobiertoData add geometry support for neighbourhood in datasets

stbnrivas:
Related /PopulateTools/issues#995

## :v: What does this PR do?

This PR give support to datasets for `has_one` Neighbourhood model which have a geojson into geometry attribute. 
However this geometry will show this into the api for request datasets

- this PR don't break existing dataset without Neighbourhood relation
- appear into the API new attributte named geometry
-  until the ETL not ready geometry in all dataset should appear as `null`

## :eyes: Screenshots
Screenshot from 2021-05-13 10-10-09

Screenshot from 2021-05-13 09-42-05


stbnrivas:
WIP: I need to do some change before about the way to show into the API

don't merge yet

ferblape:
Closing, as this is not the correct approach

PR #3900, opened 2021-06-03
Gobierto Data / Testing embed

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1293

## :v: What does this PR do?
- Add Jest to test components that use JavaScript (not Vue).
- To run the test, I had to extract the function that creates the embed and split it into two, one for the API request and one for rendering the Perspective visualization.
- mock data to avoid making API requests
- Add Jest to CircleCI

The test fails:
  - If we use deprecated methods in Perspective, i.e, [viewer.clear()](https://github.com/PopulateTools/gobierto/pull/3896)

## 📹: Video

[LOOM](https://www.loom.com/share/1fabdbbec42a4c6f99018a41f12c9a6d)


ferblape (changes requested):
Loooking great! Could you add an extra step in CircleCI to run javascript? You just need to copy a existing one and instead of running bin/rails test just run `yarn run test`

.circleci/config.yml
```diff
@@ -358,6 +358,18 @@ jobs:
...
           path: test/reports
 
+  tests_javascript:
+    <<: *defaults
+    steps:
+      - prepare_workspace
+      - run:
+          name: Run tests
+          command: |
+            TEST="$(find test -path '*javascript*' -name '*.test.js' | circleci tests split --split-by=timings)"
```
ferblape: This is not necessary ;)

test/javascript/gobierto_data/__mocks__/data.json
```diff
@@ -0,0 +1,4 @@
+{
```
ferblape, line 1: Yeah

Crashillo (approved):

PR #3903, opened 2021-06-07
fix nil option for select

### LINKED ISSUE #3827: Add blank option for custom_field of vocabulary_options
we need to add a blank option for reset vocabulary option to be able unset value, for instance: 

1. at begins option is not set
Screenshot from 2021-04-19 12-52-24

2. select some option
Screenshot from 2021-04-19 12-52-41

3. unset can not be done
Screenshot from 2021-04-19 12-52-55

stbnrivas:
Closes #3827

## :v: What does this PR do?

set the min-height for nil option of custom field select.

### Before this PR
Screenshot from 2021-06-07 16-24-11

### After this PR
Screenshot from 2021-06-07 16-23-49

### DIFF
--- a/app/assets/stylesheets/module-admin.scss
+++ b/app/assets/stylesheets/module-admin.scss
@@ -931,6 +931,10 @@ a .global_menu {
   }
 }
 
+ul.select2-results__options li.select2-results__option:first-child {
+  min-height: 16px;
+}
+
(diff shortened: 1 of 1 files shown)


ferblape:
Is this released to staging?

ferblape:
It was not, I already did it. It works fine, thanks!

PR #3907, opened 2021-06-23
3906 sanitizer from html event description

### LINKED ISSUE #3906: Gobierto Calendars / Clear event description before storing
In calendars integration, the description of the events is imported without any kind of cleanup. There are some issues with events created with inline HTML that are resulting in invalid content.

Review the import calendar process and update the [form responsible](https://github.com/PopulateTools/gobierto/blob/master/app/forms/gobierto_people/calendar_sync_event_form.rb) of storing imported events to cleanup any HTML from the description. 

Keep in mind that the text should be cleanup from html tags, not removed. There are Ruby libs and Rails libs ([html-sanitizer](https://github.com/rails/rails-html-sanitizer)) to do this. Investigate and propose one.

stbnrivas:
Closes #3906 

## :v: What does this PR do?

when an event is imported its attributte description can have html we need sanitize it

- create a new event with html into description attributte
- import it using task ~/gobierto/lib/tasks/gobierto_people/sync_agendas.rake

### DIFF
--- a/app/forms/gobierto_calendars/event_form.rb
+++ b/app/forms/gobierto_calendars/event_form.rb
@@ -143,8 +143,8 @@ module GobiertoCalendars
         event_attributes.site_id = site_id
         event_attributes.state = state
         event_attributes.title_translations = title_translations
-        event_attributes.description_translations = description_translations
-        event_attributes.description_source_translations = description_source_translations
+        event_attributes.description_translations = description_translations.transform_values { |value| sanitize(value) }
+        event_attributes.description_source_translations = description_translations.transform_values { |value| sanitize(value) }
         event_attributes.starts_at = starts_at
         event_attributes.ends_at = ends_at
         event_attributes.meta = meta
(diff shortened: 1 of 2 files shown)


app/forms/gobierto_calendars/event_form.rb
```diff
@@ -143,8 +143,8 @@ def assign_event_attributes
         event_attributes.site_id = site_id
         event_attributes.state = state
         event_attributes.title_translations = title_translations
-        event_attributes.description_translations = description_translations
-        event_attributes.description_source_translations = description_source_translations
+        event_attributes.description_translations = description_translations.transform_values { |value| sanitize(value) }
```
ferblape, line 146: I didn't know transform_values 😅 , nice!

stbnrivas: :sweat_smile:  neither me, I just discover into ruby doc

PR #3920, opened 2021-07-16
1312 GobiertoPlans add link into admin to export indicators

stbnrivas:
## :v: What does this PR do?
Into module GobiertoPlans at backoffice add a link to export indicators only when there are indicators created. there are to alternatives:

1. if there are not indicators values created (even with indicators created aka custom field plugin table)
Closes PopulateTools/issues#1312
Screenshot from 2021-07-26 09-05-00

2. with indicators values created that show the link

Screenshot from 2021-07-26 09-06-11

- visit backoffice
- plans module
- visit any plans
- configuration tab
- click link to "Indicators" after "Export to CSV:"

[https://diba-plan.gobify.net/admin/plans/41/edit](https://diba-plan.gobify.net/admin/plans/41/edit)
[https://diba-plan.gobify.net/admin/plans/43/edit](https://diba-plan.gobify.net/admin/plans/43/edit)

- you should get an csv like below:
Screenshot from 2021-07-21 13-24-05

- after download the csv file yo can modify remove or adding lines and try load indicators to check if generated file match with format. (remember the custom field record should be created before, therefore it should be load into same plan)


app/assets/stylesheets/module-admin.scss
```diff
@@ -418,6 +418,10 @@ body.gobierto_admin {
   margin-top: 220px;
 }
 
+.admin_side_actions .simulate-a {
```
furilo: Why this? This does not make any sense. Please remove. And the corresponding call in the HTML.

stbnrivas: hi, this style is required to get and output uniform like below (note the gray color instead of primary)

1. with this style
Screenshot from 2021-07-23 09-40-58

2. without force style
Screenshot from 2021-07-23 09-40-41

should remove or refactor class name?

thx

furilo:
Grey color is what should be for the text that is not linked.

app/decorators/gobierto_plans/indicator_data_decorator.rb
```diff
@@ -0,0 +1,62 @@
...
+    def csv_row(object)
+      current_columns = object.payload[object.custom_field.uid.to_s].first&.size || 0
+      @columns = current_columns if current_columns > @columns
+      values = object.payload[object.custom_field.uid.to_s].map do |record|
+        values_keeping_order(object,record)
+      end
+
+      values.map do |value|
+        [
+          plan.title_translations["en"],
```
entantoencuanto: Why force translations in "en" locale? I think the name of the item should appear here instead of the plan title

app/models/gobierto_plans/plan.rb
```diff
@@ -35,6 +35,14 @@ def nodes
       @nodes ||= Node.joins(categories: [:vocabulary]).where(terms: { vocabulary_id: categories_vocabulary&.id }).order(position: :asc, id: :asc)
     end
 
+    def have_indicators?
+      GobiertoCommon::CustomFieldRecord.includes(:custom_field).where(custom_fields: { instance_id: id, field_type: 10 }).count > 0
+    end
```
entantoencuanto: You can reuse the `indicators` method for this
```suggestion
    def have_indicators?
      indicators.exists?
    end
```

app/models/gobierto_plans/plan.rb
```diff
@@ -35,6 +35,14 @@ def nodes
...
+
+    def indicators
+      GobiertoCommon::CustomFieldRecord.includes(:custom_field).where(custom_fields: { instance_id: id, field_type: 10 })
+    end
```
entantoencuanto: It's clearer to use the symbol or string associated to an enum attribute instead of its numeric value, use `field_type: :plugin`

The indicators belongs to custom fields of `type` `plugin` which also have `plugin_type` `table`. You should include the plugin type in the query to avoid errors, with something like:
```
GobiertoCommon::CustomFieldRecord.includes(:custom_field).where(
  custom_fields: { instance_id: id, field_type: :plugin }
).where(
  "#{GobiertoCommon::CustomField.table_name}.options @> ?", { configuration: { plugin_type: :table } }.to_json
)
```

app/decorators/gobierto_plans/indicator_data_decorator.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def values_keeping_order(object,record)
+      attributes_in_order = object.custom_field.options.dig("configuration","plugin_configuration","columns").map { |r| r["id"] }
+      attributes_in_order.map do |key|
+        record[key]
+      end
+    end
+
+    def csv_row(object)
+      current_columns = object.payload[object.custom_field.uid.to_s].first&.size || 0
```
entantoencuanto: Why not use `object.value`? This method is defined for custom field records [there](https://github.com/PopulateTools/gobierto/blob/4e6941f13b2bf37af73c2a7962e207d450eb1a4b/app/models/gobierto_common/custom_field_value/plugin.rb#L16-L20)

entantoencuanto (changes requested):
Great job!

Take a look at the comments I left and consider adding a test in https://github.com/PopulateTools/gobierto/blob/master/test/controllers/gobierto_admin/gobierto_plans/plans_controller_test.rb to check the csv output.

Review also some missing spaces in the arguments of functions and some missing frozen string literal comment

ferblape:
@entantoencuanto please confirm is ok and merge, so I can release

entantoencuanto (approved):
Great!

entantoencuanto:
Ready and merged @ferblape !

PR #3933, opened 2021-09-02
Remove old rake tasks from participation module

ferblape:
Unexpected

This PR removes a task that should have been removed when the class ActiveNotifier was removed in the participation module. It was causing an error everyday in the cron call.


ferblape:
This was the appsignal error https://appsignal.com/populate/sites/5eb0fcfc74782044d1198606/exceptions/incidents/248?timestamp=2021-09-01T05%3A00%3A26Z

PR #3936, opened 2021-09-07
Filter supercomponent: gobierto data

Crashillo:
Related to https://github.com/PopulateTools/gobierto/issues/3928

### What does this PR do?
Replaces the dataset filters sidebar with the built-in components. Remove old handler code afterwards


jorgeatgu (changes requested):
Looks good! Only two things: 
- A small detail in the direction of the caret and the animation of the dropdown.
- When you've selected a dataset, and then you play the filters.

https://user-images.githubusercontent.com/2649175/132496630-db55d632-26ef-4fef-96db-8b0556d8d3d0.mov

Crashillo:
Ok, now you can check out once again 👍

PR #3937, opened 2021-09-08
3935 visualizations homepage

### LINKED ISSUE #3935: Visualizations / Enable this module as the home page of a site
We need to be able to configure a visualization from Gobierto Visualizations module as the home page of a site.  This is a bit tricky and different to other modules, where a method is defined in the module definition:

```ruby
  def self.root_path(_)
    Rails.application.routes.url_helpers.gobierto_participation_root_path
  end
```

In this case the module consists on a set of visualizations defined in the JSON. I'd propose to set a new attribute (`home`: true`) in one of the visualizations to set that one as the visualization of the homepage. Something like:

```json
{
  "visualizations": {
    "costs": {
      "enabled": true,
      "home": true,
      "data_urls": {
        "costs": "https://...."
      }
    }
  }
}
```

@furilo if you agree with this solution I think it's not very difficult to implement /cc @entantoencuanto

entantoencuanto:
Closes #3935

## :v: What does this PR do?

* Changes the visualizations module to take the module root path from the module configuration. To set a visualization as the home include a `"home": true` option in its configuration. Note that if several visualizations are configured as home, only the first one will be taken into account.
* Adds an integration test
* Adds a fix to avoid exceptions if a path is requested for a visualization not configured in the module settings  
* Adds the route `'/' `to visualizations homepage. [front]

Configure GobiertoVisualizations as the home module in the site.
Change the module settings and include in a visualization the `"home": true` setting 
[staging](https://terrassa.gobify.net/)

No

Suggestion added to gobierto.readme.io

### DIFF
--- a/app/controllers/gobierto_visualizations/visualizations_controller.rb
+++ b/app/controllers/gobierto_visualizations/visualizations_controller.rb
@@ -11,7 +11,7 @@ module GobiertoVisualizations
     private
 
     def visualization_enabled!
-      render_404 unless visualizations_config.fetch('enabled', false)
+      render_404 unless visualizations_config&.fetch("enabled", false)
     end
 
     def visualizations_config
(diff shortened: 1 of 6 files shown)


furilo:
OK on my part.

ferblape (approved):
Ok to me! @entantoencuanto feel free to deploy

PR #3946, opened 2021-09-28
Enhancement edge cases for Plans

Crashillo:
## :v: What does this PR do?
Solve some situation when a plan comes with no custom fields, but there are categories and status settled up. Avoid the plan get stucked.

Test url: https://badia.gobify.net/planes/pam-2023/2020


furilo:
Plan projects are not opening correctly CleanShot 2021-09-28 at 14 55 02@2x

Also, what are we using interact.js?

Crashillo:
> Plan projects are not opening correctly 

Projects weren't published. Now they are.

> Also, what are we using interact.js?

Using where? or doing what? 🤔

furilo (approved):

PR #3953, opened 2021-10-01
1358 prefix default vocabularies

entantoencuanto:
Closes PopulateTools/issues#1358

## :v: What does this PR do?

* Adds a concern to prefix translations with module name translations (taken from admin layouts translations)
* Uses the concern to prefix default vocabularies used in Data and Plans modules

Visit https://cortegada.gobify.net/admin/vocabularies

### After this PR

Screenshot from 2021-10-01 20-03-37

No

No


ferblape (approved):
Works fine, tested in staging

PR #3967, opened 2021-10-19
Gobierto Data / Add a new type for date

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1375

## :v: What does this PR do?

Create a new type called datetime for dates containing HH:MM:SS.

Caveats: Perspective transforms the hours incorrectly, depending on the month add one or two more hours.

related https://github.com/finos/perspective/issues/1252

## :eyes: Screenshots

**Dataset loaded with `type: timestamp`**

type_timestamp

**Dataset loaded with `type: string`**

type_string

**Dataset loaded with the new `type: datetime`**

type_datetime

Yes, changes submitted. [datasets csv schema](https://gobierto.readme.io/docs/datasets-csv-schema)

### DIFF
--- a/app/queries/gobierto_data/sql_function/transformation.rb
+++ b/app/queries/gobierto_data/sql_function/transformation.rb
@@ -38,12 +38,6 @@ module GobiertoData
           sql: "select (to_date(nullif(trim($1), '#null_value'), '#date_format'));",
           optional_params: { date_format: "YYY-MM-DD", null_value: "" }
         },
-        datetime: {
-          input_type: "text",
-          output_type: "timestamp",
-          sql: "select (to_timestamp(nullif(trim($1), '#null_value'), '#date_format'));",
-          optional_params: { date_format: "YYY-MM-DD HH24:MI:SS", null_value: "" }
-        },
         time: {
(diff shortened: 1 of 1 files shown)


ferblape:
Adding one or two hours sounds to me to time zone issue, because sometimes we are 1 hour away from UTC and others we are 2. Maybe Perspective is using your browser locale?

ferblape:
@entantoencuanto the documentation is a bit confusing without examples, to me timestamp and datetime sound the same. Could you provide an example in the same sentence that describes the type:

Screenshot 2021-11-16 at 15 43 11

Also, there's a typo in timestampz, it says timestamp **without** timezone, it should be **with** timezone, isn't it?

entantoencuanto:
I've removed the new datetime type and included the time as `HH24:MI:SS` in the default format expected both by timestamp and timestamptz format. I've seen that this addition doesn't generate errors if the received values only contain a date without time.

With these changes the new type datetime is not necessary since timestamptz works the same 

A suggestion with updates has been sent to gobierto.readme.io

entantoencuanto (approved):

PR #3998, opened 2021-11-12
Gobierto Data / Fix the permalink of the visualizations

jorgeatgu:
Closes https://github.com/PopulateTools/issues/issues/1411

## :v: What does this PR do?

Fix the permalink of the vizualizations. For the permalink we need to get the slug of the datasets that have visualizations in the page.

[Staging](https://getafe.gobify.net/datos/v/visualizaciones)

### After this PR

after-vizzs

### DIFF
--- a/app/javascript/gobierto_data/webapp/components/landingviz/VisualizationsGrid.vue
+++ b/app/javascript/gobierto_data/webapp/components/landingviz/VisualizationsGrid.vue
@@ -3,7 +3,7 @@
     v-if="publicVisualizations && publicVisualizations.length"
     class="gobierto-data-visualization--grid"
   >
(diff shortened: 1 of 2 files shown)


furilo:
If I go into https://getafe.gobify.net/datos/v/visualizaciones in Firefox and click on some vizzs, I don't get redirected.

jorgeatgu:
The problem with Firefox is related to https://github.com/PopulateTools/issues/issues/1421 so maybe we should review and merge to restore the permalinks.

Crashillo (approved):

ferblape (approved):
I've been playing around and looks fine. Ok to improve the performance in firefox with the visualization screenshots

PR #4008, opened 2021-11-17
Add HTTP cache strategy

ferblape:
This PR adds a Cache-Control header for public requests with a default TTL of 1 day. For private connections (logged in users, admin...) it set's the header value to private.

This allows to cache sites with a CDN.


ferblape:
Thanks!

PR #4009, opened 2021-11-17
1416 cache improvements

entantoencuanto:
Closes PopulateTools/issues#1416
Closes PopulateTools/issues#1413

## :v: What does this PR do?

* Defines a cache service which can be initialized for site and module and keeps cached data keys separated by prefixes
The cache service allows to clear cached data only on the module and site (currently is not working because it's prepared to use Redis)
* Extracts complex transformations of custom fields values to a service which makes use of the cache service (in this way the calculations can be called from controllers, tasks or ETLs)
* Uses the custom fields service from all API endpoints which returns resources with associated custom fields improving performance
* Uses the custom fields service to improve performance of sdgs section in plans pages avoiding a lot of requests to database

Visit the index of datasets or a plan with sdgs and a lot of projects

No

No


ferblape:
I'm missing the changes in the `config/environments/...` to support Redis. Please follow [this guide](https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore) to include:

- replace redis gem by hiredis
- define a new environment variable for REDIS cache store database (it **must** be different database that the queues cache). This should also be done in Ansible
- configure cache_store in development, staging and production to use Redis new URL

.env.example
```diff
@@ -43,6 +43,7 @@ GOOGLE_MAPS_GEOCODING_API_KEY=
 
 # Redis instance URL
 REDIS_URL=redis://localhost:6379/0
+REDIS_CACHE_URL=redis://localhost:6379/0
```
ferblape:
```suggestion
REDIS_CACHE_URL=redis://localhost:6379/1
```

Suggest a different database

.env.example
```diff
@@ -43,6 +43,7 @@ GOOGLE_MAPS_GEOCODING_API_KEY=
 
 # Redis instance URL
 REDIS_URL=redis://localhost:6379/0
```
ferblape: Should we rename this variable to something like REDIS_SIDEKIQ_URL to be explicit about the purpose of it?

app/services/gobierto_common/cache_service.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # These methods are very inefficient
+    def keys
+      @keys ||= Rails.cache.redis.keys.select do |key|
+        /^#{cache_prefix}\//.match? key
+      end
+    end
+
+    def clear_with_keys
+      @keys = nil
```
ferblape: If you clear the memoization here and is not used anywhere else, should we remove the memoization?

app/services/gobierto_common/cache_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def prefixed(name)
+      "#{cache_prefix}/#{name}"
+    end
+
+    # These methods are very inefficient
+    def keys
+      @keys ||= Rails.cache.redis.keys.select do |key|
+        /^#{cache_prefix}\//.match? key
```
ferblape:
```suggestion
        /\A#{cache_prefix}\//.match? key
```

app/services/gobierto_common/cache_service.rb
ferblape: Do we need this method yet?

app/services/gobierto_common/cache_service.rb
```diff
@@ -0,0 +1,47 @@
...
+      Rails.cache.fetch(prefixed(name), options, &block)
+    end
+
+    def prefixed(name)
+      "#{cache_prefix}/#{name}"
+    end
+
+    # These methods are very inefficient
+    def keys
+      @keys ||= Rails.cache.redis.keys.select do |key|
```
ferblape, line 34: Hmmmm, if you asume that Rails.cache.redis works then should you configure redis also in development and testing environments?

app/services/gobierto_common/cache_service.rb
entantoencuanto: You are right, it's not really necessary anymore

.env.example
entantoencuanto: Good point, I agree

.env.example
entantoencuanto: Hum, redis://localhost:6379/1 is already in use in config/cable.yml. Shouldn't we maybe change it?

app/services/gobierto_common/cache_service.rb
entantoencuanto: In development environment is configured, but not in testing. This method is used to debugging purposes, so I can make it only available in development environment... Or may should I define a different REDIS_CACHE_URL for test environment and enable it there? I'm not sure if it requires also changes in circleci configuration?

Issue #3240, opened 2020-07-16
Gobierto / Activate AppSignal js errors


ferblape:
(and remove Rollbar)

furilo:
Can we remove Rollbar, right? And also Mixpanel.

ferblape:
Yes:

1. Activate Appsignal JS and check it works
2. Then remove Rollbar and Mixpanel js

furilo:
AppSignal is already activated, right?

>

ferblape:
no the js library

ferblape:
@Crashillo now that you have access to Appsignal you can re-review this issue. The idea is to check if there's a way to get a sample of notifications, and if we can limit them to modern browsers.

PR #4044, opened 2022-02-08
4032 migrate to ahoy

### LINKED ISSUE #4032: Migrate from Google Analytics and Mixpanel to Ahoy
We are removing application  Google Analytics service and replacing it by Ahoy. If a customer setup its own analytics code, it'll be kept, meaning customers that don't set an Analytics code will have Ahoy stats.

Right now there are two analytics signatures installed:

- by default, the application installs the code defined in `config/application.yml`setting
- optional, the site has an attribute to set a specific code. We are keeping this code if has been set

## First part: remove Analytics

- [ ] remove the attribute from `config/application.yml` and update docs
- [ ] remove the app analytics JS from the templates. If the customer is using it's own code, keep it
- [ ] extra ball: remove references to Mixpanel

## Second part: add Ahoy track

- [ ] install ahoy gem
- [ ] track site visits, probably in an after filter
- [ ] track visits with a sidekiq job
- [ ] configure it to track site requests. Keep in mind the multitenancy, so you might wan to add the site host to the ahoy event

entantoencuanto:
Closes #4032 

## :v: What does this PR do?

* Removes mixpanel and google analytics tracking code (except if analytics is configured on site)
* Adds and configure ahoy to track each request creating an event by site. This includes adding a column to the events table and an association with Site model.
* Adds sidekiq-bounce and defines a `GobiertoCommon::EventCreatorJob` to create ahoy events which is used from controller concern for each request
* The event stores info of the request and its params filtering some. The ahoy config file also includes some GDPR compliance settings and:
  * Masks IPs on visits
  * Disables use of cookies by ahoy

- [x] deleted entries in `config/application.yml`?

Pending


app/controllers/concerns/application_concern.rb
```diff
@@ -56,6 +61,12 @@ def authenticate_user_in_site
     end
   end
 
+  def track(name, properties = {})
+    ahoy.track name, **properties.merge(site_id: current_site.id)
+  rescue StandardError => e
+    Rollbar.error(e)
```
furilo: @entantoencuanto Rollbar?

entantoencuanto: I've added the rescue to avoid exceptions trying to generate ahoy events, but the use of Rollbar after the rescue is optional

furilo:
But we are no longer using Rollbar (or won't soon), but Appsignal
instead...

ferblape:
FYI, we still have rollbar because of the JS https://github.com/PopulateTools/gobierto/issues/3240. I'm going to move this issue to Business as usual

app/controllers/concerns/application_concern.rb
```diff
@@ -56,10 +59,29 @@ def authenticate_user_in_site
...
+  def track_request
+    return true if ignore_tracking_request?
+
+    GobiertoCommon::EventCreatorJob.bouncer.debounce current_user&.id, current_site.id, filtered_params(request.params.merge(method: request.method))
+  rescue StandardError => e
+    Appsignal.send_error(e)
+  end
+
+  def ignore_tracking_request?
+    Rails.env.development?
```
ferblape: Add Rails.env.test?

ferblape (changes requested):
Looks good, just two things:

- should it be tested in some way
- remember to add the ansible PR and update the documentation in regards to the modifications of the `config/application.yml`

ferblape:
Hmmm, this is not working well, if you check the latest events from ahoy_events table they are without visit_id nor site_id (I'm generating them visiting Getafe staging site). Please review that

```
-[ RECORD 1 ]---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
----------------------------------------------------------------------------------------------------------------------------------
id         | 605
visit_id   |
user_id    |
site_id    |
name       | gobierto_people/people#show
properties | {"slug": "rafael-garcia", "action": "show", "locale": "es", "method": "GET", "controller": "gobierto_people/people"}
time       | 2022-02-17 05:01:23.547261
```

furilo:
It's not the same as in contratos Edu?

entantoencuanto:
I made a change to not use bounce, and by mistake it was not saving the site correctly for visits with not authenticated users, but the schema of events and method to name them (controller#action) are the same. The only difference is that instead of using an `organization_id` gobierto events use a `site_id`, I can change it.

After talking with Victor I am using a different way of associating visits and events that I think can also be useful in contratos: At the moment of enqueuing the job to create the event from the controller there is a current_visit available which is passed to the job

ferblape:
@entantoencuanto thanks, now ahoy_events are tracking the site_id properly.

Would it be possible to add the site_id also to the field ahoy_visits? If we want to give the metric unique visitors to a single site right now it's not possible (please confirm @furilo).

furilo:
Yes, that is basic.

entantoencuanto:
Yes, it's possible, but  currently the same visit can use different site_ids. For example, currently if I change in staging of site from the same browser the visit remains the same. I can investigate if a different visit can be generated for each site

Or may I do the same as I did with the user_id field, I can update the site_id when it changes

ferblape:
If the same user visits two sites it should be tracked as two visits, because sites should be independent. Customers don't care if the application is multitenant or there are N applications.

furilo:
Since a tenant only sees his visits, I don't think it's a problem thats
shared. I would not complicate it

ferblape (approved):
Now it works as expected!

ferblape:
@entantoencuanto there's something I missed in the acceptance that is broken. Your PR doesn't work when a customer has a google analytics key setup (the first point of your PR description). See that you are removing the `ga` function, while it's necessary: https://rollbar.com/Populate/gobierto/items/6057/?utm_campaign=exp_repeat_item&utm_medium=email&utm_source=rollbar-notification&utm_content=control

PR #4056, opened 2022-03-01
1510 project titles plans exports

entantoencuanto:
Closes PopulateTools/issues#1510

## :v: What does this PR do?

* Improves the detection of locale exporting projects of a plan to a CSV:
  * Uses the current locale
  * If current locale is not defined then uses the plan site default locale
  * If there are no translations for the locale previously defined the export uses the first translation found
* Reorders the columns in the CSV to export the first columns in the following order: Level 0, Level 1, Node.Title, Node.external_id, Node.Status, Node.Progress, Node.Start, Node.End
* Orders the exported rows using the external_id. If the exporter detects that the external_id is numeric it orders them numerically and alphabetically elsewhere

Visit a plan and export it from admin

No

No


ferblape (changes requested):
Maybe it's a good time to add some tests around plan exports?

ferblape (approved):

PR #4085, opened 2022-05-10
1362 cache datasets meta data summary

entantoencuanto:
Related to  PopulateTools/issues#1362

## :v: What does this PR do?

When tables associated to datasets have a big number of rows the meta endpoint of datasets takes too long the first time (the response is cached). This PR:
* Uses data cache service to store the result of the rows count query in a rows_count method defined in the dataset model. This method is called by the dataset metadata endpoint
* Adds a job to cache the method result in the callback after uploading a dataset
* The cache key includes the timestamp of the dataset. To avoid expiring the cache key the PR changes the dataset_cache operation (which saves the csv of dataset into a file) to not change the dataset timestamp 

Run a datasets load ETL with huge tables and visit a dataset for the first time. The page load should take less than 1s

No

No

### DIFF
--- a/app/models/gobierto_data/dataset.rb
+++ b/app/models/gobierto_data/dataset.rb
@@ -99,10 +99,10 @@ module GobiertoData
       cache_service.fetch("#{cache_key_with_version}/rows_count") do
         count_result = ::GobiertoData::Connection.execute_query(
           site,
-          Arel.sql("SELECT COUNT(1) FROM #{table_name} LIMIT 1"),
+          Arel.sql("SELECT reltuples::bigint AS estimate FROM pg_class WHERE oid = 'public.#{table_name}' ::regclass;"),
           include_draft: true
         )
-        count_result.first["count"] if count_result.is_a?(PG::Result)
+        count_result.first["estimate"] if count_result.is_a?(PG::Result)
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/models/gobierto_data/dataset.rb
```diff
@@ -95,6 +95,17 @@ def columns_stats
       end
     end
 
+    def rows_count
+      cache_service.fetch("#{cache_key_with_version}/rows_count") do
+        count_result = ::GobiertoData::Connection.execute_query(
+          site,
+          Arel.sql("SELECT COUNT(1) FROM #{table_name} LIMIT 1"),
+          include_draft: true
```
ferblape, line 103: Would it makes sense to try what mentions here: https://arctype.com/blog/postgres-count/ ?

I mean the 
```
SELECT
    reltuples::bigint AS estimate
FROM
    pg_class
WHERE
    oid = 'public.sample_data' ::regclass;
```

ferblape: Because we just want an estimation, right?

entantoencuanto: You are right, this is much faster

ferblape (approved):

PR #4117, opened 2022-11-03
1715 refresh welcome cms cache

entantoencuanto:
Closes PopulateTools/issues#1715

## :v: What does this PR do?

This PR:
* Adds a callback on pages form when pages are saved which detects if the page saved is used as welcome page and deletes its associated cache entries.
* Adds a callback on sites form when saved and the home_page_item_id is changed to remove associated cache

Edit in admin the page configured as welcome page in the homepage and visit the root path of the site. The html should reflect the changes
Change home page from site configuration in admin and save. The home page in front should change also 

No

No

### DIFF
--- a/app/controllers/meta_welcome_controller.rb
+++ b/app/controllers/meta_welcome_controller.rb
@@ -8,32 +8,20 @@ class MetaWelcomeController < ApplicationController
   def index
     render_404 and return if current_site.nil?
 
-    if current_site.configuration.home_page != "GobiertoCms"
-      render_404 and return
+    # This code allows the homepage to render a page without redirecting to the page url
+    page = current_site.configuration.welcome_page
+    render_404 and return if page.nil?
+
+    if @section ||= page.section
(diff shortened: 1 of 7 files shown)


furilo:
Another email reply that didn't make it :/

CleanShot 2022-11-04 at 12 06 20@2x

ferblape:
Lazy sidekiq.... they should switch to elonkiq

app/controllers/meta_welcome_controller.rb
```diff
@@ -8,32 +8,20 @@ class MetaWelcomeController < ApplicationController
   def index
     render_404 and return if current_site.nil?
 
-    if current_site.configuration.home_page != "GobiertoCms"
-      render_404 and return
+    # This code allows the homepage to render a page without redirecting to the page url
```
ferblape, line 11: 👏🏾 great refactor!

ferblape (approved):

PR #4154, opened 2023-06-13
Making the iframe view to take 100% width

furilo:
Simplifies the layout for when the show_only_calendar view is in action. 

@ferblape it would be great if we could make that event links adds a target="_top", but I can't find how to do it.


ferblape:
@furilo I couldn't make the `target="_top"` because links are generated in the calendar library

Please review https://portalobert.esplugues.cat/agendas/pilar-diaz-i-romero?only_calendar=true, all the links keep the `only_calendar` parameter so embed view is not lost.

PR #4171, opened 2023-09-19
Upgrade SSL version when connecting to agendas

ferblape:
## :v: What does this PR do?

This PR updates SSL version to TLSv1.2

Connect to a Exchange server updated with the latests versions, and everything should work as expected.


ferblape:
Tested in the customers using Exchange and works fine.

PR #4172, opened 2023-09-21
Plans api

entantoencuanto:
Closes PopulateTools/issues#1845

## :v: What does this PR do?

* Implements and endpoint to get info as admin of a plan including their projects and use the same structure provided by the API to update the plan
* It also provides some endpoints to create, update and delete projects
* Adds tests both for plans and projects API controllers
* Fixes a bug which happened when there are more than one plan with custom fields with the same uid. In this case only the first custom fields were correctly updated and the others were updated with blank values

Documentation updated in readme.com

- [ ] new site template?
- [x] significant changes in some feature?

### DIFF
--- a/app/serializers/gobierto_plans/api_node_serializer.rb
+++ b/app/serializers/gobierto_plans/api_node_serializer.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module GobiertoPlans
+  class ApiNodeSerializer < BaseSerializer
+    include ::GobiertoCommon::Versionable
+    # include ::GobiertoCommon::HasCustomFieldsAttributes
+
+    attributes(
+      :id,
+      :external_id,
+      :visibility_level,
+      :moderation_stage,
+      :name_translations,
+      :category_id,
+      :category_external_id,
+      :status_id,
+      :status_external_id,
+      :progress,
+      :starts_at,
+      :ends_at,
+      :position
+    )
+
+    def category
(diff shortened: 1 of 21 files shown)


app/serializers/gobierto_plans/api_node_serializer.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module GobiertoPlans
+  class ApiNodeSerializer < BaseSerializer
+    include ::GobiertoCommon::Versionable
+    # include ::GobiertoCommon::HasCustomFieldsAttributes
```
ferblape: Can this comment be deleted?

app/serializers/gobierto_plans/api_node_serializer.rb
```diff
@@ -0,0 +1,48 @@
...
+      return if category.blank?
+
+      category.external_id
+    end
+
+    def status_external_id
+      object.status&.external_id
+    end
+
+    # def custom_fields
```
ferblape: Delete?

app/serializers/gobierto_plans/api_node_serializer.rb
entantoencuanto: Done

ferblape (approved):
Great work! I've also published the documentation API at https://dash.readme.com/project/gobierto/v0.1/docs/gobierto-plans-create-plan

Could you double check if it's 100% accurate with the current implementation?

PR #4179, opened 2023-11-05
Block other unsecure schemas for gobierto data

ferblape:
Closes https://github.com/PopulateTools/issues/issues/1861

## :v: What does this PR do?

This PR fixes a vulnerability reported in Gobierto Data module, which allowed to execute queries under the `information_schema.` schema, allowing users to list all the tables in the database.

It was not a big issue, because each site has it's own database, and no other tables could be listed, but makes our API more restricted to just the tables users should be querying.

In a gobierto data instance run the query `SELECT table_name FROM information_schema.tables`, the API should return `Error 400 Query blocked`


ferblape:
Confirmed in staging:

Screenshot 2023-11-05 at 07 40 24

PR #4198, opened 2024-02-09
Include cache service prefix in caches_action requests

entantoencuanto:
Closes PopulateTools/issues#1900

## :v: What does this PR do?

Fixes the use of `caches_action` requests in budgets controllers to include the cache service prefix. Doing this the cache service `clear` method called from ETLs for the module removes also that cached data

Review the update data after ETL, it should change daily
image

No

No


ferblape:
Checked in staging and looks good!

PR #4226, opened 2024-04-23
Update ca.yml for untranslated moderation status and few fixes

aramollis:
Close https://github.com/PopulateTools/issues/issues/1977

Maybe for those fixes the reviewer should be @ferblape instead of @furilo right?


furilo:
No, @crashillo

Crashillo (approved):

PR #4240, opened 2024-06-20
Run linter autofixable issues

Crashillo:
## :v: What does this PR do?
This PR runs automatic fixes provided by each linting rule:

```
eslint --fix
stylelint --fix
```
Such default rules are extended via:
- eslint-plugin-vue
- stylelint-config-recommended
- stylelint-config-standard-scss

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -2,7 +2,6 @@
 @import "purecss/build/pure-nr";
 @import "purecss/build/grids-responsive";
 @import "@fortawesome/fontawesome-free/css/all.css";
-
 @import "modules/css-conf";
 
 /* Override front fonts (defined in css-conf) */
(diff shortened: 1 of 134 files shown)


ferblape:
@Crashillo could you check the conflicts?

Crashillo:
Solved

ferblape:
Hmmmm, there's a broken spec which is not broken in master, makes me suspect. Could you take a look? Ping me if you need help

Crashillo:
Yep, I noticed before, I thought it would be related to the `master` conflicts. I see it's not. I'll take a look

Crashillo:
I'm not being able to solve these tests: in my local environment they pass, so I presumed they were flaky tests... but any time I run them again in the server, they fail.

ferblape:
Fixed by refreshing CircleCI cache

PR #4244, opened 2024-07-09
Fix bubbles chart

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/1997
Closes https://github.com/PopulateTools/issues/issues/1998

## :v: What does this PR do?
Fix static year positioning and bad constructor call

https://terrassa.gobify.net/


Crashillo:
@ferblape I had to enter staging console to do `Rails.cache.clear` once again. Lastly, that's becoming a recurring step, should we include it in the deploy task?

ferblape:
Great thanks, just in time for Reus

PR #4255, opened 2024-09-23
Action Lines table configurable

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/2007

## :v: What does this PR do?
Allow to display different custom fields in the action lines table, within the categories tree, apart of the name. By default, it shows `starts_at`. `status`, `progress`.

In order to alter the default columns, in the plans configuration json, the new property is called `show_table_extra_fields` and accepts an array of column names.

https://mataro.gobify.net/planes/pla-de-mandat/2023/categoria/21991

- [x] new site configuration variable?

https://gobierto.readme.io/docs/planificiacion#configuraci%C3%B3n

### DIFF
--- a/app/javascript/gobierto_plans/webapp/components/ActionLines.vue
+++ b/app/javascript/gobierto_plans/webapp/components/ActionLines.vue
@@ -1,19 +1,19 @@
 <template>
   <ul class="action-line--list">
-    <template v-for="model in models">
-      <ActionLine
+    <ActionLine
+      v-for="model in models"
+      :key="model.id"
+      :model="model"
+      :options="options"
+    >
+      <ActionLinesTableView
         :key="model.id"
         :model="model"
+        :header="showTableHeader"
+        :open="openNode"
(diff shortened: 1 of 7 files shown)


aramollis (approved):
Works fine! 👏
Client expect that "l-valoracions" & "statues" values appear with specific css style (button with specific color for each value). I will create a new issue for this.

Crashillo:
> Client expect that "l-valoracions" & "statues" values appear with specific css style (button with specific color for each value). I will create a new issue for this.

But that would be something editable exclusively for Mataró, that is, from the backoffice, wouldn't it?

aramollis:
> > Client expect that "l-valoracions" & "statues" values appear with specific css style (button with specific color for each value). I will create a new issue for this.
> 
> But that would be something editable exclusively for Mataró, that is, from the backoffice, wouldn't it?

Yes. I created this issue https://github.com/PopulateTools/issues/issues/2026

PR #4256, opened 2024-09-24
Read costs population data from Gobierto Datos

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/2008

## :v: What does this PR do?
Gets the population values from the _gobierto datos_ endpoint, replacing the hardcoded data. Also refactor some code blocks related.

https://mataro.gobify.net/visualizaciones/costes


Crashillo:
@ferblape I had to replace the API token in the backoffice for mataró backoffice. Is that right?

ferblape:
@Crashillo in staging?

Crashillo:
Yes, the current one you see here: https://mataro.gobify.net/admin/sites/8/edit is what I've edited

PR #4259, opened 2024-10-15
Handle empty categories

Crashillo:
Closes https://github.com/PopulateTools/issues/issues/2036

## :v: What does this PR do?
Better handling categories in order to avoid `NaN` values when the selected items implies empty elements.
Selected values in the charts are highlighted in italics, and the treemaps/beeswarm are hidden when no data

https://getafe.gobify.net/visualizaciones/contratos

### DIFF
--- a/app/assets/stylesheets/modules/_comp-graph-dc.scss
+++ b/app/assets/stylesheets/modules/_comp-graph-dc.scss
@@ -26,6 +26,10 @@
   &.deselected {
     fill: #ccc;
   }
+
+  &.selected + text tspan {
+    font-style: italic;
+  }
 }
 
 .dc-chart g.row text,
--- a/app/javascript/gobierto_visualizations/modules/contracts_controller.js
+++ b/app/javascript/gobierto_visualizations/modules/contracts_controller.js
@@ -505,19 +505,12 @@ export class ContractsController {
   _updateChartsFromFilter(options) {
     const container = this.charts[options.id].container;
 
-    // https://dc-js.github.io/dc.js/docs/html/BaseMixin.html#filter__anchor
-    // - filter(null) removes any existing filter.
-    // - If all filters are set at one (options.all), we first remove the existing ones
-    // - Note: when you add more than 1 filter, you need to add an array within an array
-    if (options.all) {
-      container.filter(null);
-      container.filter([options.titles]);
-    } else {
-      container.filter(options.title);
-    }
+    // apply the filters
+    container.filter(options.all ? null : options.title);
 
     Object.values(this.charts).forEach(chart => {
-      const hasData = chart.container.data().reduce((acc, item) => acc + item.value, 0) !== 0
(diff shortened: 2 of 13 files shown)


ferblape:
Maybe you can hide also these selectors 

Screenshot 2024-10-17 at 07 03 28

Also in the lists, could you include a message like "There are no tenders for this criteria" or similar?

Crashillo:
> in the lists

You mean the _Contracts_ tab, dont' you?

ferblape:
Yes

Crashillo:
When comment in a PR, do better a review. hence, I'll be able to notify you again just clicking the :repeat: icon in the sidebar :innocent:

ferblape:
You are right

ferblape (commented):
Ok to me, but in Slack you said:

> hugo
> veo que ocurre tb en las propias gráficas, es decir, los ejes de ellas son clickables, si siclas en uno que ya no tiene nada tb se estropea la grafica

> (https://populate.slack.com/archives/C2Z5EH5MW/p1728659073466039?thread_ts=1728658562.677229&cid=C2Z5EH5MW)
> habría que anular los enlaces si no tienen valor

Links still are clickable inside the empty charts. Maybe should we hide all the charts if filters are empty? Or at least make them clicakable? I think I'd prefer to hide them because now are a bit confusing

Crashillo:
I went through that option, but I see far more complicated (links are inside the dc library). I handle the empty values as a 0 instead, and no errors are thrown (division by zero was actually the conflictive value)

ferblape:
Ok, then hidding the charts when are empty?

Crashillo:
You mean something like this? I'll give a try

imagen

ferblape:
Yes

ferblape:
I meant yes, but the "No data" label should appear in all charts, not just in those in the right

Crashillo:
Look what I said y'day in slack:

> añadido un label para ocultar los graficos crossfilter cuando no hay datos. Está desplegado, pero crossfilter actua de una manera que no termino de entender: si hay dos filtros activos y excluyentes (es decir, que deberia dar 0), crossfilter parece considerar como mínimo 1, si algo de ese gráfico está activado, por lo que no se oculta.

On the another side, the _Importe por tipo de procedimiento_ chart shows a very tiny values (i.e. 0.53252415e-7), what differs from 0, then the chart is displayed. I'm trying to figure out why or where these values come from.

Crashillo:
I was able to handle the tiny values issue, but the default behaviour of crossfilter is not possible to workaround. It only works properly when the selected sidebar filters don't match with the charts themselves. Look at the image:

imagen

Filters selected: _dates_ and _categories_, result ok, since these filters don't have to do with the charts. But, whenever you select any filter with a chart-related, crossfilter "keeps" such value (I presume in order to keep the gray bars)

imagen

ferblape:
Ok, let's consider it an edge case. OK to merge

PR #4262, opened 2024-10-21
Filters sidebar

### LINKED ISSUE #3928: Filters sidebar
Currently there are several implementations of the filters sidebar along Gobierto app. They're all slightly differents each other, but some of them can be merged.

- investments
- data
- visualizations
- documents (engine)

The best approach due to its isolation, loose coupling, high cohesion and simplicity is at [vue/mixins](https://github.com/PopulateTools/gobierto/blob/master/app/javascript/lib/vue/mixins/modules/items-filter.js) (code will be brought to gobierto base shortly).

Regarding _gobierto-data_ the refactor will be quite simple, as such module was using a needless property there (`store`)

Regarding _gobierto-investments_ will be little trickier, as the `store` property (perhaps no longer necessary) is pretty binded all through the module

Regarding _gobierto-visualizations_ will require a further look, as the filters trigger events also, in order to make dc.js listening such events

Once we have some shared, working code, it could be wrapped into a Vue component to create the default filters sidebar when no specific HTML/CSS is required (currently you choose what inner components will render)

Related to  https://github.com/PopulateTools/issues/issues/1208

Crashillo:
Closes #3928 

## :v: What does this PR do?
Refactorize _investments_ module in order to replace the ad-hoc filters sidebar with the vue component, available in _data_ module for instance.

Unfortunately, _visualizations_ module is not possible to adapt it in the same way as _investments_ does, since the endpoints requested are very different and the vue `Filters` component are quite bound to the Gobierto Data API structure.

https://mataro.gobify.net/inversiones


furilo:
How are we managing/configuring the order of the items of a filter?

Crashillo:
Such order can be defined in the backoffice, example: https://mataro.gobify.net/admin/vocabularies/103/terms, just as the vocabulary options are sorted, they'll display like that

furilo:
Please add the corresponding documentation, since filters in Contratos vizz
or here work differen

Crashillo:
Included a note in the [vue components catalogue](https://gobierto.readme.io/docs/vue-components-catalogue#filters)

Crashillo:
Can this be merged?

PR #4267, opened 2024-11-05
Replace momentjs

Crashillo:
## :v: What does this PR do?
`momentjs` has become a outdated library (maintainers recommend to use different libs), apart from its package size, so it better to replace the utilization with plain, buitl-in, no 3rd-party code. 
`fullcalendar` has been replaced with CDN scripts, since the legacy version used, requires `moment` to make it work

https://mataro.gobify.net/presupuestos/proveedores-facturas
(fullcalendar) https://madrid.gobify.net/agendas/ed-metz

### DIFF
--- a/app/views/gobierto_people/people/person_events/index.html.erb
+++ b/app/views/gobierto_people/people/person_events/index.html.erb
@@ -66,6 +66,15 @@
 
 <% description([title, t("gobierto_people.layouts.application.title"), @site.title].compact.join('. ')) %>
 
+<% content_for :gobierto_footer_content do %>
+  <%= javascript_include_tag 'https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.30.1/moment.min.js' %>
+  <%= javascript_include_tag 'https://cdnjs.cloudflare.com/ajax/libs/fullcalendar/3.10.2/fullcalendar.min.js' %>
+<% end %>
+
+<% content_for :stylesheet_module_link do %>
+  <%= stylesheet_link_tag 'https://cdnjs.cloudflare.com/ajax/libs/fullcalendar/3.10.2/fullcalendar.min.css' %>
+<% end %>
+
 <% content_for :javascript_hook do %>
   window.GobiertoPeople.person_events_controller.index();
 <% end %>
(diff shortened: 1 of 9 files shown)


furilo:
Where is this used? Can you provide links of before & after so we can review with detail?

Crashillo:
The most complex use were in budgets/invoices (OP updated); in _gencat_ has a significant use also, I'm replacing this last one

furilo:
Why are we fetching from the CDN? I think we should avoid doing this.

Crashillo:
It is an edge case: _fullcalendar_ is only used in one place (agendas), besides, since that code were developed long time ago, the version we were using was `3.10.2` then, what is totally deprecated (they're releasing the v7, actually). I tried to upgrade it, but, expectedly, that code requires a refactor, as no longer depends on `jquery`. Therefore, I chose to use a CDN instead.

IMO, is the easiest solution to take over this sort of issues, as the uses are very exceptional, just one-time actually.

furilo:
I'd try to update it.

Crashillo:
Ok, in such case, I will addressed to an issue to tackle separately from this PR, indeed, the tests are green

furilo:
If you select "Todo" there is a JS error.

Crashillo:
Issue was created [here](https://github.com/PopulateTools/gobierto/issues/4268), so this PR can be merged (same naming, but different repo! :sweat_smile: )

app/views/gobierto_people/people/person_events/index.html.erb
```diff
@@ -66,6 +66,15 @@
 
 <% description([title, t("gobierto_people.layouts.application.title"), @site.title].compact.join('. ')) %>
 
+<% content_for :gobierto_footer_content do %>
+  <%= javascript_include_tag 'https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.30.1/moment.min.js' %>
+  <%= javascript_include_tag 'https://cdnjs.cloudflare.com/ajax/libs/fullcalendar/3.10.2/fullcalendar.min.js' %>
```
ferblape, line 71: Why is the CDN necessary?

furilo: LOL https://github.com/PopulateTools/gobierto/pull/4267#issuecomment-2459554480

ferblape: right 🤔

ferblape (approved):
Everything seems to work, so I guess we can merge and remove the CDN with the calendar upgrade.

PR #4291, opened 2025-04-24
Redirect to root url if module is configured as home

ferblape:
## :v: What does this PR do?

Unexpected.

Adds the before_action required to configure a homepage module in the root url and prevent duplicates.

This is happenning in Gencat visualization, where both https://direcciopublica.transparencia.gencat.cat/ and https://direcciopublica.transparencia.gencat.cat/cargos-y-agendas work, while the second url should redirect to the root path.

That's the behaviour in other modules (such as observatory or budgets)


ferblape:
Tested in staging https://gencat.gobify.net/cargos-y-agendas

PR #4321, opened 2025-12-19
Skip nodes where published_version is not present

entantoencuanto:
## :v: What does this PR do?

This PR prevents exceptions that happen when a project of a Plan is published but its published version is not set. This should not happen (published without version) but the exception can be prevented since it is related with a scope that should only operate on projects with a published version. The PR causes projects without a published version to be ignored from this scope.

- [ ] new site template?

### DIFF
--- a/app/models/gobierto_plans/node.rb
+++ b/app/models/gobierto_plans/node.rb
@@ -71,7 +71,9 @@ module GobiertoPlans
       end
     }
     scope :versions_indexes, lambda {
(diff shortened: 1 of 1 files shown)


ferblape:
@entantoencuanto what's the status of this PR?

entantoencuanto:
It's ready. There is an underlying issue I'm reviewing, this PR fixes the unexpected effects of it so it's ready to merge

ferblape (approved):

Issue #2150, opened 2019-03-06
Activate HTTP/2


furilo:

ferblape:
We already have it:

![screen shot 2019-03-06 at 16 42 48](https://user-images.githubusercontent.com/17616/53893661-e7044200-402e-11e9-9033-301f6f17fcd0.png)

furilo:
Mmm... The Google Lighthouse Audit (inside Chrome) says otherwise: 

![image](https://user-images.githubusercontent.com/24740/53894037-a8bb5280-402f-11e9-8da7-74063cfd40a5.png)

ferblape:
http2 is already implemented in the new app server

PR #4370, opened 2026-05-13
Fix HTML rendering on person agenda pages (#2150)

ferblape:
Closes [#2150](https://github.com/PopulateTools/issues/issues/2150)

## :v: What does this PR do?

Fixes the recurring "raw HTML rendered" bug on `/agendas/<slug>` pages. `respond_to` listed `format.js` before `format.html` in `GobiertoPeople::People::{PersonEvents,PastPersonEvents}Controller#index`, so `Accept: */*` requests (curl, bots) rendered the JS body of `index.js.erb`, but `request.format` stayed `:html` — `caches_action` then wrote the JS body into the HTML cache slot, and every subsequent visitor got it back with `Content-Type: text/html`. Putting `format.html` first keeps the cached body in sync with `request.format`; legitimate AJAX pagination still routes to `format.js` via rails-ujs `Accept: text/javascript`. Includes a regression controller test and removes a duplicated `cache_service` method in `gobierto_people/application_controller.rb`.

1. After deploy, trigger a `site.updated_at` bump (any admin save on the affected site) to evict any currently poisoned cache entry.
2. `curl https://<host>/ca/agendas/<slug>` → must return real HTML (was JS body before).
3. `curl -H 'Accept: */*' https://<host>/ca/agendas/<slug>` → still HTML.
4. Click "Load more" in the agenda list in a browser → AJAX pagination still works (returns JS).

### Before this PR

See screenshot in #2150 — JS code visible as plain text on the page.

### After this PR

Normal HTML agenda page.

- [ ] new site template?


ferblape:
Released to staging

PR #4395, opened 2026-07-27
Contracts viewer: fix multi-lot grouping and derived-contract identity, add código de expediente and contracting system, and a large performance pass

eltercero:
### Summary

This PR bundles the work done on the contracts viewer:

1. **Bug fixes** to multi-lot tender rendering.
2. A new ***código de expediente*** feature (display + search).
3. The **contracting system**: a new sidebar filter and a row in the detail.
4. **Identity of derived contracts and their tenders**: a contract detail is no longer non-deterministic, and the establishment tenders get a page of their own.
5. A **performance pass** that makes the viewer usable on large sites (UJI, ~34k contracts), where filtering used to freeze the page for ~2 s.
6. An **HTTP-cache** correctness fix.
7. The **test-suite change** that fix turned out to require.

All changes are scoped to the contracts viewer (plus one shared `Table.vue` addition that is backward compatible), except items 6 and 7, which are app-wide and worth a careful read.

Items 3 and 4 depend on three columns the ETL has started publishing — `contract_id`, `tender_id` and `contracting_system`. **11 sites are still served the old CSV**, so every one of those changes degrades to the current behaviour when a column is absent; §4 spells out how, and there is a test pinning it.

---

### 1. Multi-lot tender fixes

- **Group lots by tender id, not title.** The show page rebuilt a tender's lots with `groupBatches`, filtering `contractsData` by matching `title`. That assumed every lot shared an identical title; current data gives each lot its own title (e.g. *"Quiosco 2…"* vs *"Quiosco 3…"*), so only one lot showed. Now grouped by the shared tender `id` — which is also stricter, since an id is never shared across different tenders.
- **Use the tender title for the multi-lot header.** The header showed one lot's lot-specific title instead of the tender's *objeto*. It now looks up the tender in `tendersData` by shared id and uses its title, falling back to the lot title when no tender matches.

### 2. Código de expediente

- **Display:** the tender's `document_number` (código de expediente) now appears on the detail page (labelled block with a folder icon, before *Categoría*) and as an optional, hidden-by-default column in the table. Contracts don't carry it themselves — it lives on the tender and is shared by all its lots — so it's joined into contracts data by tender id in the controller.
- **Search:** the search box now also matches código de expediente, alongside assignee and contract title, in both the table filter and the grouped/summary filter. Placeholder text updated in es/ca/en.

### 3. Contracting system

The ETL publishes `contracting_system` as the CODICE integer (0 `none`, 1–2 the framework-agreement / dynamic-acquisition **establishments**, 3–4 the contracts **derived** from them). The viewer now translates it — the labels are ours — and shows it as a row in the detail, with `0` normalised to an empty string so it behaves like every other empty field: no filter option, no row.

The sidebar gains a filter for it. It has no dc chart behind it, because most contracts have no contracting system and a distribution chart would add nothing to the summary; that is what `chartlessFilters` is for — its own crossfilter dimension and selection set, kept out of `this.charts`, whose loops assume a container and a node. The filter is deliberately **not** propagated to the tenders dataset: derived contracts carry 3 or 4 while establishment tenders carry 1 or 2, so the join would empty the tenders panel.

On a site still serving the old CSV every value is empty, the block is left with fewer than two options, and `Aside` drops it as it already does for any such filter.

### 4. Identity of derived contracts and their tenders

**The bug.** Since the ETL was fixed, `contratos.id` holds the *effective tender*, so the 49 contracts the UJI derived from one dynamic acquisition system all share `id = 4372265`. The viewer routed and resolved by `id`, so `/adjudicaciones/4372265` was **non-deterministic**: header, table mode and amount all depended on whichever derived contract happened to land first in `contractsData`. Two loads of the same URL on staging showed `BSDA/3/24-SE` / 737,36 € and `BSDA/13/26-SE` / 178.977,04 € — for a contract of 2.224,09 €.

- **Routing by `contract_id`.** `contractRoutingId({ id, contract_id }) => contract_id || id` is the single helper every route-building site now uses (table rows, beeswarm clicks, assignee page, sibling rows). `contract_id` is unique per contract — 34,314 out of 34,314 in the UJI.
- **A page for the establishment tenders.** The detail resolves a contract first and falls back to the tender with that id. Establishment tenders have no row in the contracts dataset and were unreachable until now; they reuse `ContractsShow.vue` behind an `isTenderDetail` flag rather than growing a second component. Contract-only blocks are hidden, the table is always in `derived` mode, the total is labelled as the sum of the derived contracts, and the 213 tenders with no contracts get an empty-state line. **No new routes.**
- **Each contract keeps its own title and its own amount.** `is_derived_contract` is computed from `contracting_system` in the controller, before the translation overwrites the integer and before the row is frozen. A derived contract names what was actually bought, and its amount is its own — it is not one lot of a whole. The batch number alone could not decide this: 18 of the derived contracts carry one.
- **Reaching the establishment, and finding it by searching.** A derived contract links to its establishment tender ("Licitació d'establiment"), and searching for the establishment's expediente finds all the contracts derived from it — no contract carries that expediente itself. Both read `establishment_title` / `establishment_document_number`, resolved once per load in the controller rather than per keystroke in the search.
- **Tender statuses.** The detail shows the status, and five of the values the datasets publish had no label at all (88 `provisionally_awarded`, 22 `revoked`, 12 `void`, 2 `formalized`, 1 `abandoned`) — they rendered the I18n missing-key message. Added in es/ca/en, and an empty status now stays empty, guarded like every other translated field.

**Known limitation, deliberately not fixed here:** a derived contract that is *also* a lot cannot find its sibling lots, because the ETL exports `COALESCE(establishment_tender_id, tender_id)` in a single column. Exporting both separately is the follow-up.

**Degradation.** A site without `contract_id` routes by `id`, exactly as its current URLs already do, so every contract still answers first and the tender branch is never reached from the UI; without `contracting_system` no contract is derived, so titles, amounts and table mode behave as they do today. `test_legacy_csv_degrades_to_the_previous_behaviour` strips the three columns from the fixtures and pins this.

### 5. Performance pass

On large sites, applying and (especially) clearing a filter blocked the main thread for seconds. The main culprits and fixes:

- **gobierto-vizzs BeeSwarm was O(n²).** `parse()` spread the accumulator per row and `Base.groupBy()` rebuilt each bucket per item (~10⁸ element copies at ~34k rows), and `build()` leaked a d3 `forceSimulation` on every rebuild. These are now fixed **upstream in gobierto-vizzs 3.4.0**; this PR consumes `^3.4.0` from npm and drops the temporary local subclass.
- **Stop cloning rows to attach `href`.** Rows are frozen at load, then were cloned with `{...d, href}`, which unfroze them and made Vue install reactive getters/setters on every field of every row. Nothing needed the pre-baked `href`; `Table.vue` gains an optional `hrefBuilder` prop that builds the link lazily from the frozen row. The existing `href` string API is untouched (subsidies/debts/odss/costs unaffected).
- **Defer summary vizz rebuilds while its tab is hidden.** The Summary tab lives under `v-show` next to the table; its watchers rebuilt two full treemaps + the beeswarm on every filter change even when off-screen. It now skips the rebuild while hidden and flushes once, on `nextTick`, when the tab becomes visible. Behaviour is unchanged while the tab is on screen.
- **Defer and coalesce the refresh on filter change**, so the click and dc redraw paint first and rapid toggles collapse into one refresh; reuse the original array reference when the full dataset returns so it isn't seen as a change by identity.
- **Compute the filtered row count once per redraw** instead of one full-dataset sort per bar.
- **Freeze data rows** so Vue 2 skips reactivity on tens of thousands of rows at startup.
- **Memoize `I18n.t`** per distinct key, hoist constant status labels out of filter predicates, and **debounce** the search handler (250 ms).
- **Lowercase the search query once** per filter instead of once per row.
- **Preload the contracts/tenders CSVs** from the page head (parallel with the bundle) and **load pinned, minified d3/dc** builds.

### 6. HTTP cache

`Cache-Control` emitted a bare `"public,86400"`. `86400` on its own is not a valid directive, and with no `max-age`, `Expires` or `Last-Modified` a browser cannot derive a freshness lifetime — so in practice nothing was being cached at all.

Just making it valid (`public, max-age=86400`) would turn that into a hard 24-hour cache across the whole public front end and the public API in a single step. That is more than we want, for three reasons:

- **It applies everywhere.** `set_cache_headers` is a `before_action` on both `ApplicationController` and `ApiBaseController`, so every public module and the public API are covered. (Admin is not — `GobiertoAdmin::BaseController` overrides it with `expires_now`.)
- **`public` is unsafe on pages that carry a session.** `GET /user/sessions/new` returns the header alongside `Set-Cookie: _gobierto_session` and a CSRF token. `public` explicitly permits a shared cache to store that response and re-serve it to a different visitor.
- **There is no HTML CDN in front of the app.** `asset_host` covers assets only, so a long `max-age` would be absorbed entirely by the visitor's browser — the one layer that cannot be purged.

So this PR ships a deliberately conservative value:

```
Cache-Control: private, max-age=900
Vary: Cookie
```

`Vary: Cookie` is what makes browser-side caching correct here. Locale lives in a cookie rather than in the URL (`ApplicationController#set_locale`) and nav links carry no locale param, so without it the home page would come back in the previous language right after switching. Signing in and signing out would go stale for the same reason.

`http_headers_test.rb` previously asserted only `Cache-Control.include?("public")`, which passed just as happily with the broken header. It now pins the actual directives for both the anonymous and the signed-in branch.

**Deliberately left for a follow-up PR:** moving to `public` with `s-maxage` once a purgeable CDN exists, and, as a precondition for that, keeping session cookies and CSRF tokens out of cacheable responses. Product also needs to weigh in on how much staleness each module tolerates — budgets and visualisations can take hours, CMS news minutes, anything with a form none.

### 7. Test-suite fix: browser HTTP cache

Making the header valid exposed a latent problem in the Capybara setup, so it belongs with this change.

Capybara keeps one Chrome alive for the whole run and clears cookies and storage between tests, but never the HTTP cache. Once the app started emitting a header Chrome could act on, a page fetched by one test was served from cache to the next test visiting the same URL, ignoring the new server state.

That broke `VisualizationsContractsDownloadDataTest`, where both tests visit the contracts summary with different module settings: whenever `test_download_data_filled` ran first, `test_download_data_missing` read its cached HTML and found the download link it asserts is absent. Reproducible with `--seed 1`.

`test/support/drivers.rb` now disables the browser cache through CDP (`Network.setCacheDisabled`) when the browser starts, for both `:chrome` and `:headless_chrome`. Notes on the implementation:

- No Chrome flag does this reliably — `--disk-cache-size=0` leaves the in-memory cache alone, and `--disable-application-cache` is about the long-removed AppCache.
- It runs on the first `browser` call rather than at driver registration, because Capybara builds the browser lazily and there is nothing to talk to over CDP at registration time.
- It is re-applied whenever `@browser` is nil again, so a crashed or quit browser doesn't come back with caching switched on.

---

### Performance results (validated on staging — UJI, ~34,152 contracts)

Longtask maximum via `PerformanceObserver`, driving the real filter handlers:

| Action | Before (3.2.0) | After |
|---|---|---|
| Apply filter (table) | 273 ms | ~115–145 ms |
| Clear filter (table) | 1,786 ms | ~95–110 ms |
| Full RESUM rebuild over 34k (return from table) | ~4,954 ms | ~0.7–0.8 s |

No console errors; the beeswarm, treemaps and bar charts render correctly over the full dataset.

---

### Dependency

- `gobierto-vizzs` `^3.2.0` → `^3.4.0` (resolved from the npm registry). 3.4.0 carries the BeeSwarm/`groupBy` O(n) rewrites, the `forceSimulation` lifecycle fix, and a `Gantt.parse` O(n) fix. See the gobierto-vizzs PR.

---

### Derived-contract behaviour (validated on staging — `uji.gobify.net`)

| Case | Result |
|---|---|
| `/adjudicaciones/4372265` (establishment `SDA/1/23`) | Tender page: its own title, estimated value 592.413,50 €, 3 lots, "Contracte derivat" table with 49 rows and 49 distinct links, total labelled "Import total dels contractes derivats" = 178.977,04 €. No award date, no formalisation, no single assignee |
| Three full reloads of it | Identical all three times — this is the regression being closed |
| `/adjudicaciones/6163088` (`BSDA/2/24-SE`) | Its own title, and "Licitació d'establiment → SDA/1/23" linking to `/4372265` |
| `/adjudicaciones/10031368` (`BSDA/13/26-SE`, derived **and** carrying `batch_number = 1`) | 2.224,09 €, not the 178.977,04 € it used to show |
| `/adjudicaciones/320042` (ordinary 3-lot tender) | Unchanged: "Lot" column with 1/2/3 and the summed amount, 123.884,30 € |
| A minor contract | Unchanged: no siblings, no establishment block, its own expediente |
| One of the 213 tenders with no contracts | Tender page with the empty-state line |
| Searching `SDA/1/23` | 49 results, all of them derived from it, none carrying that expediente itself |

Degradation (§4) is covered by the integration suite rather than by staging, since every staging site already serves the new CSV.

---

### Notes for the reviewer

- Items 6 and 7 are the only app-wide changes; everything else is confined to the contracts viewer.
- `Table.vue`'s `hrefBuilder` prop is additive and defaults to `null`; existing consumers that pass a pre-baked `href` string are unaffected.
- The código de expediente join happens in `contracts_controller.js` by tender id.
- `contractRoutingId` and `effectiveTenderId` have jest tests, but **`yarn test` cannot run in this repo**: `babel.config.js` loads `babel-plugin-macros` and `@babel/preset-env`, neither of which is declared in `package.json` or present in `node_modules`, so every jest suite fails to start — the pre-existing ones included. `babel.config.js` also has a missing comma between `'@babel/plugin-proposal-class-properties'` and the array that follows it. Both are worth their own PR; the two helpers were verified by bundling them with esbuild and running the assertions in node.
- Both `is_derived_contract` and the establishment fields are precalculated in the controller, before `Object.freeze`: the components must not mutate a frozen row, and the search walks the whole dataset on every keystroke.
- A handful of `contract_id`s collide with a tender id. The contract wins and that tender stays unreachable — residual, not addressed here.
- Rails reorders `Cache-Control` directives when it serializes the response (`max-age=900, private`), which is why the test compares the set of directives rather than the literal string.
- The history keeps two intermediate steps for the gobierto-vizzs fix (local subclass → vendored tarball) that the final `^3.4.0` commit supersedes; they're left in place since this PR merges without squashing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aramollis (commented):
Issues reviewed:

1. Search by cod. exp. works ok 

2. All AM and SDA/restringidos doesn't show their cod. exp. and aren't linked to specific contract they all link to https://uji.gobify.net/visualizaciones/contratos/adjudicaciones/ we need to fix in order to show expedient ID and link to contract page

Captura de pantalla 2026-07-28 a les 14 53 06

3. Those Basado SDA are shown as "restringido" instead "basado en SDA". 
In GC we can filter as basados / restringido eg. [SDA 1/22 CC-2ª remesa ](https://contratos.gobierto.es/explorador/busqueda?ignore_tracking=&id=&search%5Bfilter_class%5D=SearchContract&from_entity=&search%5Btitle%5D%5B%5D=SDA+1%2F22+CC-2%C2%AA+remesa) because the differences from PLACE / PCSCP. Maybe the easier solution is to add new filter "Sistema de contratación"

Image

4. The batches link to the same contract [e.g. exp. SE/46/25](https://uji.gobify.net/visualizaciones/contratos/adjudicaciones/5013670) as in [PLACE](https://contrataciondelestado.es/wps/poc?uri=deeplink:detalle_licitacion&idEvl=yYve5xKg71U7%2B9FIQYNjeQ%3D%3D), but the "desierto" batch isn't shown, and we are reflecting the nº de licitadores only from one batch, not for each one (Lote 3: 1, Lote 2: 3, etc.).
Solutions: 
A - We could reflect the complete batch table data (including all batches, including "desierto", add a column for ofertas recibidas) 
B - or link each batch to its own contract page (similar to internal contracts in GC) 
C - or leave as it is ATM.

eltercero:
@aramollis could you please review again? I think points 1, 2, and 3 are alreay fixed. Please be sure to force refresh the browser, thisi s heavily cached from what I've seen. We can see this together if you want.

aramollis:
> @aramollis could you please review again? I think points 1, 2, and 3 are alreay fixed. Please be sure to force refresh the browser, thisi s heavily cached from what I've seen. We can see this together if you want.

Yes all those issues fixed! maybe we can merge and leave ATM 4 as it is

